PR #9, opened 2017-03-07
Fix a couple of typos in Events.md

Mokou:


docs/static/Events.md
```diff
@@ -39,7 +39,7 @@ are all run under the same supervisor. As part of the setup for each shard, it
 creates its own producer under a supervisor.
 
 As events are sent to the shard, the following happens:
- 1. Shard looks to see what type of event it is, only dispatch events are send to
+ 1. Shard looks to see what type of event it is, only dispatch events are sent to
  the producer.
 
  2. If the event is a `Dispatch`, the payload is converted to a an atom-keyed map.
```
Kraigie: Could you also correct the typo on this line? `to a an atom-keyed` -> `to an atom-keyed`

Kraigie:
Thanks!

docs/static/Events.md
Mokou: Np, didn't notice this one. Btw I think github gives you write access to push request branches by default.

Kraigie: If I'm understanding you correctly, you're right. I took the easy way out and delegated the work to you. :)

PR #11, opened 2017-03-28
Check for Keyword in process_request_body

cookkkie:
This is more accurate than checking for a list. 👍

### DIFF
--- a/lib/nostrum/api/base.ex
+++ b/lib/nostrum/api/base.ex
@@ -15,7 +15,7 @@ defmodule Nostrum.Api.Base do
     do: ""
   defp process_request_body({:multipart, _} = body),
     do: body
-  defp process_request_body(body) when is_list(body),
+  defp process_request_body(body) when Keyword.keyword?(body),
     do: Enum.into(body, %{}) |> Poison.encode!(body)
   defp process_request_body(body),
     do: Poison.encode!(body)


cronokirby:
I dont think that's allowed in a guard clause though...

Kraigie:
@cronokirby is right that `Keyword.keyword?` is not supported inside of guard clauses.

A list of supported guards can be found [here](http://elixir-lang.org/getting-started/case-cond-and-if.html#expressions-in-guard-clauses), or we could roll our own macro to define the guard (although I'm not sure how useful it would be).

What's interesting here is the actual intention of the guard clause. It was originally put in place to convert kwls into maps when passed as the body. This was often the case with API methods taking multiple parameters, for example: 
```Elixir
  @doc """
  Modifies a guild role.

  Role to modify specified by `guild_id` and `role_id`.

  `options` is a keyword list with the following optional keys:
   * `name` - Name of the role.
   * `permissions` - Bitwise of the enabled/disabled permissions.
   * `color` - RGB color value.
   * `hoist` - Whether the role should be displayed seperately in the sidebar.
   * `mentionable` - Whether the role should be mentionable.
  """
  @spec modify_guild_role(integer, integer, [
      name: String.t,
      permissions: integer,
      color: integer,
      hoist: boolean,
      mentionable: boolean
    ]) :: error | {:ok, Nostrum.Struct.Guild.Role.t}
  def modify_guild_role(guild_id, role_id, options) do
    case request(:patch, Constants.guild_role(guild_id, role_id), options) do
      {:ok, body} ->
        {:ok, Poison.decode!(body)}
      other ->
        other
    end
  end
```

That said, some functions now take a list of maps, causing this clause to error:
```Elixir
  @doc """
  Reorders a guild's roles.

  Guild to modify roles for is specified by `guild_id`.

  `options` is a list of maps with the following keys:
   * `id` - Id of the role.
   * `position` - Sorting position of the role.
  """
  @spec modify_guild_role_positions(integer, [%{
      id: integer,
      position: integer
    }]) :: error | {:ok, [Nostrum.Struct.Guild.Role.t]}
  def modify_guild_role_positions(guild_id, options) do
    case request(:patch, Constants.guild_roles(guild_id), options) do
      {:ok, body} ->
        {:ok, Poison.decode!(body)}
      other ->
        other
    end
  end
```

For now it should be sufficient to remove the guard clause. We never explicitly pass a kwl as the body, only as the `options` parameter to a lot of methods. The docs won't reflect the correct formal parameters, but it'll be consistent. I've started slowly doing a pass through all of the API methods to work on documentation and spec's as there are a lot of inconsistencies and this is another kink that needs to be worked out.

Thanks @cookkkie for unintentionally bringing this to my attention!

cookkkie:
My bad, didn't know about it not being supported there 😅 .
You're welcome, haha.

PR #12, opened 2017-03-29
Better HEARTBEAT and HEARTBEAT_ACK handling

cookkkie:
Heartbeat job is now in a task. And not receiving HEARTBEAT_ACK in time will disconnect the shard from the gateway. Also, after a disconnect the heartbeat task is now stopped.

This is barely tested but should work fine hopefully haha.

### DIFF
--- a/lib/nostrum/shard.ex
+++ b/lib/nostrum/shard.ex
@@ -64,11 +64,14 @@ defmodule Nostrum.Shard do
     {:ok, state}
   end
 
-  def websocket_info({:heartbeat, interval}, _ws_req, state) do
-    now = Util.now()
-    :websocket_client.cast(self(), {:binary, Payload.heartbeat_payload(state.seq)})
-    Event.heartbeat(self(), interval)
-    {:ok, %{state | last_heartbeat: now}}
+  def websocket_info(:heartbeat, _ws_req, state) do
+    if state.heartbeat_ack do
+      :websocket_client.cast(self(), {:binary, Payload.heartbeat_payload(state.seq)})
+      {:ok, %{state | heartbeat_ack: false}}
+    else
+      Log.info "HEARTBEAT_ACK not received in time. Disconnecting..."
+      {:close, "", state}
+    end
   end
 
   def websocket_info(:identify, _ws_req, state) do
@@ -84,6 +87,9 @@ defmodule Nostrum.Shard do
   def ondisconnect(reason, state) do
     Logger.warn "WS DISCONNECTED BECAUSE: #{inspect reason}"
     Logger.debug "STATE ON CLOSE: #{inspect state}"
+
+    Task.shutdown(state.heartbeat_task)
+
     if state.reconnect_attempts > 3 do
       {:close, reason, state}
     else
--- a/lib/nostrum/shard.ex
+++ b/lib/nostrum/shard.ex
@@ -69,7 +69,7 @@ defmodule Nostrum.Shard do
       :websocket_client.cast(self(), {:binary, Payload.heartbeat_payload(state.seq)})
       {:ok, %{state | heartbeat_ack: false}}
     else
-      Log.info "HEARTBEAT_ACK not received in time. Disconnecting..."
+      Logger.info "HEARTBEAT_ACK not received in time. Disconnecting..."
       {:close, "", state}
     end
   end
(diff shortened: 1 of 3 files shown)


lib/nostrum/shard/payload.ex
```diff
@@ -30,10 +30,10 @@ defmodule Nostrum.Shard.Payload do
       seq: nil,
       session: nil,
       reconnect_attempts: 0,
-      last_heartbeat: 0,
       shard_pid: pid,
       producer_pid: nil,
-      heartbeat_intervals: Enum.map(1..10, fn _ -> 0 end)
+      heartbeat_ack: false,
```
Kraigie: If we initialize this to false, we'll disconnect every time we start up. When we go to send the heartbeat for the first time, it won't see a previous ack.

lib/nostrum/shard/event.ex
```diff
@@ -69,7 +66,8 @@ defmodule Nostrum.Shard.Event do
   end
 
   def heartbeat(pid, interval) do
-    Process.send_after(pid, {:heartbeat, interval}, interval)
+    Process.sleep(interval)
+    send(pid, :heartbeat)
```
Kraigie: Shouldn't we loop here to keep sending heartbeats?

lib/nostrum/shard/event.ex
```diff
@@ -45,11 +42,11 @@ defmodule Nostrum.Shard.Event do
     else
       Logger.debug "IDENTIFYING"
       identify(self())
-
-      heartbeat(self(), payload.d.heartbeat_interval)
     end
 
-    state
+    heartbeat_task = Task.async(fn -> heartbeat(self(), payload.d.heartbeat_interval) end)
```
Kraigie: The call to `self()` here will be in the context of the task, when it should be in the context of the shard process.

lib/nostrum/shard.ex
```diff
@@ -64,11 +64,14 @@ defmodule Nostrum.Shard do
...
-    now = Util.now()
-    :websocket_client.cast(self(), {:binary, Payload.heartbeat_payload(state.seq)})
-    Event.heartbeat(self(), interval)
-    {:ok, %{state | last_heartbeat: now}}
+  def websocket_info(:heartbeat, _ws_req, state) do
+    if state.heartbeat_ack do
+      :websocket_client.cast(self(), {:binary, Payload.heartbeat_payload(state.seq)})
+      {:ok, %{state | heartbeat_ack: false}}
+    else
+      Log.info "HEARTBEAT_ACK not received in time. Disconnecting..."
```
Kraigie: `Log.info` -> `Logger.info` 😄

lib/nostrum/shard.ex
```diff
@@ -84,6 +87,12 @@ defmodule Nostrum.Shard do
   def ondisconnect(reason, state) do
     Logger.warn "WS DISCONNECTED BECAUSE: #{inspect reason}"
     Logger.debug "STATE ON CLOSE: #{inspect state}"
+
+    if state.heartbeat_task do
+      Task.shutdown(state.heartbeat_task)
+      state = %{state | heartbeat_task: nil}
```
Kraigie: It's considered unsafe (read: bad practice) to set a variable inside of `if`, `cond`, `case`, etc. Consider rewriting this to something like 
```Elixir
state = 
  if state.heartbeat_task do
    Task.shutdown(state.heartbeat_task)
    %{state | heartbeat_task: nil}
  end
```

lib/nostrum/shard/event.ex
```diff
@@ -32,10 +32,7 @@ defmodule Nostrum.Shard.Event do
 
   def handle(:heartbeat_ack, _payload, state) do
     Logger.debug "HEARTBEAT_ACK"
-    heartbeat_intervals = state.heartbeat_intervals
-    |> List.delete_at(-1)
-    |> List.insert_at(0, Util.now() - state.last_heartbeat)
-    %{state | heartbeat_intervals: heartbeat_intervals}
```
Kraigie: We were keeping track of this to supply some form of gateway latency to the user. Do you still think it's a good idea to remove it? I'm not too keen on what would be useful to supply to the user from the shard state, maybe we should have a discussion on this.

lib/nostrum/shard/payload.ex
Kraigie: We should also add these to the state payload docs

lib/nostrum/shard/event.ex
Kraigie: Also not sure on the use of `Task.async` here

lib/nostrum/shard/event.ex
cookkkie: Yep, forgot to loop... 😅

lib/nostrum/shard/event.ex
cookkkie: In case of discord outage, the gateway generally keep sending heartbeat_ack events, even if nothing is working. So I don't really know about the relevance of tracking that.

```
It is worth noting that in the event of a service outage where you stay connected to the gateway, you should continue to heartbeat and receive ACKs. The gateway will eventually respond and issue a session once it is able to.
```
From https://discordapp.com/developers/docs/topics/gateway#gateway-heartbeat

Kraigie:
Committed with d8d11022dc891a6b231794c3efb4cdcaf06879b7, thanks. Ended up reverting the task change until we switch to the new ws lib.

PR #13, opened 2017-04-17
Adds a ConsumerSupervisor behaviour

cronokirby:
This is an "alternate" model for event subscription, where events are handled in seperate tasks, pooled by the `:max_demand` if the supervisor. This should be useful as the final consumer of a user's pipeline, allowing for commands to be run from events in a much more flexible way. Multiplying this consumer multiple times is rather needless, although it can be done.


Kraigie:
Committed with b30aef33f1784ecbc72dc1ea912a094e00e96b8d, thanks!

PR #22, opened 2017-07-30
Update voice states on Gateway's Voice State Update events

ndac-todoroki:
### WHY
In order to interact with voice channels (ex. `!summon` commands in some bots, or just listing users in the same voice channels), nostrum must keep track of the `VOICE_STATE_UPDATE` events.
Although, other than the initial state update when the bot has connected to the guild, the `voice_states` value kept the same.

### WHAT
Made voice states update on hooks.
Only one state per member on channel should be kept, so it deletes the previous one depending on the `user_id`.

#### Example
Randomly selecting one person from the voice channel which the message sender belongs to, should be available as:
```ex
with \
  {:ok, channel = %{"guild_id" => guild_id_str}} <- API.get_channel(message.channel_id),
  {guild_id, ""} <- Integer.parse(guild_id_str),
  {:ok, server} <- Nostrum.Cache.Guild.GuildServer.get(id: guild_id),
  {:ok, voice_channel_id} <- server.voice_states
    |> Enum.find(%{}, fn(map) -> map.user_id == message.author.id end)
    |> Map.fetch(:channel_id)
do
  me = self()
  server.voice_states
  |> Enum.filter(fn(map) -> map.channel_id == voice_channel_id end)
  |> Enum.map(fn (voice_state) ->
    {:ok, member} = API.get_member(guild_id, voice_state.user_id)
    spawn_link fn ->
      (send me, {member})
    end
  end)
  |> Enum.map(fn (_pid) ->
    receive do { result } -> result end
  end)
  |> Enum.flat_map(fn
    %{"user" => %{"bot" => true}} -> []
    user -> [user]
  end)
  |> Enum.take_random(1)
end
```


lib/nostrum/shard/dispatch.ex
```diff
@@ -194,7 +194,7 @@ defmodule Nostrum.Shard.Dispatch do
     do: {event, UserCache.update(p), state}
 
   def handle_event(:VOICE_STATE_UPDATE = event, p, state),
-    do: {event, p, state}
+    do: {event, GuildServer.voice_state_update(p.guild_id, p), state}
```
Kraigie: The documentation for this event is located [here](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/consumer.ex#L112). Could you modify this as well to better represent what is now being returned?

lib/nostrum/cache/guild/guild_server.ex
```diff
@@ -442,4 +447,14 @@ defmodule Nostrum.Cache.Guild.GuildServer do
...
 
+  def handle_call({:update, :voice_state, guild_id, voice_state}, _from, state) do
+    old_voice_states = state.voice_states
+    new_voice_states =
+      [
+        voice_state |
+        state.voice_states
+        |> Enum.reject(fn(map) -> map.user_id == voice_state.user_id end)
+      ]
+    {:reply, {guild_id, old_voice_states, new_voice_states}, %{state | voice_states: new_voice_states}}
```
Kraigie: As a user would it be more useful to receive a new and old list, or just the singular new and old voice state?

If we decide the latter we'll need some new logic/design choices on how to represent a voice_state being created/deleted. If the former then this looks good. I'm open to either!

lib/nostrum/shard/dispatch.ex
ndac-todoroki: Is the structs only for documenting? I've created a new struct for this, but I can't see where it should be used (in code). As far as I track, Nostrum stores voice states in the GenServer as plain maps currently, should this change to structs too?

lib/nostrum/cache/guild/guild_server.ex
ndac-todoroki: With the list, one can tell other's states in the guild, but with only the single state not so much information will be provided (It can, if the user keeps track on all the states, but that's what nostrum does for him, so it would be ridiculous), so I think the former will be good enough for now...

lib/nostrum/shard/dispatch.ex
Kraigie: Documentation is one of reasons for structs. They are currently maps, and since this is being implemented properly I'd like them to be in structs. To implement this you simply need to add similar logic to the methods [here](https://github.com/Kraigie/nostrum/blob/08ea92bc4a35d1824f8a92a949898d5557b585e0/lib/nostrum/struct/guild.ex#L138) for the new struct you create.

lib/nostrum/cache/guild/guild_server.ex
Kraigie: My line of thinking is that if people need to see the voice stats of others, they'll be able to fetch the guild and retrieve it from there. Since the number of voice states will be fewer than other lists (such as presences), it may not be a big deal to dispatch whole lists. 

Ultimately I'll leave the decision up to you. If later on down the road whatever choice made here proves to be the incorrect choice, it can always be changed!

Kraigie:
Sorry for taking so long to get back to you on this!

lib/nostrum/cache/guild/guild_server.ex
ndac-todoroki: Yes, I agree with that, and in that point of view, I'm very questioning about receiving new/old lists *nor* new/old states. What will be the usages (if there are other ways to go)? So far I can't see any of it, so maybe we could consider it later; I'll go leave this as it is, just for now.

lib/nostrum/shard/dispatch.ex
ndac-todoroki: Understood very well! Thanks. I've updated my request.

lib/nostrum/cache/guild/guild_server.ex
ndac-todoroki: Also,

> Since the number of voice states will be fewer than other lists (such as presences), it may not be a big deal to dispatch whole lists.

The number of voice states might become the same count as the presences list, because currently we do not delete the state from the list, when the state for one user becomes `channel_id: nil`. So the max count of voice states equals the number of the guild members (including kicked members).

ndac-todoroki:
Fixed the documentation!

ndac-todoroki:
I've noticed that sometimes the `state` [here at the function parameter](https://github.com/Kraigie/nostrum/pull/22/commits/03509f2f13d2eba2ccff3023cb6325ade8d51339#diff-62c12668f4193ebb841467cf1497269cL450) is a map, not a Guild struct. In the last commit I tried to go around it, but `%{state | voice_states: new_voice_states}}` will raise an `KeyError` anyway. 
I could work around that too, but I'm suspicious about this state-can-be-a-map-or-struct thing, and while I am digging this further, I want  to keep this pr open.

ndac-todoroki:
I think something like #26 will solve this↑

Kraigie:
Looks like you merged #26 into this pr. Was this your intention? If so we can close #26 and move discussion to this pr. 

Regardless, work on this PR will likely have to wait until I get back from vacation as this internet in my hotel is shoddy at best, and I'll likely want to do some testing since these are big changes. 

Thanks!

jchristgit:
Hi @ndac-todoroki, are you still interested in working on this PR? I understand if you're not, just drop a short note and we'll take up on it

jchristgit:
I'm closing this since there's been no activity here for a while, and voice state updates have been introduced in the meantime - if there's some functionality that you're still missing, feel free to open an issue!

PR #25, opened 2017-08-13
Fix :unavailable key not found in :GUILD_DELETE p

### LINKED ISSUE #24: nostrum crashes on receiving particular event types
1. `iex -S mix`
2. let bot join server
3. kick bot out of server
4. application exits!

sample log (from where bot gets kicked): [gist](https://gist.github.com/ndac-todoroki/42d8657862827099e9666f490435e40e)

In [dispatch.ex 's `GUILD_DELETE` pattern](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/shard/dispatch.ex#L109-L112),  it tries to get a `p.unavailable` boolean.
But from the [Discord document](https://discordapp.com/developers/docs/topics/gateway#guild-delete-guild-delete-event-fields) it says `unavailable` is `true` or *not set*. (hey is that what we call a boolean!?)

> whether the guild is unavailable, should always be true. if not set, this signifies that the user was removed from the guild

So the dot-access method might be the very first problem here.

cookkkie:
Should fix #24

### DIFF
--- a/lib/nostrum/shard/dispatch.ex
+++ b/lib/nostrum/shard/dispatch.ex
@@ -108,7 +108,7 @@ defmodule Nostrum.Shard.Dispatch do
 
   def handle_event(:GUILD_DELETE = event, p, state) do
     :ets.delete(:guild_shard_map, p.id)
-    {event, {GuildServer.delete(p.id), p.unavailable}, state}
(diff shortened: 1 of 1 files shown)


Kraigie:
Thanks! 😄

PR #26, opened 2017-08-15
always make GuildRegistry hold Guild structs strictly

ndac-todoroki:
## Why
Some functions, for example `Nostrum.Cache.Guild.GuildServer.get/1` is documented to return a `%Guild{}` struct, but sometimes it did not. This was because `get/1` relies on `transform/1`, which relies on getting values from the Registry.
The Registry initially holds a `Nostrum.Struct.Guild` struct, by passing that struct to `GuildRegister.create_guild_process/2`, but after that almost all `handle_call`s updated the `state` with a simple map.
This becomes a problem such as: Dot-notations will raise an error when it was supposed to return a `nil` (if it was a struct). Since `GUILD_UPDATE` renewals the whole `state`, the entire thing becomes a map, and it breaks eg. `Nostrum.Cache.Guild.GuildServer.get/1`'s type document.

## What
- `handle_call`s' `state`s are restricted to `%Nostrum.Guild{}`s
- always do `to_struct/1` on updating and creating new elements
    - vals which supposed to be lists but having empty maps as default values inside `to_struct/1`, now have lists as default value

### concerns
- should the default empty maps in `to_struct/1` should be an empty struct?

### DIFF
--- a/lib/nostrum/cache/guild/guild_server.ex
+++ b/lib/nostrum/cache/guild/guild_server.ex
@@ -339,20 +339,21 @@ defmodule Nostrum.Cache.Guild.GuildServer do
     {:reply, tf.(state), state}
   end
 
-  def handle_call({:update, guild}, _from, state) do
+  def handle_call({:update, guild}, _from, %Guild{} = state) do
+    guild = guild |> Guild.to_struct()
     {:reply, {state, guild}, guild}
   end
 
-  def handle_call({:delete}, _from, state) do
+  def handle_call({:delete}, _from, %Guild{} = state) do
     {:stop, :normal, state, %{}}
   end
 
-  def handle_call({:create, :member, guild_id, member}, _from, state) do
-    new_members = [member | state.members]
+  def handle_call({:create, :member, guild_id, member}, _from, %Guild{} = state) do
+    new_members = [member |> Guild.Member.to_struct() | state.members]
     {:reply, {guild_id, member}, %{state | members: new_members}}
   end
 
-  def handle_call({:update, :member, guild_id, new_partial_member}, _from, state) do
+  def handle_call({:update, :member, guild_id, new_partial_member}, _from, %Guild{} = state) do
     member_index = Enum.find_index(state.members, fn member ->
       member.user.id == new_partial_member.user.id
     end)
@@ -361,12 +362,12 @@ defmodule Nostrum.Cache.Guild.GuildServer do
         {:ok, old_member} -> old_member
         :error -> %{user: %{}, roles: []}
       end
-    new_member = Map.merge(old_member, new_partial_member)
+    new_member = Map.merge(old_member, new_partial_member) |> Guild.Member.to_struct()
     new_members = List.replace_at(state.members, member_index, new_member)
     {:reply, {guild_id, old_member, new_member}, %{state | members: new_members}}
   end
 
-  def handle_call({:delete, :member, guild_id, user}, _from, state) do
+  def handle_call({:delete, :member, guild_id, user}, _from, %Guild{} = state) do
     member_index = Enum.find_index(state.members, fn member -> member.user.id == user.id end)
     deleted_member =
       case Enum.fetch(state.members, member_index || length(state.members) + 1) do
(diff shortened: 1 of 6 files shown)


lib/nostrum/cache/guild/guild_server.ex
```diff
@@ -339,20 +339,21 @@ defmodule Nostrum.Cache.Guild.GuildServer do
...
-  def handle_call({:delete}, _from, state) do
+  def handle_call({:delete}, _from, %Guild{} = state) do
     {:stop, :normal, state, %{}}
   end
 
-  def handle_call({:create, :member, guild_id, member}, _from, state) do
-    new_members = [member | state.members]
+  def handle_call({:create, :member, guild_id, member}, _from, %Guild{} = state) do
+    new_members = [member |> Guild.Member.to_struct() | state.members]
     {:reply, {guild_id, member}, %{state | members: new_members}}
```
Kraigie: Another issue I can see here (and it's replicated throughout the module) is that the info we're sending will not be in a struct. Here, the state of the guild will have this member as a struct, but the event that the user handles will be a map (when according to documentation it should be a struct).

Do you think it would be better to move the conversion logic to the client side of the server? We could also move it back to the dispatch module, although it may be weird there as the other caches are stored-ish in ETS and as such won't need this type of conversion. Of course we could also just do the conversion first on the server side. 

Let me know your thoughts on this!

lib/nostrum/struct/embed.ex
```diff
@@ -95,7 +95,7 @@ defmodule Nostrum.Struct.Embed do
     |> Map.update(:video, %{}, &Video.to_struct(&1))
     |> Map.update(:provider, %{}, &Provider.to_struct(&1))
     |> Map.update(:author, %{}, &Author.to_struct(&1))
-    |> Map.update(:fields, %{}, &Util.list_to_struct_list(&1, Field))
+    |> Map.update(:fields, [], &Util.list_to_struct_list(&1, Field))
```
Kraigie: Definitely a derp moment from me here. Makes much more sense to store these as the proper type.

lib/nostrum/struct/embed.ex
```diff
@@ -95,7 +95,7 @@ defmodule Nostrum.Struct.Embed do
     |> Map.update(:video, %{}, &Video.to_struct(&1))
     |> Map.update(:provider, %{}, &Provider.to_struct(&1))
     |> Map.update(:author, %{}, &Author.to_struct(&1))
```
Kraigie: Good point on whether we should store these as structs. I think the answer is that they should be.

I don't think it's worth your time to go and change every instance of this. Soon I'd like to replace these methods with the macro found [here](https://github.com/Kraigie/nostrum/blob/08ea92bc4a35d1824f8a92a949898d5557b585e0/lib/nostrum/util.ex#L23).

ndac-todoroki:
I realized that this might break things in many places, if we extend using structs instead of maps in many other places. Examples:

- [code like this](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/cache/guild/guild_server.ex#L364) will end up replacing some vals with unexpected `nil`s, because discord only sends us partial data
  - this code currently works because `new_partial_member` is not a struct but a map (though currently this merge breaks the struct because `new_partial_member` has an unknown `guild_id` key, it's another problem)
   - but if [we convert those to structs more earlier](https://github.com/Kraigie/nostrum/pull/26#discussion_r133190641), meaning `new_partial_member` will be a struct and thus `Map.merge` will be bad.
- [`to_struct` functions which have other `to_struct`s](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/struct/guild/member.ex#L59) needs rewriting too because `Map.update/3` won't work
  - although I think just making four patterns will do:
  1. when `&1` is `nil`,
  2. `&1` is already the struct type we want,
  3. `&1` is a map,
  4. anything else (return error or set default empty struct)

I'm afraid any code currently relying to maps (calling Enumerable or Access, or expecting a KeyError) will no longer work. Is it good to keep doing this? :( @Kraigie

jchristgit:
Thank you for the PR. We have updated the guild caching logic to use ETS, and the ETS table contains `Nostrum.Struct.Guild` structs. If there's still some missing functionality for you, feel free to open an issue!

PR #28, opened 2017-09-01
Include proper channel caching

PixeLInc:
Changes it so that guild_id in cached channels is not nil.


lib/nostrum/shard/dispatch.ex
```diff
@@ -85,21 +85,25 @@ defmodule Nostrum.Shard.Dispatch do
   end
 
   def handle_event(:GUILD_CREATE, p, state) do
-    p.members
+    updated = p.channels
```
Kraigie: The convention I try to follow for this style of assignment/piping can be found [here](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/shard/dispatch.ex#L177).

Could you modify this to be more in line?

Kraigie:
Thanks again!

PR #31, opened 2017-10-15
Add missing events

PixeLInc:
- [x] WEBHOOKS_UPDATE 
- [x] MESSAGE_REACTION_REMOVE_ALL

### DIFF
--- a/lib/nostrum/consumer.ex
+++ b/lib/nostrum/consumer.ex
@@ -111,6 +111,7 @@ defmodule Nostrum.Consumer do
   @type user_update :: {:USER_UPDATE, {old_user :: Nostrum.Struct.User.t, new_user :: Nostrum.Struct.User.t}, ws_state}
   @type voice_state_update :: {:VOICE_STATE_UPDATE, {map}, ws_state}
   @type voice_server_update :: {:VOICE_SERVER_UPDATE, {map}, ws_state}
+  @type webhooks_update :: {:WEBHOOKS_UPDATE, {webhook :: Nostrum.Struct.Webhook.t}, ws_state}
 
   @type event :: channel_create |
     channel_delete |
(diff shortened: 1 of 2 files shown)


lib/nostrum/consumer.ex
```diff
@@ -111,6 +112,7 @@ defmodule Nostrum.Consumer do
   @type user_update :: {:USER_UPDATE, {old_user :: Nostrum.Struct.User.t, new_user :: Nostrum.Struct.User.t}, ws_state}
   @type voice_state_update :: {:VOICE_STATE_UPDATE, {map}, ws_state}
   @type voice_server_update :: {:VOICE_SERVER_UPDATE, {map}, ws_state}
+  @type webhooks_update :: {:WEBHOOKS_UPDATE, {webhook :: Nostrum.Struct.Webhook.t}, ws_state}
```
Kraigie: Could we leave the return type here as a map? Most of these types are lies anyways atm, but since Nostrum doesn't track webhooks at all currently its even further from the truth,

lib/nostrum/consumer.ex
```diff
@@ -102,6 +102,7 @@ defmodule Nostrum.Consumer do
   @type message_update :: {:MESSAGE_UPDATE, {updated_message :: Nostrum.Struct.Message.t}, ws_state}
   @type message_reaction_add :: {:MESSAGE_REACTION_ADD, map}
   @type message_reaction_remove :: {:MESSAGE_REACTION_REMOVE, map}
+  @type message_reaction_remove_all :: {:MESSAGE_REACTION_REMOVE_ALL, map}
```
Kraigie: This, and all of the other types here should be in the format of {event_type , {things_to_dispatch}, ws_state}.

I see some of the surrounding types are still incorrect here, could you fix them up to more in line with the rest? Thanks!

PixeLInc: Ah alright. I based it off of the previous ones about reactions. I'll fix em up in my next commit

Kraigie: No problem! I'm the one that lead you astray. 😅

Kraigie:
Thanks again!

PR #33, opened 2017-11-11
Add an example for using Cache

PixeLInc:
Just a simple example of how to use the Cache system.

### DIFF
--- a/examples/cache.ex
+++ b/examples/cache.ex
@@ -1,0 +1,65 @@
+defmodule ExampleSupervisor do
+  def start do
+    import Supervisor.Spec
+
+    # List comprehension creates a consumer per cpu core
+    children = for i <- 1..System.schedulers_online, do: worker(ExampleConsumer, [], id: i)
+
+    Supervisor.start_link(children, strategy: :one_for_one)
+  end
+end
+
+defmodule ExampleConsumer do
+  use Nostrum.Consumer
+
+  alias Nostrum.Api
+
+  require Logger
+
+  def start_link do
+    Consumer.start_link(__MODULE__, :state)
+  end
+
+  def handle_event({:MESSAGE_CREATE, {msg}, ws_state}, state) do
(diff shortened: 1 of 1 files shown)


examples/cache.ex
```diff
@@ -0,0 +1,76 @@
...
+      "!userinfo" -> # !userinfo user_id
+        if tuple_size(args) == 2 do # Check to make sure there are 2 arguments, meaning the command, and then an argument
+
+          # We grab the second thing from the tuple
+          user_id = elem(args, 1)
+
+          if user_id != nil do
+            try do
+              # try and convert the supplied input to an integer
+              user_id = user_id |> String.to_integer
```
Kraigie: I'd recommend we use `Integer.parse/1` here. By doing so we could remove the try/rescue and flatten the `with` statement even further.

examples/cache.ex
```diff
@@ -0,0 +1,76 @@
...
+                   {:ok, channel} <- Nostrum.Cache.ChannelCache.get(id: msg.channel_id),
+                                     Api.get_channel(msg.channel_id),
+                   {:ok, guild} <- Nostrum.Cache.Guild.GuildServer.get(id: channel.guild_id),
+                                   Api.get_guild(channel.guild_id)
+
+              do
+                Api.create_message!(msg.channel_id, "#{user_id} belongs to a person named: #{user.username}\nMessage sent in: #{channel.name}\nChannel is in: #{guild.name}")
+              else
+                # For this, we just print out we can't find anyone while using the error from the with statement. If it can't find someone, it will print out `user_not_found` as the reason.
+                {:error, reason} ->
```
Kraigie: If we decide to remove the try, this should include a case for failing to parse the integer.

Kraigie:
At this point we might as well flatten everything into the `with`. I believe that'd more idiomatic.

Kraigie:
Thanks!

PR #34, opened 2017-11-17
Implement new statuses

PixeLInc:
Modify the update_status methods to allow a `type` parameter. 
This is to allow users to use the new type statuses that have been implemented (Listening & Watching).

It also works out for future updates to the presence api so long as they just add in a new type, whatever that may be.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -99,12 +99,12 @@ defmodule Nostrum.Api do
     - `pid` - Pid of the shard.
     - `status` - Status of the bot.
     - `game` - The 'playing' text of the bot. Empty will clear.
-    - `type` - The type of status to show.
-    - `streaming` - URL of twitch.tv stream
+    - `type` - The type of status to show. 0 (Playing) | 1 (Streaming) | 2 (Listening) | 3 (Watching)
+    - `stream` - URL of twitch.tv stream
   """
-  @spec update_status(pid, status, String.t) :: :ok
+  @spec update_shard_status(pid, status, String.t, integer, String.t) :: :ok
   def update_shard_status(pid, status, game, type \\ 0, stream \\ nil) do
-    Session.update_status(pid, to_string(status), {game, stream}, type)
+    Session.update_status(pid, to_string(status), game, stream, type)
     :ok
   end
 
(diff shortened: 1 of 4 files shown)


lib/nostrum/api.ex
```diff
@@ -99,22 +99,23 @@ defmodule Nostrum.Api do
     - `pid` - Pid of the shard.
     - `status` - Status of the bot.
     - `game` - The 'playing' text of the bot. Empty will clear.
+    - `type` - The type of status to show.
```
Kraigie: Could we list the different types of status here?

lib/nostrum/api.ex
```diff
@@ -99,22 +99,23 @@ defmodule Nostrum.Api do
...
     - `streaming` - URL of twitch.tv stream
   """
   @spec update_status(pid, status, String.t) :: :ok
```
Kraigie: This spec is not correct (was incorrect prior to your changes) as it's missing the default parameters, could it be fixed to be in line with what's expected?

lib/nostrum/shard/payload.ex
```diff
@@ -41,28 +41,14 @@ defmodule Nostrum.Shard.Payload do
   end
 
   @doc false
-  # No stream
-  def status_update_payload(idle_since, game, status, afk, nil) do
+  def status_update_payload(idle_since, {game, stream}, status, afk, type) do
```
Kraigie: I believe the intent was to pattern match on the `stream` variable here to check for nil values. This is assuming that we can't send a url param with a nil value. If we can, then this is fine.

lib/nostrum/api.ex
```diff
@@ -99,22 +99,23 @@ defmodule Nostrum.Api do
...
-  def update_shard_status(pid, status, game, stream \\ nil) do
-    Session.update_status(pid, to_string(status), game, stream)
+  def update_shard_status(pid, status, game, type \\ 0, stream \\ nil) do
+    Session.update_status(pid, to_string(status), {game, stream}, type)
```
Kraigie: I believe the reason for having a tuple here was lost. It should be sufficient to simply pass the variables as they are.

lib/nostrum/api.ex
```diff
@@ -99,22 +99,23 @@ defmodule Nostrum.Api do
     - `pid` - Pid of the shard.
     - `status` - Status of the bot.
     - `game` - The 'playing' text of the bot. Empty will clear.
+    - `type` - The type of status to show.
     - `streaming` - URL of twitch.tv stream
```
Kraigie: This should be `stream`, my mistake, might as well fix it while you're in here. :^).

Kraigie:
Thanks!

PR #38, opened 2018-01-15
Fixed ChannelCache initial state.

Skippe:
## Why

The ChannelCache GenServer was incorrectly starting with an empty list as an initial state instead of an empty map. This meant that DM channels could not be properly stored inside of ChannelCache's state.

## What

Changed the initial state of ChannelCache to `%{}` instead of an empty list `[]`


Kraigie:
Thanks!

PR #41, opened 2018-01-21
Fix typespec for Nostrum.Api.create_message/3

ppraisethesun:
# Why
Typespec is incorrect for `create_message/3`

# What
Changed key from "content" to "file_name" in type `message_content`


Kraigie:
Thanks! In the future feel free to combine pr's like these two, as they're covering a lot of the same area.

PR #46, opened 2018-03-12
Api Correction Part 1: User struct

Skippe:
This is part one of the the bigger Api correction PR that I'm starting to break apart.

This PR accomplishes the following:

* Ensures the User struct snowflakes are decoded as integers instead of strings
* Fixes User-related functions typespecs, docs, and return values in Api module. Example: certain functions not returning a User struct where specified.
* Adds "bangified" functions for User-related Api calls.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -1360,7 +1360,7 @@ defmodule Nostrum.Api do
   end
 
   @doc """
-  Gets a user by its `user_id`.
+  Gets a user by its `t:Nostrum.Struct.User.id/0`.
 
   If the request is successful, this function returns `{:ok, user}`, where 
   `user` is a `Nostrum.Struct.User`. Otherwise, returns `{:error, reason}`.
@@ -1435,7 +1435,7 @@ defmodule Nostrum.Api do
   @spec modify_current_user!(keyword | map) :: no_return | User.t
   def modify_current_user!(params) do
     modify_current_user(params)
-    |> bangify()
+    |> bangify
   end
 
   @doc """
@@ -1852,7 +1852,7 @@ defmodule Nostrum.Api do
   defp handle_request_with_decode({:ok, body}, type) do 
     convert =  
       body 
-      |> Poison.decode!()
+      |> Poison.decode!
       |> Util.cast(type)
  
     {:ok, convert}
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -324,18 +324,34 @@ defmodule Nostrum.Api do
   @doc """
   Gets all users who reacted with an emoji.
 
-  Retrieves a list of users who have reacted with an emoji.
-
-  Returns `{:ok, [Nostrum.Struct.User]}` if successful, `{:error, reason}` otherwise.
+  If the request was successful, this function returns `{:ok, users}`, where 
+  `users` is a list of `Nostrum.Struct.User`. Otherwise, this function 
+  returns `{:error, reason}`.
   """
-  @spec get_reactions(integer, integer, String.t | Nostrum.Struct.Emoji.custom_emoji) :: error | {:ok, [Nostrum.Struct.User]}
+  @spec get_reactions(Channel.id, Message.id, String.t | Emoji.custom_emoji) :: error | {:ok, [User.t]}
   def get_reactions(channel_id, message_id, emoji) do
     case request(:get, Constants.channel_reactions_get(channel_id, message_id, emoji)) do
       {:ok, body} ->
-        {:ok, Poison.decode!(body)}
+        users =
+          body 
+          |> Poison.decode!() 
+          |> Enum.map(fn user_dto -> 
+            User.to_struct(user_dto) 
+          end) 
+         
+        {:ok, users}
       other ->
         other
     end
+  end
+
+  @doc """
+  Same as `get_reactions/3`, but raises `Nostrum.Error.ApiError` in case of failure.
+  """
+  @spec get_reactions!(Channel.id, Message.id, String.t | Emoji.custom_emoji) :: no_return | [User.t]
+  def get_reactions!(channel_id, message_id, emoji) do
+    get_reactions(channel_id, message_id, emoji)
+    |> bangify
   end
 
   @doc """
@@ -1355,16 +1371,30 @@ defmodule Nostrum.Api do
   end
 
   @doc """
-  Gets a user.
+  Gets a user by its `user_id`.
 
-  User to get is specified by `user_id`.
+  If the request is successful, this function returns `{:ok, user}`, where 
+  `user` is a `Nostrum.Struct.User`. Otherwise, returns `{:error, reason}`.
   """
-  @spec get_user(integer) :: error | {:ok, Nostrum.Sturct.User.t}
+  @spec get_user(User.id) :: error | {:ok, User.t}
   def get_user(user_id) do
-    request(:get, Constants.user(user_id))
+    case request(:get, Constants.user(user_id)) do 
+      {:ok, body} -> 
+        user =  
+          body 
+          |> Poison.decode!() 
+          |> User.to_struct() 
+ 
+        {:ok, user} 
+      other -> 
+        other 
+    end 
   end
 
-  @spec get_user!(integer) :: no_return | Nostrum.Struct.User.t
+  @doc """
+  Same as `get_user/1`, but raises `Nostrum.Error.ApiError` in case of failure.
+  """
+  @spec get_user!(User.id) :: no_return | User.t
   def get_user!(user_id) do
     get_user(user_id)
     |> bangify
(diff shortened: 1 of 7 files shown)


lib/nostrum/api.ex
```diff
@@ -1355,54 +1360,82 @@ defmodule Nostrum.Api do
   end
 
   @doc """
-  Gets a user.
+  Gets a user by its `user_id`.
```
Kraigie: Could we link this to the actual type in the `User` module?

lib/nostrum/util.ex
```diff
@@ -204,6 +204,39 @@ defmodule Nostrum.Util do
       String.to_atom(token)
   end
 
+   # Generic casting function
```
Kraigie: Don't think this should be indented a space.

lib/nostrum/struct/user.ex
```diff
@@ -7,8 +7,22 @@ defmodule Nostrum.Struct.User do
   A `member` has everything that a `user` has, but also additional information on a per guild basis. This includes things like a `nickname` and a list of `roles`.
   """
 
+  alias Nostrum.Struct.Snowflake
+  alias Nostrum.Util
+
+  defstruct [
```
Kraigie: This should go above types and was something I missed. Good catch.

lib/nostrum/api.ex
```diff
@@ -1813,4 +1846,15 @@ defmodule Nostrum.Api do
     Application.get_env(:nostrum, :token)
   end
 
+  defp handle_request_with_decode(response, type) 
+  defp handle_request_with_decode({:error, _} = error, _type), do: error 
+ 
+  defp handle_request_with_decode({:ok, body}, type) do 
+    convert =  
+      body 
+      |> Poison.decode!()
```
Kraigie: Can omit the `()` here

lib/nostrum/api.ex
```diff
@@ -1355,54 +1360,82 @@ defmodule Nostrum.Api do
...
-  def modify_current_user(options) do
-    case request(:patch, Constants.me, options) do
-      {:ok, body} ->
-        {:ok, Poison.decode!(body)}
-      other ->
-        other
-    end
+  @spec modify_current_user(keyword | map) :: error | {:ok, User.t}
+  def modify_current_user(params)
+  def modify_current_user(params) when is_list(params), do: modify_current_user(Map.new(params))
```
Kraigie: If we're going to support this here, we should do it elsewhere when applicable.

lib/nostrum/api.ex
```diff
@@ -1355,54 +1360,82 @@ defmodule Nostrum.Api do
...
-  **Example**
-  ```Elixir
-  avatar = %{avatar: "data:image/jpeg;base64," <> "YXl5IGJieSB1IGx1a2luIDQgc3VtIGZ1az8="}
-  {:ok, user} = Nostrum.Api.modify_current_user(avatar)
-  ```
+  ## Params 
+ 
+  The following params are optional: 
+ 
+    * `:username` (string) - new username
```
Kraigie: This kind of `param` documentation is missing on the other functions. We should discuss if we should include it everywhere, or only when accepting lists/maps and optional parameters. 

I'm leaning towards the latter, but the former could be helpful if the user isn't familiar with the API overall. But this could probably be handled by with types and such. Let me know what you think.

lib/nostrum/api.ex
```diff
@@ -1813,4 +1846,15 @@ defmodule Nostrum.Api do
     Application.get_env(:nostrum, :token)
   end
 
+  defp handle_request_with_decode(response, type)
```
Kraigie: I think this function name is a little verbose. For now it's fine as this is a WIP and something like `handle` is already in use. In the future we should maybe consider changing the name. 🙃

lib/nostrum/struct/user.ex
```diff
@@ -42,27 +56,18 @@ defmodule Nostrum.Struct.User do
...
   end
 
   @doc false
-  # TODO: Necessary?
-  def to_struct(%{__struct__: _} = map), do: map
   def to_struct(map) do
-    struct(__MODULE__, map)
+    new = 
+      map 
+      |> Map.new(fn {k, v} -> {Util.maybe_to_atom(k), v} end)
```
Kraigie: I'm probably missing something but what's the purpose of this?

lib/nostrum/util.ex
```diff
@@ -204,6 +204,39 @@ defmodule Nostrum.Util do
...
+    Map.new(new_list, fn item ->
+      {struct_get_in(item, index_by), item}
+    end)
+  end
+
+  def cast(value, {:struct, module}) when is_map(value) do
+    module.to_struct(value)
+  end
+
+  def cast(value, module) do
```
Kraigie: This really only applies to the `Snowflake` module currently, correct? Do you see a use case in the future for this? Probably if we need to implement another "custom" type?

lib/nostrum/api.ex
Skippe: I think it is a good idea to allow keyword lists or maps for these request params. We should support it. If it's fine with you, I'll be doing something similar to this for the rest of the API changes.

lib/nostrum/api.ex
Skippe: Agreed with the latter. Was planning to change the rest of the API functions to only include a "Request Params" section whenever lists/maps were needed.

lib/nostrum/api.ex
Skippe: I think verbose is okay for self-documentation purposes, but I am okay with a rename in the future. Renaming a function like this will not be hard.

lib/nostrum/struct/user.ex
Skippe: Line 68 converts all of the map's keys into atoms using `maybe_to_atom/1` from the Util module.

There is a function in Util that already does something like this (`safe_atom_map/1` off the top of my head), but I did not want to use it due to having a weird side effect. That function attempts to convert all map keys into atoms, including inner maps as well. I think `safe_atom_map/1` is a problem for `to_struct/1` functions, since we don't want to convert all of the inner map keys into atoms. For example, using `safe_atom_map/1` on a `Nostrum.Struct.Guild` struct will affect each of the "indexed" data fields in the struct. Off the top of my head, these are `:members`, `:channels`, and `:roles`. For these fields, `safe_atom_map/1` will attempt to convert their ids (4317843914731) into atoms.

Rather than change `safe_atom_map/1`, I'd just do what I did in line 68 for a "flat" atom-map function.

lib/nostrum/util.ex
Skippe: I did see a potential case for keeping this in the `Nostrum.Struct.Guild` module. That module has indexed data fields (:members, :channels, :roles) that could each be represented as an individual type. 

We can remove that code for sure, but I think it would be good to open up that conversion function for extension.

lib/nostrum/api.ex
Skippe: We can. Changes have been pushed to reflect this in the latest commit.

lib/nostrum/util.ex
Skippe: Good catch 👍 . Fixed in latest commit.

lib/nostrum/api.ex
Skippe: Fixed in latest commit.

lib/nostrum/api.ex
Kraigie: Yep, I'm fine with this approach.

lib/nostrum/struct/user.ex
Kraigie: Thanks for the explanation, no problem with this then.

lib/nostrum/util.ex
Kraigie: I see what you mean, we can leave it her for now then.

lib/nostrum/api.ex
```diff
@@ -1355,54 +1360,82 @@ defmodule Nostrum.Api do
   end
 
   @doc """
-  Gets a user.
+  Gets a `Nostrum.Struct.User` by its `user_id`.
```
Kraigie: For this I was actually referring to the `user_id` type portion.

lib/nostrum/api.ex
```diff
@@ -1355,54 +1360,82 @@ defmodule Nostrum.Api do
...
+    |> handle_request_with_decode({:struct, User})
+  end
+
+  @doc """
+  Same as `modify_current_user/3`, but raises `Nostrum.Error.ApiError` in case of failure.
+  """
+  @spec modify_current_user!(keyword | map) :: no_return | User.t
+  def modify_current_user!(params) do
+    modify_current_user(params)
+    |> bangify()
```
Kraigie: Another `()` we can omit.

lib/nostrum/struct/snowflake.ex
```diff
@@ -0,0 +1,62 @@
...
+
+  Snowflakes are 64-bit unsigned integers used to represent discord 
+  object ids. In JSON, they are typically represented as strings due 
+  to some languages not being able to represent such a large number.
+  
+  Nostrum represents snowflakes as integers since Elixir allows us 
+  to represent large integers.
+  """
+
+  @type external_snowflake :: String.t
```
Kraigie: I think I'd prefer we move some of the documentation in the `moduledoc` to `typedoc`s.

lib/nostrum/api.ex
Skippe: Would something like this be what we are looking for?

```Elixir
@doc """
Gets a user by its `t:Nostrum.Struct.User.id/0`.
"""
```

lib/nostrum/api.ex
Skippe: Omitted in latest commit

lib/nostrum/struct/snowflake.ex
Skippe: Latest commit has reflected these changes. They look weird to me, but it is likely because of how I refactored the documentation.

lib/nostrum/api.ex
Kraigie: Yeah that'd work, could we use the alias'd version of the User module? I'm not sure if it works for types. 

Also it'd be nice to add this elsewhere when applicable.

Skippe: Fixed in latest commits. `mix docs` wouldn't compile the documentation properly if I used an alias.

Kraigie:
Thanks!

PR #48, opened 2018-03-20
Api Correction Part 2: Emojis

Skippe:
This PR focuses on correcting the Api by cleaning up the pre-existing structs.

This PR does the following:

1. Cleans up the emoji struct by organizing it. Additionally, it adds more documentation to the Emoji module's elements.
2. Corrects specs in the Api module regarding Emojis.
3. Adds a couple of utility functions in the Emoji module for helping library users use Emojis with the Api.
4. Fixes the Emoji struct assembly process when reading outside data (converting snowflakes, other structs, etc.)

Notes:

1. I think `delete_reaction/4` in the Api does not have a matching name with the corresponding Discord Api endpoint (DELETE USER REACTION). This is a minor problem, but we could try renaming this function to match it with the endpoint name.
2. I added `to_api_name/1` and `to_markdown/1` in the Emoji module to help lib users use Emoji structs with the Api. I'm thinking of potentially renaming these to `get_api_name/1` and `format_markdown/1`. Up for discussion.
3. `format_custom_emoji/2` in the Emoji module does not seem to be useful from my perspective. Potential deprecation or removal?
4. I ended up renaming the `custom_emoji` type in favor of `emoji_api_name`. I think the latter is more representative of the type's purpose. It also lets us cover the api names of all emojis (not just custom ones).

TODO:

- Investigate URL encoding
- Reduce low-level binding of Emoji functions.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -285,28 +285,32 @@ defmodule Nostrum.Api do
   ## Examples
 
   ```Elixir
+  # Using a Nostrum.Struct.Emoji.
+  emoji = %Nostrum.Struct.Emoji{id: 43819043108, name: "foxbot"}
+  Nostrum.Api.create_reaction(123123123123, 321321321321, emoji)
+
   # Using a base 16 emoji string.
   Nostrum.Api.create_reaction(123123123123, 321321321321, "\xF0\x9F\x98\x81")
 
   # Using a URI encoded emoji string.
   Nostrum.Api.create_reaction(123123123123, 321321321321, URI.encode("\u2b50"))
-
-  # Using a Nostrum.Struct.Emoji.
-  emoji = %Nostrum.Struct.Emoji{id: 43819043108, name: "foxbot"}
-  Nostrum.Api.create_reaction(123123123123, 321321321321, Nostrum.Struct.Emoji.to_api_name(emoji))
   ```
 
   For other emoji string examples, see `t:Nostrum.Struct.Emoji.emoji_api_name/0`.
   """
-  @spec create_reaction(Channel.id, Message.id, Emoji.emoji_api_name) :: error | {:ok}
-  def create_reaction(channel_id, message_id, emoji) do
-    request(:put, Constants.channel_reaction_me(channel_id, message_id, emoji))
+  @spec create_reaction(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name) :: error | {:ok}
+  def create_reaction(channel_id, message_id, emoji)
+  def create_reaction(channel_id, message_id, %Emoji{} = emoji), 
+    do: create_reaction(channel_id, message_id, Emoji.get_api_name(emoji))
+
+  def create_reaction(channel_id, message_id, emoji_api_name) do
+    request(:put, Constants.channel_reaction_me(channel_id, message_id, emoji_api_name))
   end
 
   @doc """
   Same as `create_reaction/3`, but raises `Nostrum.Error.ApiError` in case of failure.
   """
-  @spec create_reaction!(Channel.id, Message.id, Emoji.emoji_api_name) :: no_return | {:ok}
+  @spec create_reaction!(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name) :: no_return | {:ok}
   def create_reaction!(channel_id, message_id, emoji) do
     create_reaction(channel_id, message_id, emoji)
     |> bangify
@@ -318,19 +322,21 @@ defmodule Nostrum.Api do
   If the request was successful, this function returns `{:ok}`. Otherwise, 
   this function returns `{:error, reason}`.
 
-  ## Examples
-
   See `create_reaction/3` for similar examples.
   """
-  @spec delete_own_reaction(Channel.id, Message.id, Emoji.emoji_api_name) :: error | {:ok}
-  def delete_own_reaction(channel_id, message_id, emoji) do
-    request(:delete, Constants.channel_reaction_me(channel_id, message_id, emoji))
+  @spec delete_own_reaction(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name) :: error | {:ok}
+  def delete_own_reaction(channel_id, message_id, emoji)
+  def delete_own_reaction(channel_id, message_id, %Emoji{} = emoji), 
+    do: delete_own_reaction(channel_id, message_id, Emoji.get_api_name(emoji))
+  
+  def delete_own_reaction(channel_id, message_id, emoji_api_name) do
+    request(:delete, Constants.channel_reaction_me(channel_id, message_id, emoji_api_name))
   end
 
   @doc ~S"""
   Same as `delete_own_reaction/3`, but raises `Nostrum.Error.ApiError` in case of failure.
   """
-  @spec delete_own_reaction!(Channel.id, Message.id, Emoji.emoji_api_name) :: no_return | {:ok}
+  @spec delete_own_reaction!(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name) :: no_return | {:ok}
   def delete_own_reaction!(channel_id, message_id, emoji) do
     delete_own_reaction(channel_id, message_id, emoji)
     |> bangify
@@ -342,19 +348,21 @@ defmodule Nostrum.Api do
   If the request was successful, this function returns `{:ok}`. Otherwise, 
   this function returns `{:error, reason}`.
 
-  ## Examples
-
   See `create_reaction/3` for similar examples.
   """
-  @spec delete_user_reaction(Channel.id, Message.id, Emoji.emoji_api_name, User.id) :: error | {:ok}
-  def delete_user_reaction(channel_id, message_id, emoji, user_id) do
-    request(:delete, Constants.channel_reaction(channel_id, message_id, emoji, user_id))
+  @spec delete_user_reaction(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name, User.id) :: error | {:ok}
+  def delete_user_reaction(channel_id, message_id, emoji, user_id)
+  def delete_user_reaction(channel_id, message_id, %Emoji{} = emoji, user_id), 
+    do: delete_user_reaction(channel_id, message_id, Emoji.get_api_name(emoji), user_id)
+
+  def delete_user_reaction(channel_id, message_id, emoji_api_name, user_id) do
+    request(:delete, Constants.channel_reaction(channel_id, message_id, emoji_api_name, user_id))
   end
 
   @doc ~S"""
   Same as `delete_user_reaction/4`, but raises `Nostrum.Error.ApiError` in case of failure.
   """
-  @spec delete_user_reaction!(Channel.id, Message.id, Emoji.emoji_api_name, User.id) :: no_return | {:ok}
+  @spec delete_user_reaction!(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name, User.id) :: no_return | {:ok}
   def delete_user_reaction!(channel_id, message_id, emoji, user_id) do
     delete_user_reaction(channel_id, message_id, emoji, user_id)
     |> bangify
@@ -367,20 +375,22 @@ defmodule Nostrum.Api do
   `users` is a list of `Nostrum.Struct.User`. Otherwise, this function 
   returns `{:error, reason}`.
 
-  ## Examples
-
   See `create_reaction/3` for similar examples.
   """
-  @spec get_reactions(Channel.id, Message.id, Emoji.emoji_api_name) :: error | {:ok, [User.t]}
-  def get_reactions(channel_id, message_id, emoji) do
-    request(:get, Constants.channel_reactions_get(channel_id, message_id, emoji))
+  @spec get_reactions(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name) :: error | {:ok, [User.t]}
+  def get_reactions(channel_id, message_id, emoji)
+  def get_reactions(channel_id, message_id, %Emoji{} = emoji), 
+    do: get_reactions(channel_id, message_id, Emoji.get_api_name(emoji))
+  
+  def get_reactions(channel_id, message_id, emoji_api_name) do
+    request(:get, Constants.channel_reactions_get(channel_id, message_id, emoji_api_name))
     |> handle_request_with_decode({:list, {:struct, User}})
   end
 
   @doc """
   Same as `get_reactions/3`, but raises `Nostrum.Error.ApiError` in case of failure.
   """
-  @spec get_reactions!(Channel.id, Message.id, Emoji.emoji_api_name) :: no_return | [User.t]
+  @spec get_reactions!(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name) :: no_return | [User.t]
   def get_reactions!(channel_id, message_id, emoji) do
     get_reactions(channel_id, message_id, emoji)
     |> bangify
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -346,17 +346,17 @@ defmodule Nostrum.Api do
 
   See `create_reaction/3` for similar examples.
   """
-  @spec delete_reaction(Channel.id, Message.id, Emoji.emoji_api_name, User.id) :: error | {:ok}
-  def delete_reaction(channel_id, message_id, emoji, user_id) do
+  @spec delete_user_reaction(Channel.id, Message.id, Emoji.emoji_api_name, User.id) :: error | {:ok}
+  def delete_user_reaction(channel_id, message_id, emoji, user_id) do
     request(:delete, Constants.channel_reaction(channel_id, message_id, emoji, user_id))
   end
 
   @doc ~S"""
-  Same as `delete_reaction/4`, but raises `Nostrum.Error.ApiError` in case of failure.
+  Same as `delete_user_reaction/4`, but raises `Nostrum.Error.ApiError` in case of failure.
   """
-  @spec delete_reaction!(Channel.id, Message.id, Emoji.emoji_api_name, User.id) :: no_return | {:ok}
-  def delete_reaction!(channel_id, message_id, emoji, user_id) do
-    delete_reaction(channel_id, message_id, emoji, user_id)
+  @spec delete_user_reaction!(Channel.id, Message.id, Emoji.emoji_api_name, User.id) :: no_return | {:ok}
+  def delete_user_reaction!(channel_id, message_id, emoji, user_id) do
+    delete_user_reaction(channel_id, message_id, emoji, user_id)
     |> bangify
   end
 
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -279,22 +279,26 @@ defmodule Nostrum.Api do
   @doc ~S"""
   Creates a reaction for a message.
 
-  Parameter `emoji` can be any of the following:
-
-    * A `t:Nostrum.Struct.Emoji.emoji_api_name/0`.
-    * A base 16 unicode emoji string.
-    * A URI encoded string.
-
   If the request was successful, this function returns `{:ok}`. Otherwise, 
   this function returns `{:error, reason}`.
 
   ## Examples
 
-      Nostrum.Api.create_reaction(123123123123, 321321321321, "\xF0\x9F\x98\x81")
+  ```Elixir
+  # Using a base 16 emoji string.
+  Nostrum.Api.create_reaction(123123123123, 321321321321, "\xF0\x9F\x98\x81")
 
-      Nostrum.Api.create_reaction(123123123123, 321321321321, URI.encode("\u2b50"))
+  # Using a URI encoded emoji string.
+  Nostrum.Api.create_reaction(123123123123, 321321321321, URI.encode("\u2b50"))
+
+  # Using a Nostrum.Struct.Emoji.
+  emoji = %Nostrum.Struct.Emoji{id: 43819043108, name: "foxbot"}
+  Nostrum.Api.create_reaction(123123123123, 321321321321, Nostrum.Struct.Emoji.to_api_name(emoji))
+  ```
+
+  For other emoji string examples, see `t:Nostrum.Struct.Emoji.emoji_api_name/0`.
   """
-  @spec create_reaction(Channel.id, Message.id, String.t | Emoji.emoji_api_name) :: error | {:ok}
+  @spec create_reaction(Channel.id, Message.id, Emoji.emoji_api_name) :: error | {:ok}
   def create_reaction(channel_id, message_id, emoji) do
     request(:put, Constants.channel_reaction_me(channel_id, message_id, emoji))
   end
@@ -302,7 +306,7 @@ defmodule Nostrum.Api do
   @doc """
   Same as `create_reaction/3`, but raises `Nostrum.Error.ApiError` in case of failure.
   """
-  @spec create_reaction!(Channel.id, Message.id, String.t | Emoji.emoji_api_name) :: no_return | {:ok}
+  @spec create_reaction!(Channel.id, Message.id, Emoji.emoji_api_name) :: no_return | {:ok}
   def create_reaction!(channel_id, message_id, emoji) do
     create_reaction(channel_id, message_id, emoji)
     |> bangify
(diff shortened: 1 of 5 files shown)


lib/nostrum/struct/emoji.ex
```diff
@@ -3,30 +3,111 @@ defmodule Nostrum.Struct.Emoji do
...
+  Formats an emoji struct into its respective markdown.
+
+  ## Examples
+
+      emoji = %Nostrum.Struct.Emoji{id: 43819043108, name: "foxbot"}
+      Nostrum.Struct.Emoji.to_markdown(emoji)
+      "<:foxbot:43819043108>"
+  """
+  @spec to_markdown(t) :: String.t
+  def to_markdown(emoji)
```
Kraigie: For these two functions, do we want to include a function definition that bypasses having to supply a struct?

I'm thinking just a 2 arity function of the same name that takes an `id` and a `name`. Or maybe a 3 arity with default animated set to `false`?

Maybe the use case is kind of niche? Let me know what you think.

lib/nostrum/struct/emoji.ex
```diff
@@ -3,30 +3,111 @@ defmodule Nostrum.Struct.Emoji do
...
+  @doc """
+  Formats an emoji struct into its respective markdown.
+
+  ## Examples
+
+      emoji = %Nostrum.Struct.Emoji{id: 43819043108, name: "foxbot"}
+      Nostrum.Struct.Emoji.to_markdown(emoji)
+      "<:foxbot:43819043108>"
+  """
+  @spec to_markdown(t) :: String.t
```
Kraigie: Correct me if I'm wrong, but the purpose of this here is for sending emojis in messages, yeah? It might be a good idea to state this or include an example of doing so.

As it stands I think `to_markdown` can be a little vague.

lib/nostrum/struct/emoji.ex
```diff
@@ -3,30 +3,111 @@ defmodule Nostrum.Struct.Emoji do
...
+  @type roles :: [Snowflake.t] | nil
 
-  Some API endpoints take an `emoji`.
-  The emoji should be structured as `"id:name"`.
-  """
-  @type custom_emoji :: String.t
+  @typedoc "User that created this emoji"
+  @type user :: User.t | nil
+
+  @typedoc "Whether this emoji must be wrapped in colons "
```
Kraigie: Extra space here at the end.

Kraigie:
```I think delete_reaction/4 in the Api does not have a matching name with the corresponding Discord Api endpoint (DELETE USER REACTION). This is a minor problem, but we could try renaming this function to match it with the endpoint name.```
Yep, feel free to rename it better match the endpoint as the rest of the API module does, good catch!

```I added to_api_name/1 and to_markdown/1 in the Emoji module to help lib users use Emoji structs with the Api. I'm thinking of potentially renaming these to get_api_name/1 and format_markdown/1. Up for discussion.```
I agree that the names could use work. I'm not set on the `name` portion of `get_api_name` as it isn't really a name is it, more of an identifier?

```format_custom_emoji/2 in the Emoji module does not seem to be useful from my perspective. Potential deprecation or removal?```
Yep, go ahead and axe it. We're still < v1.0 so who needs to deprecate. 🙃 

```I ended up renaming the custom_emoji type in favor of emoji_api_name. I think the latter is more representative of the type's purpose. It also lets us cover the api names of all emojis (not just custom ones).```
No problems with this, good change IMO.

File: lib/nostrum/api.ex [hunk 4]
```diff
@@ -276,59 +277,99 @@ defmodule Nostrum.Api do
...
 
-  **Example**
-  ```Elixir
-  Nostrum.Api.create_reaction(123123123123, 321321321321, "\xF0\x9F\x98\x81")
-  Nostrum.Api.create_reaction(123123123123, 321321321321, URI.encode("\u2b50"))
-  ```
+  ## Examples
 
-  Returns `{:ok}` if successful, `{:error, reason}` otherwise.
+      Nostrum.Api.create_reaction(123123123123, 321321321321, "\xF0\x9F\x98\x81")
```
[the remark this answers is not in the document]
Skippe: The examples were already code-blocked due to four-space indentation, but I've gone ahead and added some explicit triple-tick markdown for the examples.

I decided to go ahead and remove `iex> ` from the examples I've pushed. IIRC, `iex>` should only be used whenever a function can be tested using ExUnit. Source: https://hexdocs.pm/elixir/master/writing-documentation.html#doctests

lib/nostrum/api.ex
```diff
@@ -276,59 +277,99 @@ defmodule Nostrum.Api do
...
+
+      Nostrum.Api.create_reaction(123123123123, 321321321321, URI.encode("\u2b50"))
```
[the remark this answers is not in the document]
Skippe: I've added an example with `Emoji.to_api_name/1`, as well as made the other emoji functions point to `create_reaction/3` for examples. Let me know how it looks.

lib/nostrum/struct/emoji.ex
```diff
@@ -3,30 +3,111 @@ defmodule Nostrum.Struct.Emoji do
...
+  @spec to_markdown(t) :: String.t
+  def to_markdown(emoji)
+  def to_markdown(%__MODULE__{id: nil, name: name}), do: name
+  def to_markdown(%__MODULE__{animated: true, id: id, name: name}), do: "<a:#{name}:#{id}>"
+  def to_markdown(%__MODULE__{id: id, name: name}), do: "<:#{name}:#{id}>"
+
+  @doc """
+  Formats an emoji struct into its `t:Nostrum.Struct.Emoji.emoji_api_name/0`.
+
+  ## Examples
```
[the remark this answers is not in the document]
Skippe: Yeah I think it'd be useful. Added one with latest commit.

lib/nostrum/struct/emoji.ex
Skippe: Seems niche IMO. Even if I were to add it, I'd prefer to create a new module for that. Functions in `Nostrum.Struct.Emoji` should explicitly deal with the Emoji struct and nothing else IMO. However, we could add this for backwards compatibility. I'll think about it.

With that being said, I'll probably allow maps (in addition to the Emoji struct) to work with this function. Just now, I realized that parts of the library aren't properly decoded to their specific structs. Functions like these will break if I don't allow maps to work.

lib/nostrum/struct/emoji.ex
Skippe: I'll add an example in this upcoming commit.

I think `to_markdown` is vague as well. Maybe `format` would be a better name? Let me know your thoughts. Also, I was actually thinking about having the `String.Chars` protocol handle the markdown conversion. This would allow the emojis to be used in strings in a more natural method: 

```Elixir
emoji = %Nostrum.Struct.Emoji{id: 4314314, name: "craigpls"}

Kernel.to_string(emoji)
"<:craigpls:4314314>"

"I get to use this emoji now #{emoji}"
"I get to use this emoji now <:craigpls:4314314>"
```

And then we can always leave a markdown conversion function in case the user wanted a more explicit function call:

```Elixir
emoji = %Nostrum.Struct.Emoji{id: 4314314, name: "craigpls"}

md = Nostrum.Struct.Emoji.format(emoji)
"<:craigpls:4314314>"

"Got some new text #{md}"
"Got some new text <:craigpls:4314314>"
```

Let me know your thoughts.

lib/nostrum/struct/emoji.ex
Skippe: Fixed in latest commit.

Skippe:
> Yep, feel free to rename it better match the endpoint as the rest of the API module does, good catch!

Will rename it in this upcoming commit.

> I agree that the names could use work. I'm not set on the name portion of get_api_name as it isn't really a name is it, more of an identifier?

I tried using a name like `emoji_api_id`, but I couldn't settle on a good name. `emoji_api_name` was the one that I thought would work the best. If we can think of a better name, I'm open ears. In any case, I think `emoji_api_name` is better than the `custom_emoji` name we had earlier.

> Yep, go ahead and axe it. We're still < v1.0 so who needs to deprecate. 🙃

It's axing time :^).

> No problems with this, good change IMO.

lib/nostrum/api.ex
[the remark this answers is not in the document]
Kraigie: I actually didn't know about the four-space indentation. Do they auto highlight to elixir syntax? If so then it's not a big deal going forward. 

And oh yeah, duh, we didn't have any doctests prior and I forgot about them. Given that is the case, I think it should be removed from `modify_current_user/1`.

lib/nostrum/struct/emoji.ex
Kraigie: If the payload isn't being decoded properly (which is the case a lot of the time, probably), then that is an issue and I don't think we should support that. I partly agree with your idea that we should keep it limited to functions dealing with Emoji structs. If we're going down this route we should stick to it and not allow maps.

I could see users already knowing the id and name and wanting to bypass having to create a struct, but in that case I guess they can just form their own string?

lib/nostrum/struct/emoji.ex
Kraigie: I think `format` or `format_emoji` would be good choices.

I like the protocol idea, lets go with that! Just make sure its documented properly.

lib/nostrum/api.ex
```diff
@@ -310,10 +320,7 @@ defmodule Nostrum.Api do
 
   ## Examples
 
-      Nostrum.Api.delete_own_reaction(123123123123, 321321321321, "\xF0\x9F\x98\x81")
-      
-      Nostrum.Api.delete_own_reaction(123123123123, 321321321321, URI.encode("\u2b50"))
-      
+  See `create_reaction/3` for similar examples.
```
Kraigie: I don't think we need the `Examples` sub-header if we're going to point to another function like this.

lib/nostrum/struct/emoji.ex
Skippe: I'll revert the map changes then.

For the idea you proposed with having an `id` and `name`, I have a solution, but it might be hard to implement right now. It also suggests having a rather strong reliance on using structs in the project:

In my current project, I currently bind emoji markdown to function names for ease of access:

```Elixir
defmodule Emojis do
  def unranked, do: "<:unranked:425713028852350976>"
  def silver, do: "<:silver:425713020862332930>"
end

# Example

"This text will be sent with this emoji: #{Emojis.silver()}"
"This text will be sent with this emoji: <:silver:425713020862332930>"
```

I was thinking in the future, we could have something like an `EmojiCache` to select emojis. Assume `EmojiCache.get/1` accepts a keyword list and returns a `Nostrum.Struct.Emoji` or `nil`:

```Elixir
defmodule Emojis do
  alias Nostrum.Cache.EmojiCache

  def unranked, do: EmojiCache.get(name: "unranked", id: 425713028852350976)
  def silver, do: EmojiCache.get(name: "silver", id: 425713020862332930)
end

# Example

Emojis.silver()
%Nostrum.Struct.Emoji{
  id: 425713020862332930,
  name: "silver",
  roles: []
}

# Assuming `Nostrum.Struct.Emoji` will implement the `String.Chars` protocol, 
# this will be possible:

"This text will be sent with this emoji: #{Emojis.silver()}"
"This text will be sent with this emoji: <:silver:425713020862332930>"

# As well as pulling information wherever necessary:

Emojis.silver().name
"silver"

Emojis.silver().roles
[]
```

Having an `EmojiCache` would also open users up to error handling in case an emoji is not there:

```Elixir
alias Nostrum.Cache.EmojiCache

case EmojiCache.get(id: 4839104319) do
  nil -> "\:silver\:"  # In case the emoji got deleted
  emoji -> emoji
end
```

While this next positive might be beyond the scope of nostrum, we could also open up this `EmojiCache` to fetch an emoji via markdown:

```Elixir
"This text will be sent with this emoji: <:silver:425713020862332930>"
|> String.split(" ")
|> Enum.each(fn token -> 
  case token do
    "<:" <> _rest = emoji_token -> 
      EmojiCache.get(markdown: emoji_token)
    _ ->
      :noop
  end
end)
```

Let me know your thoughts. For the most part, I think this explains how I was thinking of going about doing this.

lib/nostrum/struct/emoji.ex
Skippe: This next commit will reflect these changes.

lib/nostrum/api.ex
Skippe: Removed in latest commit.

Skippe:
I'll be changing the PR to [WIP] for now. I realized that there are a few Emoji endpoints that haven't been added to the API. In particular:

* List Guild Emojis
* Get Guild Emoji
* Create Guild Emoji
* Modify Guild Emoji
* Delete Guild Emoji

I'm going to do those and comment back here when I get those done.

Skippe:
Okay I've added a couple of things now with this last push:

* Added the following endpoints: LIST_GUILD_EMOJI, GET_GUILD_EMOJI, CREATE_GUILD_EMOJI, MODIFY_GUILD_EMOJI, DELETE_GUILD_EMOJI
* Allowed the Reaction endpoints to accept a `Nostrum.Struct.Emoji` in addition to an Emoji's api name.
* Added some documentation regarding the `String.Chars` protocol for Emojis.
* Renamed `to_markdown/1` to `format_mention/1`
* Renamed `to_api_name/1` to `get_api_name/1`

Notes:
1. `create_guild_emoji/2` made me realize that we should likely provide a way to build data URIs for images. Image data URIs being of form:

```
"data:image/png;base64,YXl5IGJieSB1IGx1a2luIDQgc3VtIGZ1az8="
```

There's nothing in the normal Elixir library that allows users to build these from an image. Might be a good idea to not require users to do this. Here is how I've been building them to test:

```Elixir
bytes = File.read!("/dir/here/image.png")
base_encode = Base.encode64(bytes)
data_uri = "data:image/png;base64," <> base_encode

Nostrum.Api.create_guild_emoji(431890431431, name: "emoji", image: data_uri)
```

lib/nostrum/struct/emoji.ex
Kraigie: I don't think this is a bad idea. 

My issue is how we go about this with other modules. Take a module which has permissions for example. Undoubtedly we'll have some way of interacting with those, likely in a way in which supplying a struct wouldn't make sense. In that case, I don't think a similar approach would work.

What do you think?

lib/nostrum/struct/emoji.ex
```diff
@@ -1,50 +1,170 @@
...
+      |> Map.new(fn {k, v} -> {Util.maybe_to_atom(k), v} end)
+      |> Map.update(:id, nil, &Util.cast(&1, Snowflake))
+      |> Map.update(:roles, nil, &Util.cast(&1, {:list, Snowflake}))
+      |> Map.update(:user, nil, &Util.cast(&1, {:struct, User}))
+
+    struct(__MODULE__, new)
+  end
+end
+
+defimpl String.Chars, for: Nostrum.Struct.Emoji do
```
Kraigie: I think we should move this into the module definition.

lib/nostrum/api.ex
```diff
@@ -778,6 +831,149 @@ defmodule Nostrum.Api do
...
+  ## Options 
+ 
+    * `:name` (string) - name of the emoji
+    * `:image` (base64 data URI) - the 128x128 emoji image. Maximum size of 256kb
+    * `:roles` (list of `t:Nostrum.Struct.Snowflake.t/0`) - roles for which this emoji will be whitelisted
+    (default: [])
+
+  `:name` and `:image` are always required.
+
+  ## Return Values
```
Kraigie: I've noticed that when there are options like you've included this sub-section. I don't think I'm a big fan of having it be its own section, but I'm open for discussion. We should decide how we want to display this, and make sure it's homogeneous throughout the module.

The elixir docs are kind of unorganized in how they do this, probably because their functions are a lot less linear than ours.

lib/nostrum/api.ex
```diff
@@ -276,49 +277,95 @@ defmodule Nostrum.Api do
...
   Nostrum.Api.create_reaction(123123123123, 321321321321, URI.encode("\u2b50"))
   ```
 
-  Returns `{:ok}` if successful, `{:error, reason}` otherwise.
+  For other emoji string examples, see `t:Nostrum.Struct.Emoji.emoji_api_name/0`.
   """
-  @spec create_reaction(integer, integer, String.t | Nostrum.Struct.Emoji.custom_emoji) :: error | {:ok}
-  def create_reaction(channel_id, message_id, emoji) do
-    request(:put, Constants.channel_reaction_me(channel_id, message_id, emoji))
+  @spec create_reaction(Channel.id, Message.id, Emoji.t | Emoji.emoji_api_name) :: error | {:ok}
```
Kraigie: Maybe we should include an `emoji` type at the top of this module that is comprised of this definition here? Namely the `Emoji.t | Emoji.emoji_api_name` portion.

lib/nostrum/api.ex
```diff
@@ -778,6 +831,149 @@ defmodule Nostrum.Api do
...
+
+  ## Examples
+
+  ```Elixir
+  image = "data:image/png;base64,YXl5IGJieSB1IGx1a2luIDQgc3VtIGZ1az8="
+
+  Nostrum.Api.create_guild_emoji(43189401384091, name: "nostrum", image: image, roles: [])
+  ```
+  """
+  @spec create_guild_emoji(Guild.id, keyword | map) :: error | {:ok, Emoji.t}
```
Kraigie: Hmm, throughout this module we'll have `keyward | map` in specs whenever there are optional parameters. Do you think we should similarly include something like an `optional` type that describes the map/keyword list functionality we have here?

If we decide to do something like this feel free to open an issue so we can keep track of it. We probably don't want to go back to the `user` functions in this pr and change them.

Kraigie:
I'm not sure I'm keen on including the functionality you described. Our API is relatively lower level as it stands, and this I think might be a little out of place? I could maybe see the argument for the inclusion of the data URI information (with some work), but even then having an example may be sufficient, maybe a link to some proper docs on the subject?

Skippe:
Now that I think about it, I think providing examples on data URI related functions will already give the library user an idea of what to work with. I'd say the examples give enough information for the user to google up if they are struggling. We probably don't need to include something like this in nostrum.

lib/nostrum/api.ex
Skippe: Agreed. I'll be adding that in this upcoming commit.

lib/nostrum/struct/emoji.ex
Skippe: If I'm understanding correctly, I do not see this a problem. Discord permissions are usually bound to some kind of discord object (channel, role, overwrite, etc). I don't think there would be a scenario where we would have a cache that shares individual permissions. I may need more context.

lib/nostrum/struct/emoji.ex
Skippe: Agreed. Added in the upcoming commit.

lib/nostrum/api.ex
Skippe: I'll create an issue specifically for tracking these. However, I'll be changing the `keyword | map` types for the current Emoji-related endpoints.

lib/nostrum/api.ex
Kraigie: Do you think it'd be okay to move under the original doc definition? I just feel it's out of place since every method has results (while not every method has optional params), but the section is only included on those methods that have optional params. I feel like that could be confusing?

A lot of times I point to the elixir docs just to see how they do it. [Here](https://hexdocs.pm/elixir/Enum.html#flat_map/2) for example they fit it in pretty smoothly.
While at the same time from the last pr we have [places](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/api.ex#L1409) that miss returning the value altogether.

I think it might be worth it to hash out a design or some sort of template (maybe through the discord channel for quicker iteration) we could use to kind of bring all of these docs in line. And then going forward we can follow that and I'll go through and fix up the previous PR if need be.

lib/nostrum/struct/emoji.ex
Kraigie: I wasn't thinking of needed a cache for the permissions. I meant to say that we'll probably want some helper methods around permissions, and I could see those taking in some extraneous parameters? I guess in that case you'd still want the discord object struct to perform the operation on, so it'd be okay.

I guess I can't think of anything off the top of my head as an example. In the case of permissions we might just want a different module that handles any logical operations on the subject? We can handle this when we get there.

Skippe: Yeah, I had the same idea with the permissions. Ideally, we'd create a module for the permissions that has functions for working with them. This would be similar to the `Nostrum.Struct.Snowflake` module.

Skippe:
After our discord discussion with how the documentation should be handled, I've updated the documentation to all functions relating to this specific PR. We will have to go back sometime to update the functions we didn't get to.

I also added in another banged function. `delete_all_reactions/2` and `delete_all_reactions!/2`  went past my review for some reason.

lib/nostrum/api.ex
```diff
@@ -276,86 +290,155 @@ defmodule Nostrum.Api do
...
-  Create a rection for a message.
+  Creates a reaction for a message.
 
-  Creates a reaction using an `emoji` for the message specified by `message_id` and
-  `channel_id`. `emoji` can be a `Nostrum.Struct.Emoji.custom_emoji.t`, a
-  base 16 unicode emoji string, or a uri encoded string.
+  This endpoint requires the `VIEW_CHANNEL` and `READ_MESSAGE_HISTORY` 
+  permissions. Additionally, if nobody else has reacted to the message with 
+  the `emoji`, this endpoint requires the `ADD_REACTIONS` permission. It 
+  fires the `MESSAGE_REACTION_ADD` event.
```
Kraigie: When talking about what events are fired, do you think we should link to the event in the `Consumer` module? Their type is listed there.

lib/nostrum/api.ex
```diff
@@ -276,86 +290,155 @@ defmodule Nostrum.Api do
...
   ```Elixir
+  # Using a Nostrum.Struct.Emoji.
+  emoji = %Nostrum.Struct.Emoji{id: 43819043108, name: "foxbot"}
+  Nostrum.Api.create_reaction(123123123123, 321321321321, emoji)
+
+  # Using a base 16 emoji string.
   Nostrum.Api.create_reaction(123123123123, 321321321321, "\xF0\x9F\x98\x81")
+
+  # Using a URI encoded emoji string.
   Nostrum.Api.create_reaction(123123123123, 321321321321, URI.encode("\u2b50"))
```
Kraigie: I've made encoding the URL the default behavior, so this is no longer necessary (and actually wasn't in the first place).

lib/nostrum/struct/emoji.ex
```diff
@@ -1,50 +1,166 @@
...
+  ## Examples
+
+  ```Elixir
+  # Custom Emojis
+  "nostrum:431890438091489"
+
+  # Unicode Emojis
+  "≡ƒæì"
+  "\xF0\x9F\x98\x81"
+  URI.encode("\u2b50")
```
Kraigie: We don't need the `URI.encode` here anymore, but I think it's a good example of using codepoints w/ the sigil so I think the rest of the example can stay here.

lib/nostrum/api.ex
Skippe: For now, I think this is a good idea. Going to put these in the next commit. I said for now, because I remembered you mentioned something about removing the Consumer module.

lib/nostrum/api.ex
Skippe: Fixed in latest commit.

lib/nostrum/struct/emoji.ex
Skippe: Removed in latest commit.

lib/nostrum/struct/emoji.ex
```diff
@@ -61,7 +60,6 @@ defmodule Nostrum.Struct.Emoji do
   # Unicode Emojis
   "≡ƒæì"
   "\xF0\x9F\x98\x81"
-  URI.encode("\u2b50")
```
Kraigie: Sorry if I wasn't clear. Could we still have the example here just without the `URI.encode` function call?

Skippe: Oops. Fixed in latest commit.

Skippe:
Rebased and squashed :^)

Kraigie:
🎉 🎆 🎉 🎆

Kraigie:
❤️ 💚 💙 💛 💜

PR #54, opened 2018-04-07
Cleanup for Embeds and its components

skippi:
This PR makes the following changes:

1. Updates the typespecs for Embed fields and its components.
2. Updates the `to_struct/1` functions for each Embed and its components.
3. Implements the Poison.Encoder protocol for each Embed and its components.

### DIFF
--- a/lib/nostrum/struct/embed.ex
+++ b/lib/nostrum/struct/embed.ex
@@ -16,12 +16,11 @@ defmodule Nostrum.Struct.Embed do
     |> put_url("https://google.com/")
     |> put_timestamp("2016-05-05T21:04:13.203Z")
     |> put_color(431_948)
-    |> add_field("Field 1", "Test")
-    |> add_field("Field 2", "More test", true)
+    |> put_field("Field 1", "Test")
+    |> put_field("Field 2", "More test", true)
   ```
 
-  Alternatively, it is possible to build `Nostrum.Struct.Embed`s using standard map syntax.
-  However, we recommend sticking to the aforementioned builder functions.
+  Alternatively, `Nostrum.Struct.Embed`s can be built using standard map syntax.
 
   ```Elixir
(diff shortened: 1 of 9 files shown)


skippi:
UPDATE: We've decided to expand this PR to cover these topics:

* Added builder functions for simplifying the embed building process.
* Added tests to test these builder functions.

lib/nostrum/struct/embed.ex
```diff
@@ -1,6 +1,41 @@
...
+    |> put_description("nostrum")
+    |> put_url("https://google.com/")
+    |> put_timestamp("2016-05-05T21:04:13.203Z")
+    |> put_color(431_948)
+    |> add_field("Field 1", "Test")
+    |> add_field("Field 2", "More test", true)
+  ```
+
+  Alternatively, it is possible to build `Nostrum.Struct.Embed`s using standard map syntax.
+  However, we recommend sticking to the aforementioned "builder functions".
```
z64: I don't really like leaving anecdotal comments like this unless you actually explain why, as the reader is going to be asking themselves this and left to guess exactly what you mean. I would expand on this or remove it.

lib/nostrum/struct/embed.ex
```diff
@@ -87,7 +122,146 @@ defmodule Nostrum.Struct.Embed do
...
+    }
+
+    %__MODULE__{embed | author: author}
+  end
+
+  @doc ~S"""
+  Adds a `Nostrum.Struct.Embed.Field` under `:fields` in `embed`.
+  """
+  @spec add_field(t, Field.name(), Field.value(), Field.inline()) :: t
+  def add_field(embed, name, value, inline \\ nil)
```
z64: I wonder about the verbiage here. Would a consistency of `put_field` be preferred? I know that internally this method functions "differently", but that is an implementation detail

Kraigie:
Thanks!

PR #55, opened 2018-04-10
Added `timestamp/1` to Snowflake

skippi:
A timestamp function should be included as a Snowflake utility function. 

While I was working on this function, I've also gone ahead and cleaned up the Snowflake tests.


lib/nostrum/struct/snowflake.ex
```diff
@@ -62,4 +62,16 @@ defmodule Nostrum.Struct.Snowflake do
   @spec dump(t) :: external_snowflake
```
Kraigie: Sorry for not catching this before, but shouldn't this spec also have a `no_return` clause? Also, thoughts on `!`ing this method since it can throw an error?

skippi: I've strengthened the `is_snowflake/1` guard to make sure `creation_time/1` always works under the correct conditions.

skippi:
Since this is doing work in the Snowflake module, I've gone ahead and fixed some inconsistencies in the module regarding specs and types.

PR #56, opened 2018-04-12
Fixing API and structs for Messages

skippi:
Changelist (ONGOING):

* Added Activity and Application structs for the Message struct.
* Added missing fields and types from Message.
* Fixed `to_struct/1` in Message.
* Reworked `create_message/3` implementation.


skippi:
I think we should talk about how to change the `create_message` to be more flexible. Currently, we have a three arity function with varying signatures that can be used to create messages:

```elixir
Nostrum.Api.create_message(351194183568195585, "This is only content", false)

Nostrum.Api.create_message(351194183568195585, [file_name: "This is a file and content", file: "/file/path/file.txt"], false)

Nostrum.Api.create_message(351194183568195585, [content: "This is an embed and content", embed: %{title: "embed", description: "desc"}], false)
```

We could change this to a two arity function using a kw list or map to pass in args. This would more closely make it more flexible at the cost of a longer function call:

```elixir
Nostrum.Api.create_message(351194183568195585, content: "This is only content")

Nostrum.Api.create_message(351194183568195585, content: "This is file and content.", file: "/path/to/file.txt")

Nostrum.Api.create_message(351194183568195585, content: "This is embed and content.", embed: %{title: "embed", description: "desc"})

import Nostrum.Struct.Embed
embed = 
  %Nostrum.Struct.Embed{}
  |> put_title("embed")
  |> put_description("desc")

Nostrum.Api.create_message(351194183568195585, content: "This is file, embed, and content.", embed: embed, file: "/path/to/file.txt")
```

If anyone has a better idea on how to make the function more flexible, let me know. I'll be working on other stuff in the meantime.

lib/nostrum/api.ex
```diff
@@ -138,51 +138,108 @@ defmodule Nostrum.Api do
...
+  def create_message!(channel_id, options) do
+    create_message(channel_id, options)
+    |> bangify
+  end
+
+  @doc ~S"""
+  Same as `create_message/2`, but takes a `Nostrum.Struct.Message` instead of a channel_id.
+  """
+  @spec reply_to_message(Message.t(), options) :: error | {:ok, Message.t()}
+  def reply_to_message(%Message{} = message, options) do
```
Kraigie: I don't think I like having a separate method for passing in a message struct. I can understand it's probably this way since right now we're guarding against the list and we don't want to have to deal with the combinations. However, given the choice I think it's probably better to match on the message struct and handle the list conversion inside the function.

lib/nostrum/api.ex
```diff
@@ -138,51 +138,108 @@ defmodule Nostrum.Api do
...
+  def create_message(channel_id, options) when is_list(options),
+    do: create_message(channel_id, Map.new(options))
 
-  # Sending regular messages
-  def create_message(channel_id, content, tts) when is_binary(content),
-    do: do_create_message(channel_id, %{content: content, tts: tts})
+  def create_message(channel_id, options) when is_snowflake(channel_id) do
+    case options do
+      %{file: _} -> create_message_with_multipart(channel_id, options)
+      %{} -> create_message_with_json(channel_id, options)
```
Kraigie: Do you think it would be better to guard against `options` being a list or map? If the user passes in something garbage here for `options` I think it'd be better to say there's no matching function, as opposed to no match for the case which the user doesn't really need to worry about. We can just make this second statement match on `_`. If they send in junk within the enumerable they'll get an API error which I think is fine. Let me know your thoughts.

lib/nostrum/api.ex
```diff
@@ -136,158 +138,215 @@ defmodule Nostrum.Api do
...
+    |> put_title("embed")
+    |> put_description("new desc")
+  Nostrum.Api.create_message(43189401384091, embed: embed)
+
+  Nostrum.Api.create_message(43189401384091, file: "/path/to/file.txt")
+
+  Nostrum.Api.create_message(43189401384091, content: "hello world!", embed: embed, file: "/path/to/file.txt")
   ```
   """
-  @spec create_message(Message.t(), message_content, boolean) :: error | {:ok, Message.t()}
```
Kraigie: We should remove the `message_content` type at the top of this file as it won't be used anymore.

Kraigie:
Thanks!

PR #58, opened 2018-04-18
HTTP API Cleanup for Member struct

skippi:
Similar to the other PRs. Name changes, typespec cleanups, struct updates, etc.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -1253,21 +1253,37 @@ defmodule Nostrum.Api do
   end
 
   @doc """
-  Gets a list of guild members.
+  Gets a list of a guild's members.
 
-  ## Parameters
-  `guild_id` - Guild to get members.
-  `options` - Map with the following *optional* keys:
-    - `limit` - Max number of members to return (1-1000)
-    - `after` - Highest user id of the previous page.
+  ## Options
+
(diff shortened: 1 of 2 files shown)


skippi:
Think this PR is ready for code review. Was planning to get this done tomorrow, but looks like I got it done earlier than expected.

lib/nostrum/api.ex
```diff
@@ -1229,89 +1229,163 @@ defmodule Nostrum.Api do
...
-          limit: 1..1000,
-          after: integer
-        }) :: error | {:ok, [Nostrum.Struct.Guild.Member.t()]}
-  def get_guild_members(guild_id, options) do
+  If successful, returns `{:ok, members}`. Otherwise, returns a `t:Nostrum.Api.error/0`.
+
+  ## Options
+
+    * `:limit` (integer) - max number of members to return (1-1000) (default: 1)
+    * `:after` (`t:Nostrum.Struct.Snowflake.t/0`) - the highest user id in the previous page (default: 0)
```
Kraigie: I saw you switched this from `User.id()` to `Snowflake.t`, what was the reasoning behind this?

skippi: I was experimenting with the types a little bit to see which one would be more informative to our users.  Looks like I accidentally sent this change with my self code-review. I've fixed it with the recent commit.

Kraigie:
Thanks!

PR #60, opened 2018-04-21
Fast Forwarding Previous Struct Work

skippi:
This PR focuses on fixing struct-specific features that should have been or should now be changed due to decisions we've made in the most recent PRs. Many of these include corrections to fix inconsistencies. Pretty much a fast-forward PR. This includes:

* Adding mentions to specific structs.
* Improving documentation to match our current standard.
* Improving typespecs where possible.
* Improving function names where possible.
* Creating and improving tests and doctests.

I'll hit API specific features in a future PR.

### DIFF
--- a/test/nostrum/struct/user_test.exs
+++ b/test/nostrum/struct/user_test.exs
@@ -7,7 +7,7 @@ defmodule Nostrum.Struct.UserTest do
 
   describe "String.Chars" do
     test "matches `mention/1`" do
-      user = %User{id: 150061853001777154}
+      user = %User{id: 150_061_853_001_777_154}
 
       assert(to_string(user) === User.mention(user))
     end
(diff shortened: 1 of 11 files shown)


skippi:
I think this PR is done, but let me know if it needs to be broken up. It is a pretty big PR that covers a lot of different changes.

test/nostrum/struct/user_test.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Nostrum.Struct.UserTest do
+  use ExUnit.Case, async: true
+
+  alias Nostrum.Struct.User
+
+  doctest User
+
+  describe "String.Chars" do
```
Kraigie: I think it's a little redundant to have a doctest and an actual test. What do you think? This is the case throughout this pr. Maybe it doesn't matter?

skippi: They aren't redundant, but I do wish I could have made them doctests. They test functionality for the `String.Chars` protocol, but the protocols aren't directly doctestable IIRC. If there is a way to make them doctestable, I would be on board.

Kraigie:
Thanks!

PR #61, opened 2018-04-21
HTTP API Cleanup for Channel struct

skippi:
Fixes mainly for updating the struct, updating documentation, and fixing decoding errors.


skippi:
Since the channel struct is a bit complex, I've tried using a different typespec format to describe `t:Nostrum.Struct.Guild.Channel.t/0`. Would like some feedback on that, as I think it is promising, but needs some work.

skippi:
I think this PR is ready for code-review. As a disclaimer, I've deleted DMChannel and moved Channel under `Nostrum.Struct`.

My only toss-ups are typedocs for each channel type. Don't know how we should go about doing that.

lib/nostrum/cache/channel_cache.ex
```diff
@@ -6,11 +6,10 @@ defmodule Nostrum.Cache.ChannelCache do
   use GenServer
 
   alias Nostrum.Cache.Guild.GuildServer
-  alias Nostrum.Struct.Guild.Channel
-  alias Nostrum.Struct.DMChannel
+  alias Nostrum.Struct.Channel
   alias Nostrum.Util
 
-  @type channel :: Channel.t() | DMChannel.t()
+  @type channel :: Channel.dm_channel()
```
Kraigie: I don't think this is correct. In `get/1` for example, the spec states we return this type when it could be any time of channel. I think this should be removed and the references to a `channel` type in this module should instead use the struct `t/0` type or some other appropriate type.

lib/nostrum/cache/channel_cache.ex
```diff
@@ -24,8 +23,9 @@ defmodule Nostrum.Cache.ChannelCache do
   @doc ~S"""
   Retrieves a channel from the cache.
 
-  Internally, the ChannelCache process only stores DMChannel references. To get
-  channel information, a call is made to a `Nostrum.Cache.Guild.GuildServer`.
+  Internally, the ChannelCache process only stores
```
Kraigie: Should probably link to channel struct's `dm_channel` type.

lib/nostrum/struct/channel.ex
```diff
@@ -0,0 +1,266 @@
...
+  iex> channel = %Nostrum.Struct.Channel{id: 381889573426429952}
+  ...> Nostrum.Struct.Channel.mention(channel)
+  "<#381889573426429952>"
+  ```
+  """
+  @spec mention(t) :: String.t()
+  def mention(%__MODULE__{id: id}), do: "<##{id}>"
+
+  @doc false
+  def p_encode do
```
Kraigie: We shouldn't need this function anymore, but if you want to leave it for now that's fine as that's what we've done in other modules.

lib/nostrum/struct/channel.ex
```diff
@@ -0,0 +1,266 @@
...
+  @typedoc "The application id of the group DM creator if it is bot-created"
+  @type application_id :: Snowflake.t() | nil
+
+  @typedoc "The id of the parent category for a channel"
+  @type parent_id :: Snowflake.t() | nil
+
+  @typedoc "When the last pinned message was pinned"
+  @type last_pin_timestamp :: String.t() | nil
+
+  @type guild_text_channel :: %__MODULE__{
```
Kraigie: Do you think it would be useful to include documentation for these types and the type "groups" down below?

lib/nostrum/struct/channel.ex
```diff
@@ -0,0 +1,266 @@
...
+
+  ## Examples
+
+  ```Elixir
+  iex> channel = %Nostrum.Struct.Channel{id: 381889573426429952}
+  ...> Nostrum.Struct.Channel.mention(channel)
+  "<#381889573426429952>"
+  ```
+  """
+  @spec mention(t) :: String.t()
```
Kraigie: Can you mention things like channel categories, which are included in the `t/0` spec? If not then this should be changed to be more accurate.

lib/nostrum/api.ex
```diff
@@ -486,107 +487,119 @@ defmodule Nostrum.Api do
...
+    * `:position` (integer) - the position of the channel in the left-hand listing
+    * `:topic` (string) (`t:Nostrum.Struct.Channel.text_channel/0` only) -
+    0-1024 character channel topic
+    * `:nsfw` (boolean) (`t:Nostrum.Struct.Channel.text_channel/0` only) -
+    if the channel is nsfw
+    * `:bitrate` (integer) (`t:Nostrum.Struct.Channel.voice_channel/0` only) -
+    the bitrate (in bits) of the voice channel; 8000 to 96000 (128000 for VIP servers)
+    * `:user_limit` (integer) (`t:Nostrum.Struct.Channel.voice_channel/0` only) -
+    the user limit of the voice channel; 0 refers to no limit, 1 to 99 refers to a user limit
+    * `:permission_overwrites` (list of `t:Nostrum.Struct.Overwrite.t/0`) - channel
```
Kraigie: I think it's a little weird to say this has to be a struct, but I think it's fine as a general precaution. Maybe we could modify it to be something like `list of 't:Nostrum.Struct.Overwrite.t/0', or an equivalent map`?

lib/nostrum/api.ex
```diff
@@ -1156,74 +1169,113 @@ defmodule Nostrum.Api do
...
+  Gets a list of guild channels.
+
+  If successful, returns `{:ok, channels}`. Otherwise, returns a `t:Nostrum.Api.error/0`.
+
+  ## Examples
 
-  Guild to get channels for is specified by `guild_id`.
+  ```Elixir
+  Nostrum.Api.get_guild_channels(81384788765712384)
+  {:ok, [%Nostrum.Struct.Channel{} | _]}
```
Kraigie: Kind of a meh example IMO, but I don't really have any suggestions to make it better. 🙃

lib/nostrum/api.ex
Kraigie: I think this is also relevant in the `create_guild_channel` method

lib/nostrum/cache/channel_cache.ex
skippi: Right now, the `ChannelCache` only stores DM Channels. This part of the Dispatch module is responsible for this:

https://github.com/skippi/nostrum/blob/0e2124b8190fd7d7e7bd5eb05a9d98e1bca14061/lib/nostrum/shard/dispatch.ex#L38

Therefore, I think `Channel.dm_channel()` is fairly accurate for describing the channels that `ChannelCache` works with. Though, I do think that the `ChannelCache` has room for improvement in the future (such as working for all kinds of channels).

lib/nostrum/cache/channel_cache.ex
skippi: It's there. It's below the line you linked.

lib/nostrum/struct/channel.ex
skippi: I've been keeping the `p_encode/0` functions in for compatibility with structs that haven't been updated. I do plan on removing these once we remove all `p_encode/0` references.

lib/nostrum/api.ex
skippi: If we ever do go around mocking these API requests, these will at least cover the decoding aspect. Maybe this specific example should of had `%Nostrum.Struct.Channel{guild_id: 81384788765712384}` instead of plain `%Nostrum.Struct.Channel{}`.

lib/nostrum/api.ex
skippi: I'm down for this change, but I have an alternate solution we could try (one that is more struct heavy):

To solve this problem, could a `new/4` function be feasible? Might simplify the process a little bit:

```elixir
# @spec new(Overwrite.id(), Overwrite.name(), Overwrite.allow(), Overwrite.deny())

iex> alias Nostrum.Struct.Overwrite
...> ow = Overwrite.new(89918932789497856, "member", 0, 0x00001)
...> Nostrum.Api.modify_channel(381889573426429952, permission_overwrites: [ow])
{:ok, %Nostrum.Struct.Channel{}}
```

Alternatively, we could also try making builder functions for the Overwrite module (like Embed). But, I don't think this is a really elegant solution.

lib/nostrum/struct/channel.ex
Kraigie: Yep, that's fine.

lib/nostrum/cache/channel_cache.ex
Kraigie: Don't know how I missed that. 👍

lib/nostrum/cache/channel_cache.ex
Kraigie: The `ChannelCache` does currently work for all kinds of channels which is why I brought this up

lib/nostrum/api.ex
Kraigie: `new` sounds very OOP, but for modules with permissions maybe we can think about doing something like this. We can save it for later when we do other related work. For now I think the doc change would be sufficient.

lib/nostrum/api.ex
Kraigie: I'm still on the fence of how useful API mocks would be since there's not a whole lot of logic behind them. I guess we can leave this alone for now.

lib/nostrum/struct/channel.ex
Kraigie: I guess if it is a valid mention (so Discord has some interaction with it) then it should be a valid type for this function.

lib/nostrum/struct/overwrite.ex
```diff
@@ -14,7 +16,7 @@ defmodule Nostrum.Struct.Overwrite do
   ]
 
   @typedoc "Role or User id"
-  @type id :: Snowflake.t()
+  @type id :: Role.id() | User.id()
```
Kraigie: I think this is better left the way we had it. For things like `guild_id` in the same module and throughout other modules, we don't link to `Guild.id()`, but to `Snowflake.t()` instead.

I think for consistency we should leave it how it was or change all other types lol

lib/nostrum/api.ex
```diff
@@ -1178,7 +1178,7 @@ defmodule Nostrum.Api do
 
   ```Elixir
   Nostrum.Api.get_guild_channels(81384788765712384)
-  {:ok, [%Nostrum.Struct.Channel{} | _]}
+  {:ok, [%Nostrum.Struct.Channel{guild_id: 81384788765712384} | _]}
```
Kraigie: Off the top of my head I think there was another example like this in the pr. Could it be changed too?

lib/nostrum/api.ex
skippi: API mocks would be useful for testing if payloads are decoded properly. Although, the question of whether the effort is worth the prize is certainly debatable.

lib/nostrum/api.ex
skippi: Found a couple that had errors in their examples. Fixed in latest commit.

lib/nostrum/api.ex
skippi: Fixed.

lib/nostrum/consumer.ex
```diff
@@ -74,7 +74,7 @@ defmodule Nostrum.Consumer do
   @typedoc """
   The two different types of channels - Guild channels and DM channels.
```
Kraigie: I think this should be changed to be more accurate, or we could remove this type from this module and change the types below to point towards the Channel struct type.

skippi:
Looks like we are good to go!

Kraigie:
Thanks!

Issue #53, opened 2018-04-02
Cache get() methods take keyword list


Marqin:
[Cache get methods](https://github.com/kraigie/nostrum/blob/master/lib/nostrum/cache/channel_cache.ex#L34) take keyword list for just one argument. It makes it harder to use pipes:

`msg.channel_id |> ChannelCache.get!` doesn't work and I have to use `msg.channel_id |> (&ChannelCache.get! id: &1).()`

There is method that takes whole message, but it works only for normal `get()` and not for banged `get!()` :(

PR #67, opened 2018-05-06
Guild Struct and API Fixes

skippi:
- Fix guild struct decoding values.
- Fix guild-related API endpoints not decoding to the correct types.
- Remove id-indexing in roles, members, channels of the guild struct.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -1123,38 +1123,82 @@ defmodule Nostrum.Api do
   end
 
   @doc """
-  Modify a guild's settings.
+  Modifies a guild's settings.
 
-  `options` is a map with the following optional keys:
-   * `name` - Guild name.
-   * `region` - Guild voice region id.
-   * `verification_level` - Guild verification level.
-   * `default_message_notifications` - Notifications setting.
-   * `afk_channel_id` - Id for afk channel.
-   * `afk_timeout` - Afk timeout in seconds.
-   * `icon` - Base64 encoded 128x128 jpeg image for guild icon.
-   * `owner_id` - User id to transfer guild ownership to.
-   * `splash` - Base64 encoded 128x128 jpeg image for guild splash.
+  This endpoint requires the `MANAGE_GUILD` permission. It fires the
+  `t:Nostrum.Consumer.guild_update/0` event.
+
+  If successful, returns `{:ok, guild}`. Otherwise, returns a `t:Nostrum.Api.error/0`.
+
+  ## Options
+
+    * `:name` (string) - guild name
+    * `:region` (string) - guild voice region id
+    * `:verification_level` (integer) - verification level
+    * `:default_message_notifications` (integer) - default message
+    notification level
+    * `:explicit_content_filter` (integer) - explicit content filter level
+    * `:afk_channel_id` (`t:Nostrum.Struct.Snowflake.t/0`) - id for afk channel
+    * `:afk_timeout` (integer) - afk timeout in seconds
+    * `:icon` (base64 data URI) - 128x128 jpeg image for the guild icon
+    * `:owner_id` (`t:Nostrum.Struct.Snowflake.t/0`) - user id to transfer
+    guild ownership to (must be owner)
+    * `:splash` (base64 data URI) - 128x128 jpeg image for the guild splash
+    (VIP only)
+    * `:system_channel_id` (`t:Nostrum.Struct.Snowflake.t/0`) - the id of the
+    channel to which system messages are sent
+
+  ## Examples
+
+  ```Elixir
+  Nostrum.Api.modify_guild(81384788765712384, name: "Elixir Guild")
+  {:ok, %Nostrum.Struct.Channel{id: 81384788765712384, name: "Elixir Guild"}}
+
+  Nostrum.Api.modify_channel(81384788765712384)
+  {:ok, %Nostrum.Struct.Channel{id: 81384788765712384}}
+  ```
   """
-  @spec edit_guild(integer, %{
-          name: String.t(),
-          region: String.t(),
-          verification_level: integer,
-          default_message_notifications: boolean,
-          afk_channel_id: integer,
-          afk_timeout: integer,
-          icon: String.t(),
-          owner_id: integer,
-          splash: String.t()
-        }) :: error | {:ok, Nostrum.Struct.Guild.t()}
-  def edit_guild(guild_id, options) do
-    case request(:patch, Constants.guild(guild_id), options) do
-      {:ok, body} ->
-        {:ok, Poison.decode!(body)}
+  @spec modify_guild(Guild.id(), options) :: error | {:ok, Guild.t()}
+  def modify_guild(guild_id, options)
 
-      other ->
-        other
+  def modify_guild(guild_id, options) when is_list(options),
+    do: modify_guild(guild_id, Map.new(options))
+
+  def modify_guild(guild_id, %{} = options) when is_snowflake(guild_id) do
+    validate_modify_guild_options!(options)
+
+    request(:patch, Constants.guild(guild_id), options)
+    |> handle_request_with_decode({:struct, Guild})
+  end
+
+  defp validate_modify_guild_options!(options) do
+    valid_option_names = [
+      :name,
+      :region,
+      :verification_level,
+      :default_message_notifications,
+      :explicit_content_filter,
+      :afk_channel_id,
+      :afk_timeout,
+      :icon,
+      :owner_id,
+      :splash,
+      :system_channel_id
+    ]
+
+    case Enum.reject(options, fn {k, _} -> k in valid_option_names end) do
+      [] -> :ok
+      invalid_options -> raise ArgumentError, "invalid options #{inspect(invalid_options)}"
     end
+  end
+
+  @doc """
+  Same as `modify_guild/2`, but raises `Nostrum.Error.ApiError` in case of failure.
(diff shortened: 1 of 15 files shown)


lib/nostrum/api.ex
```diff
@@ -1095,68 +1095,112 @@ defmodule Nostrum.Api do
...
+
+  def modify_guild(guild_id, %{} = options) when is_snowflake(guild_id) do
+    validate_modify_guild_options!(options)
+
+    request(:patch, Constants.guild(guild_id), options)
+    |> handle_request_with_decode({:struct, Guild})
+  end
+
+  defp validate_modify_guild_options!(options) do
+    valid_option_names = [
```
Kraigie: Overall I like this. My only concern is that it isn't consistent with the way we handle the rest of the `option` parameters throughout this module. I think if we want to add something like this it's fine but we should open up an issue or something to get the rest of the modules in line.

Some grips I have with this current implementation:
1. It'll be hard to maintain, but not un-doable. If more fields are added to the guild we'll need to add them here as well. Maybe there's a clever way of doing this, it might be worth looking into. 
2. The `modify_guild/2` method here will throw an error in a non-banged method which is something I think we should avoid. We'd likely want a way to return an error, maybe an `ApiError` instead of `{:ok, guild}` here.

lib/nostrum/cache/guild_cache.ex
```diff
@@ -0,0 +1,82 @@
...
+    case ChannelGuild.get_guild(channel_id) do
+      {:ok, guild_id} ->
+        get_by(%{id: guild_id})
+
+      {:error, _} ->
+        {:error, :invalid_channel_id}
+    end
+  end
+
+  def get_by(_), do: {:error, :invalid_clauses}
```
Kraigie: I think we should delegate this kind of error to the underlying feature. I think if we define our own `invalid_clause` error tuple it could be confusing and not consistent with how we handle this kind of stuff elsewhere.

lib/nostrum/cache/guild_cache.ex
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  # TODO: Possibly remove this. If channels store the guild_id with it, why should we need this?
+  def get_by(%{channel_id: channel_id}) when is_snowflake(channel_id) do
+    case ChannelGuild.get_guild(channel_id) do
+      {:ok, guild_id} ->
+        get_by(%{id: guild_id})
+
+      {:error, _} ->
+        {:error, :invalid_channel_id}
```
Kraigie: We have a defined error for the cache modules for lookup errors like this. I think it may need expanded upon as it's currently implemented pretty weakly. Something to consider.

lib/nostrum/cache/guild_cache.ex
```diff
@@ -0,0 +1,82 @@
...
+  ```Elixir
+  iex> Nostrum.Cache.GuildCache.get(0)
+  {:ok, %Nostrum.Struct.Guild{id: 0}}
+
+  iex> Nostrum.Cache.GuildCache.get(10)
+  {:error, :invalid_guild_id}
+  ```
+  """
+  @spec get(Guild.id()) :: {:ok, Guild.t()} | {:error, reason}
+  def get(id) when is_snowflake(id) do
```
Kraigie: How do you feel about porting over the rest of the methods into this module (`transform` et. al.)? In general they're all supposed to work as "cache" methods. And then we wouldn't have documentation for getting guilds across two modules.

lib/nostrum/cache/guild_cache.ex
```diff
@@ -0,0 +1,82 @@
...
+      {:ok, guild_pid} ->
+        guild = GenServer.call(guild_pid, {:transform, fn state -> state end})
+        {:ok, guild}
+
+      {:error, _} ->
+        {:error, :invalid_guild_id}
+    end
+  end
+
+  # TODO: Possibly remove this. If channels store the guild_id with it, why should we need this?
```
Kraigie: It might be useful it someone only had a `channel_id`, like in the event they were just debugging something. Just something to think about.

skippi: Having thought about this just now, you're right. Some of the events only provide `channel_id`s, so it'd be incredibly inconvenient to not have a clause like this. I'll be removing this TODO.

lib/nostrum/cache/guild_cache.ex
skippi: I'm on the same page as you. I'll be porting over much of the old code into the new `GuildCache` module (maybe some refactors as well). This commit was more or less an early review to see if we agree on the function layout for the new `GuildCache`.

skippi: Speaking of which, I took the opportunity to do something similar to #53 for this new module. I've changed some of the names around to better reflect how some of the functions would be used. Let me know if anything can be named better or structured better.

lib/nostrum/api.ex
skippi: Sorry for the long comment! I'll put up an issue if we can get a solid and maintainable design for handling these.

1. There is likely a clever way for doing these, but I'm not sure on a specific one for now. It'd be hard to automate option creation IMO. We would have to write it for each endpoint one way or another. Granted, we could likely write some macros to lessen code duplication.

I know for this particular Modify Guild endpoint, the possible options will always be a subset of the Guild struct fields. Although it wouldn't be 100% accurate, this means that we *could* use the Guild struct's keys to reduce code duplication. Here is the same `validate_modify_guild_options/1` function with my proposed design:

```elixir
defp validate_modify_guild_options!(options) do
  # This would return all keys in the guild struct. 
  #Ideally, this `__METADATA__/1` function would be hidden and generated by a macro (something like nostrum_struct).
  valid_option_names = Nostrum.Struct.Guild.__METADATA__(:keys) 

  case Enum.reject(options, fn {k, _} -> k in valid_option_names end) do
    [] -> :ok
    invalid_options -> raise ArgumentError, "invalid options #{inspect(invalid_options)}"
  end
end
```

With that out of the way, I think listing the options is only part of the problem. I think the entire problem would be maintaining both typespecs and the actual options list:

If we plan on adding specific option typespecs per endpoint in the future, then the project would have strong integration with code linters (dialyzer). However, maintaining something like this would be incredibly difficult I'd imagine. Read the comments for context:

```elixir
# Maintaining these typespecs isn't so much problematic IMO.
@type modify_guild_option :: 
  {:name, String.t()}
  | {:region, String.t()}
  | {:verification_level, integer}
  | {:default_message_notifications, integer}
# There are more options, but that is beyond the point of this code block.

# We have a map type because we support maps
@type modify_guild_options :: [modify_guild_option] | map

# Yikes. Two places where we would have to maintain the list of options.
defp validate_modify_guild_options!(options) do
  valid_option_names = [
    :name,
    :region,
    :verification_level,
    :default_message_notifications,
    :explicit_content_filter,
    :afk_channel_id,
    :afk_timeout,
    :icon,
    :owner_id,
    :splash,
    :system_channel_id
  ]

  case Enum.reject(options, fn {k, _} -> k in valid_option_names end) do
    [] -> :ok
    invalid_options -> raise ArgumentError, "invalid options #{inspect(invalid_options)}"
  end
end
```

Maybe we can write a macro for maintaining both the typespecs and options list simultaneously? I'm wondering if something like this is even necessary as well. If we go with pure typespecs, we'd have people rely on a code linter to fix their code (IMO, not a bad thing, but people without one would be in the dark). Again, not sure how we would go about simplifying this. I'll have to give this more thought. 

I'm starting to think that this is a change that shouldn't go in this PR due to its size.

2. I think we should throw an ArgumentError here. I throw an ArgumentError here because the validation works much like a guard clause. If there is a way to make this validation into an actual guard clause, I would have done it.

Having it return an error tuple is fine for API errors IMO. Let me know what you think.

lib/nostrum/cache/guild_cache.ex
skippi: Having looked at this again, maybe we should raise an ArgumentError here. It's certainly an error if the user does not give us a key that we support in the typespec.

lib/nostrum/api.ex
Kraigie: > I'm starting to think that this is a change that shouldn't go in this PR due to its size.

Agreed, lets take this out for now and make it it's own PR. 

If we can find a way to do it cleanly in a macro then I'm all for it. I don't think it's a good idea to solely have users rely on dialyzer for something like this. I'm hesistant to say this but we could rely on a struct - probably not a good idea though. 

> If there is a way to make this validation into an actual guard clause, I would have done it.

How does discord react if we send bad keys in this payload? It might be more worth it just to forgo all of this in lieu of discord errors. A lot less work on our end and the user is given instructions on what to expect. Something to think about. Especially when we convert to API version 7 with the more detailed error messages.

lib/nostrum/cache/guild_cache.ex
Kraigie: I'm more okay with an ArgumentError.

lib/nostrum/cache/guild_cache.ex
skippi: I've changed the error tuples to reflect this proposal, but not confident on how I've handled it. Review it whenever you can.

skippi:
Looks like this PR is _finally_ ready for a complete code review. Only concern I have is if we should check for keyword lists as an ArgumentError in Api functions. I have a couple of examples in the recent commits.

Kraigie:
Thanks!

PR #68, opened 2018-05-11
Correction for a doc example.

Andris0:
`Nostrum.Cache.UserCache.get/1` instead of undefined `Nostrum.Cache.User.get/1`.


Kraigie:
Thanks!

PR #70, opened 2018-05-16
Change the WS state from a map to a struct, fixes #65

### LINKED ISSUE #65: websocket state should be a struct
Currently the state of the websocket-controlling process is just stored as a map. We should instead store this as a struct so it can be properly documented as well as provide the guarantees of a struct.

The current state documentation can be found [here](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/consumer.ex#L60).

unleashy:
elixir is kinda cool actually

### DIFF
--- a/lib/nostrum/consumer.ex
+++ b/lib/nostrum/consumer.ex
@@ -31,19 +31,19 @@ defmodule Nostrum.Consumer do
   `event` is a tuple describing the event. The tuple will include information in
   the following format:
   ```Elixir
-  {event_name, {event_payload(s)}, ws_state}
+  {event_name, {event_payload(s)}, WSState.t}
   ```
 
   For example, a message create will look like this
   ```Elixir
-  {:MESSAGE_CREATE, {Nostrum.Struct.Message.t}, ws_state}
+  {:MESSAGE_CREATE, {Nostrum.Struct.Message.t}, WSState.t}
   ```
 
   In some cases there will be multiple payloads when something is updated, so as
   to include the new and the old versions. In the event of there being two payloads,
   the old payload will always be first, followed by the new payload.
   ```Elixir
-  {:USER_UPDATE, {old_user :: Nostrum.Struct.User.t, new_user :: Nostrum.Struct.User.t}, ws_state}
+  {:USER_UPDATE, {old_user :: Nostrum.Struct.User.t, new_user :: Nostrum.Struct.User.t}, WSState.t()}
   ```
 
   For a full listing of events, please see `t:Nostrum.Consumer.event/0`.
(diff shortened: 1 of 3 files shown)


lib/nostrum/consumer.ex
```diff
@@ -59,25 +60,8 @@ defmodule Nostrum.Consumer do
...
-  heartbeat hasn't been sent it will be the time the websocket process was started
-  * `:last_heartbeat_ack` (`t:DateTime.t/0` | nil) - the time the last heartbeat was acknowledged,
-  will be nil if a heartbeat hasn't been ACK'd yet
-  * `:heartbeat_ack` (boolean) - whether or not the last hearbeat sent was ack'd
-  * `:heatbeat_interval` (integer | nil) - interval at which heartbeats are sent
-  * `:heartbeat_process` (pid | nil) - pid responsible for triggering next heartbeat send
-  * `:zlib_ctx` (ref | nil) - reference to the current zlib context
   """
-  @type ws_state :: map
+  @type ws_state :: WSState
```
Kraigie: This should either 

1. Be changed to say `@type ws_state :: WSState.t` (after you make the other changes mentioned in this PR)

2. Removed, and instead change the types below to include `WSState.t`

lib/nostrum/struct/ws_state.ex
```diff
@@ -0,0 +1,66 @@
+defmodule Nostrum.Struct.WSState do
+  @moduledoc """
+  Struct representing the current WS state.
+  """
+
+  defstruct [
```
Kraigie: We should include a `t` type in this module, like is present [here](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/struct/user.ex#L67).

lib/nostrum/consumer.ex
```diff
@@ -22,6 +22,7 @@ defmodule Nostrum.Consumer do
 
   alias Nostrum.Shard.Stage.Cache
   alias Nostrum.Struct.Channel
+  alias Nostrum.Struct.WSState
```
Kraigie: This alias and the channel above it can be grouped.

Kraigie:
Also, we're not actually putting this struct into the gw state anywhere, we should probably do that [here](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/shard/session.ex#L33).

lib/nostrum/struct/ws_state.ex
```diff
@@ -0,0 +1,82 @@
...
+  The time the last heartbeat was sent, if a heartbeat hasn't been sent it
+  will be the time the websocket process was started
+  """
+  @type last_heartbeat_send :: DateTime.t()
+
+  @typedoc """
+  The time the last heartbeat was acknowledged, will be nil if a heartbeat
+  hasn't been ACK'd yet
+  """
+  @type last_heartbeat_ack :: DateTime.t()
```
Kraigie: This can also be `nil`

Kraigie:
Thanks!

PR #72, opened 2018-05-18
Add banged get with Message struct in channel_cache.ex, fixes #53

### LINKED ISSUE #53: Cache get() methods take keyword list

unleashy:
~~I decided to go with a more conservative approach and simply make a banged version of the get with Message struct instead of reworking `get` to not have a single keyword arg but idk.~~

Nevermind, went with reworking `get` to take in a simple argument list.

### DIFF
--- a/lib/nostrum/cache/channel_cache.ex
+++ b/lib/nostrum/cache/channel_cache.ex
@@ -25,11 +25,11 @@ defmodule Nostrum.Cache.ChannelCache do
   `t:Nostrum.Struct.Channel.dm_channel/0` references. To get channel
   information, a call is made to a `Nostrum.Cache.GuildCache`.
   """
-  @spec get(integer | Nostrum.Struct.Message.t()) :: {:error, atom} | {:ok, Channel.t()}
+  @spec get(Channel.id() | Nostrum.Struct.Message.t()) :: {:error, atom} | {:ok, Channel.t()}
   def get(%Nostrum.Struct.Message{channel_id: channel_id}), do: get(channel_id)
   def get(id), do: GenServer.call(ChannelCache, {:get, id})
 
-  @spec get!(integer | Nostrum.Struct.Message.t()) :: no_return | Channel.t()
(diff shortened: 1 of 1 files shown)


lib/nostrum/cache/channel_cache.ex
```diff
@@ -25,12 +25,16 @@ defmodule Nostrum.Cache.ChannelCache do
   `t:Nostrum.Struct.Channel.dm_channel/0` references. To get channel
   information, a call is made to a `Nostrum.Cache.GuildCache`.
   """
-  @spec get(id: integer | Nostrum.Struct.Message.t()) :: {:error, atom} | {:ok, Channel.t()}
-  def get(id: id), do: GenServer.call(ChannelCache, {:get, id})
-  def get(%Nostrum.Struct.Message{channel_id: channel_id}), do: get(id: channel_id)
+  @spec get(integer | Nostrum.Struct.Message.t()) :: {:error, atom} | {:ok, Channel.t()}
```
Kraigie: Can `integer` here be changed to `Channel.id()` to better reflect the type?

lib/nostrum/cache/channel_cache.ex
```diff
@@ -25,12 +25,16 @@ defmodule Nostrum.Cache.ChannelCache do
...
+  def get(%Nostrum.Struct.Message{channel_id: channel_id}), do: get(channel_id)
+  def get(id), do: GenServer.call(ChannelCache, {:get, id})
 
-  def get!(id: id) do
-    get(id: id)
+  @spec get!(integer | Nostrum.Struct.Message.t()) :: no_return | Channel.t()
```
Kraigie: And here also 🙃

Kraigie:
Thanks!

PR #73, opened 2018-05-22
Adjusted arguments for UserCache user lookup.

Andris0:
Instead of passing option list, where there is only 1 field that can be used, `Nostrum.Cache.UserCache.get/1` and its banged version will now take `Nostrum.Struct.User.id` as its only argument.

### DIFF
--- a/lib/nostrum/cache/user_cache.ex
+++ b/lib/nostrum/cache/user_cache.ex
@@ -30,15 +30,13 @@ defmodule Nostrum.Cache.UserCache do
   end
 
   @doc ~s"""
-  Retrieves a user from the cache.
-
-  `get/1` requires a keyword list as its only argument.
+  Retrieves a user from the cache by id.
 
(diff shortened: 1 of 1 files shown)


lib/nostrum/cache/user_cache.ex
```diff
@@ -30,38 +30,33 @@ defmodule Nostrum.Cache.UserCache do
...
   Returns `Nostrum.Struct.User.t` if found.
   Raises `Nostrum.Error.CahceError` if not found.
   """
-  @spec get!(id: integer) :: no_return | Nostrum.Struct.User.t()
-  def get!(id: id) do
-    get(id: id)
-    |> Util.bangify_find(id, __MODULE__)
-  end
+  @spec get!(User.id()) :: no_return | Nostrum.Struct.User.t()
+  def get!(id), do: get(id) |> Util.bangify_find(id, __MODULE__)
```
Kraigie: We should pipe `id` here.

```elixir
id |> get |> Util.bangify_find(id, __MODULE__)
```

PR #75, opened 2018-06-04
Update Invite struct

skippi:
It's resuming time!

This PR plans to fix conversions for the Invite struct. This affects the API and Gateway.


lib/nostrum/struct/invite.ex
```diff
@@ -3,74 +3,93 @@ defmodule Nostrum.Struct.Invite do
...
   def to_struct(map) do
-    new =
-      map
-      |> Map.update(:guild, %{}, &Guild.to_struct(&1))
-      |> Map.update(:channel, %{}, &Channel.to_struct(&1))
-      |> Map.update(:inviter, %{}, &User.to_struct(&1))
+    atom_map = map |> Map.new(fn {k, v} -> {Util.maybe_to_atom(k), v} end)
+
+    invite =
+      struct(__MODULE__, atom_map)
```
Kraigie: We should pipe `__MODULE__` in here.

lib/nostrum/struct/invite.ex
```diff
@@ -3,74 +3,93 @@ defmodule Nostrum.Struct.Invite do
...
+      |> Map.update(:guild, nil, &Util.cast(&1, {:struct, Guild}))
+      |> Map.update(:channel, nil, &Util.cast(&1, {:struct, Channel}))
+      |> Map.update(:inviter, nil, &Util.cast(&1, {:struct, User}))
+      |> cast_metadata(atom_map)
+
+    invite
```
Kraigie: We can remove this and just return the above.

lib/nostrum/api.ex
```diff
@@ -51,6 +51,7 @@ defmodule Nostrum.Api do
   alias Nostrum.Struct.Channel
   alias Nostrum.Struct.Emoji
   alias Nostrum.Struct.Guild.{Member, Role}
+  alias Nostrum.Struct.Invite
```
Kraigie: We should probably consolidate and nest alias here to correct some credo issues.

lib/nostrum/struct/invite.ex
```diff
@@ -3,74 +3,93 @@ defmodule Nostrum.Struct.Invite do
...
-
-  @typedoc "Whether the invite gives temporary membership"
-  @type temporary :: boolean
-
-  @typedoc "When the invite was created"
-  @type created_at :: String.t()
+  This field is `nil` unless this invite is returned by
+  `Nostrum.Api.get_channel_invites/1`.
+  """
+  @type metadata :: Metadata.t()
```
Kraigie: Sorry, but I think I got a little confused by the chat we had in the Discord channel.

Do we want to break away from the Discord docs on this one? I do agree that we'll have a bunch of nils in one of our Invite types, but that's what we've been doing.

What are you thoughts on this?

skippi: You are right. We probably should stick to the original layout we've been doing for sometime now. I'll be revising my spec and design.

However, hear me out on a concern of mine. My thoughts right now are that we're going to have to eventually break from the API. Especially when we start building structs for the Gateway events. The gateway events alter all sorts of JSON schema for its payloads. For example, some of the Gateway events contain an extra `:guild_id` field. Because the discord API has so many weird rules that bend out a struct's schema, I was wondering if we should break away from modeling a perfect 1 to 1 with the discord API. Maybe gateway structs should be an exception to the design so far? Let me know your thoughts while I wrap this PR up. 😈

skippi:
Ready for more code review.

lib/nostrum/struct/invite.ex
Kraigie: I think there's a couple different ways we could handle it.

1. We could just make new structs for the gw payloads. This would be a lot of rewriting, but would follow suit with what we've been doing.

2. We could forgo gateway structs in favor of another format, like a tuple or something. Since the user eventually handles these events as a tuple, I think this would be sane. For something like an added `guild_id` we could supply a tuple like `{:TYPING_START, guild_id :: Snowflake.t(), member :: Member.t()}`. Downside is that there would be some oddities like members inside message creates, as we probably don't want users to have to have `{:MESSAGE_CREATE, member :: Member.t(), message :: Message.t()}`.

3. We could expose these new fields in their respective structs. This has the downside of weird keys in our structs, but we've handled this before by just have different types where we explicitly state something will be `nil`. 

I think I'm leaning towards a combination of 1 & 2 so we could have something like

```elixir
%MessageCreateStruct{
  message: Message.to_struct(payload.msg),
  member: Member.to_struct(payload.member)
}
 ```

The downside here is that we'd have to redo all of the types for the consumer, but we're doing that anyway. I think this would be nice so we don't have to send nested tuples, and it'd be nice to have something like

```elixir
%UserUpdateStruct{
  old_user: old_user_struct,
  new_user: new_user_struct
}
```

Let me know your thoughts, maybe we should look more at actual cases of this to better make our choice.

Kraigie:
Thanks!

PR #77, opened 2018-06-15
Fixed typo with `:joined_at` field in Guild struct

skippi:
* Issue was that the `:joined_at` documentation was incorrectly stating that this field returned the guild's creation date.
* It's now been corrected to say that `:joined_at` is the bot user's join date for that guild.


skippi:
Ready for review.

Kraigie:
Thanks!

PR #78, opened 2018-06-16
Updated Travis to use Elixir 1.6.5

skippi:
* Fixes some formatting errors that are important for our travis builds.


Kraigie:
Thanks!

PR #79, opened 2018-06-16
Added `image_url/1` to Emoji module

skippi:
* We should add this because it makes getting an emoji's image URL less cumbersome.

### DIFF
--- a/lib/nostrum/struct/emoji.ex
+++ b/lib/nostrum/struct/emoji.ex
@@ -156,6 +156,36 @@ defmodule Nostrum.Struct.Emoji do
   def api_name(%__MODULE__{id: nil, name: name}), do: name
   def api_name(%__MODULE__{id: id, name: name}), do: "#{name}:#{id}"
 
+  @doc """
+  Returns the url of a custom emoji's image. If the emoji is not a custom one,
+  returns `nil`.
+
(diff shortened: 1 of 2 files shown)


lib/nostrum/struct/emoji.ex
```diff
@@ -156,6 +156,36 @@ defmodule Nostrum.Struct.Emoji do
...
+  iex> Nostrum.Struct.Emoji.image_url(emoji)
+  nil
+  ```
+  """
+  @spec image_url(t) :: String.t() | nil
+  def image_url(emoji)
+  def image_url(%__MODULE__{id: nil}), do: nil
+
+  def image_url(%__MODULE__{animated: true, id: id}),
+    do: "https://cdn.discordapp.com/emojis/#{id}.gif"
```
Kraigie: I think we should move these URLs to the constants folder.

skippi:
Putting on hold until #82 is complete.

lib/nostrum/struct/emoji.ex
skippi: Addressed in 0dac258.

skippi:
Ready for more code review.

Kraigie:
Thanks!

PR #80, opened 2018-06-16
Guild Utility Functions

skippi:
This adds in two functions for the guild module:

* `image_url/2`: For retrieving a guild's icon URL.
* `splash_url/2`: For retrieving a guild's splash URL.

### DIFF
--- a/lib/nostrum/struct/guild.ex
+++ b/lib/nostrum/struct/guild.ex
@@ -8,7 +8,7 @@ defmodule Nostrum.Struct.Guild do
   alias Nostrum.Struct.Guild.Member
   alias Nostrum.Struct.Guild.Role
   alias Nostrum.Struct.Snowflake
-  alias Nostrum.Util
+  alias Nostrum.{Constants, Util}
 
   defstruct [
     :id,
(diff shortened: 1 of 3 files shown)


lib/nostrum/struct/guild.ex
```diff
@@ -287,6 +287,60 @@ defmodule Nostrum.Struct.Guild do
...
+  iex> Nostrum.Struct.Guild.icon_url(guild)
+  nil
+  ```
+  """
+  @spec icon_url(t, String.t()) :: String.t() | nil
+  def icon_url(guild, image_format \\ "webp")
+  def icon_url(%__MODULE__{icon: nil}, _), do: nil
+
+  def icon_url(%__MODULE__{icon: icon, id: id}, image_format),
+    do: "https://cdn.discordapp.com/icons/#{id}/#{icon}.#{image_format}"
```
Kraigie: I think we should move these URLs to the constants file

skippi:
Putting on hold until #82 is complete.

lib/nostrum/struct/guild.ex
skippi: Addressed in af9af94.

skippi:
Ready for more code review.

Kraigie:
Thanks!

PR #82, opened 2018-06-16
Move avatar CDN URLs to constants.ex

### LINKED ISSUE #81: Move cdn urls to constants file
The CDN URLs [here](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/struct/user.ex#L124) and [here](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/struct/user.ex#L128) should be moved to the [constants folder](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/constants.ex).

JoeBanks13:
Implements 3 new functions to `lib/nostrum/constants.ex`:
```elixir
iex(4)> Nostrum.Constants.cdn_url
"https://cdn.discordapp.com"
iex(5)> Nostrum.Constants.cdn_avatar "1", "2", "3"
"/avatars/1/2.3"
iex(6)> Nostrum.Constants.cdn_embed_avatar "0"    
"/embed/avatars/0.png"
```

Closes #81

### DIFF
--- a/lib/nostrum/constants.ex
+++ b/lib/nostrum/constants.ex
@@ -3,6 +3,7 @@ defmodule Nostrum.Constants do
 
   # REFERENCE: https://gist.github.com/SinisterRectus/9518f3e7d0d1ccb4335b2a0d389c30b0
   def base_url, do: Application.get_env(:nostrum, :api_url, "https://discordapp.com/api/v6")
+  def cdn_url, do: Application.get_env(:nostrum, :cdn_url, "https://cdn.discordapp.com")
   def gateway, do: "/gateway"
   def gateway_bot, do: "/gateway/bot"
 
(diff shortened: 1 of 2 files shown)


lib/nostrum/constants.ex
```diff
@@ -3,6 +3,7 @@ defmodule Nostrum.Constants do
 
   # REFERENCE: https://gist.github.com/SinisterRectus/9518f3e7d0d1ccb4335b2a0d389c30b0
   def base_url, do: Application.get_env(:nostrum, :api_url, "https://discordapp.com/api/v6")
+  def cdn_url, do: Application.get_env(:nostrum, :cdn_url, "https://cdn.discordapp.com")
```
Kraigie: Can you change this function and the `base_url` one above it to just be a constant? Basically, just remove the `get_env` call since they're not needed anymore.

lib/nostrum/struct/user.ex
```diff
@@ -23,7 +23,9 @@ defmodule Nostrum.Struct.User do
   """
 
   alias Nostrum.Struct.Snowflake
-  alias Nostrum.Util
+  alias Nostrum.{Util, Constants}
+
+  use HTTPoison.Base
```
Kraigie: Why do we need this here?

lib/nostrum/constants.ex
JoeBanks13: Changed in 8e1b772

lib/nostrum/struct/user.ex
JoeBanks13: I... don't know. Removed in 4c2b2c5

JoeBanks13:
Reviews addressed.

PR #83, opened 2018-06-17
Permission Utility Functions

skippi:
This PR adds in some well-needed permission utility functions.

* Added in a Permission module for working with permissions.
* Added in `guild_permissions/2` to the Member module for getting a guild member's perms.

More to go, as this is WIP.

### DIFF
--- a/lib/nostrum/struct/guild/member.ex
+++ b/lib/nostrum/struct/guild/member.ex
@@ -89,10 +89,10 @@ defmodule Nostrum.Struct.Guild.Member do
   guild = Nostrum.Cache.GuildCache.get!(279093381723062272)
   member = Enum.find(guild.members, & &1.id === 177888205536886784)
   Nostrum.Struct.Guild.Member.guild_permissions(member, guild)
-  #=> #MapSet<[:administrator]>
+  #=> [:administrator]
   ```
   """
-  @spec guild_permissions(t, Guild.t()) :: Permission.permission_set()
+  @spec guild_permissions(t, Guild.t()) :: [Permission.t()]
   def guild_permissions(member, guild)
 
   def guild_permissions(%__MODULE__{user: %{id: user_id}}, %Guild{owner_id: owner_id})
@@ -114,7 +114,7 @@ defmodule Nostrum.Struct.Guild.Member do
       end)
       |> Permission.from_bitset()
 
-    if MapSet.member?(member_permissions, :administrator) do
+    if Enum.member?(member_permissions, :administrator) do
       Permission.all()
     else
       member_permissions
@@ -131,16 +131,16 @@ defmodule Nostrum.Struct.Guild.Member do
   member = Enum.find(guild.members, & &1.id === 177888205536886784)
   channel_id = 381889573426429952
   Nostrum.Struct.Guild.Member.guild_channel_permissions(member, guild, channel_id)
-  #=> #MapSet<[:manage_messages]>
+  #=> [:manage_messages]
   ```
   """
-  @spec guild_channel_permissions(t, Guild.t(), Channel.id()) :: Permission.permission_set()
+  @spec guild_channel_permissions(t, Guild.t(), Channel.id()) :: [Permission.t()]
   def guild_channel_permissions(%__MODULE__{} = member, guild, channel_id) do
     use Bitwise
 
     guild_perms = guild_permissions(member, guild)
 
-    if MapSet.member?(guild_perms, :administrator) do
+    if Enum.member?(guild_perms, :administrator) do
       Permission.all()
     else
       channel = Enum.find(guild.channels, &(&1.id == channel_id))
@@ -153,16 +153,17 @@ defmodule Nostrum.Struct.Guild.Member do
         channel.permission_overwrites
         |> Enum.filter(&(&1.id in overwrite_ids))
         |> Enum.map(fn overwrite -> {overwrite.allow, overwrite.deny} end)
-        |> Enum.map(fn {allow, deny} ->
-          {Permission.from_bitset(allow), Permission.from_bitset(deny)}
-        end)
-        |> Enum.reduce(fn {allow, deny}, {acc_allow, acc_deny} ->
-          {MapSet.union(acc_allow, allow), MapSet.difference(acc_deny, deny)}
+        |> Enum.reduce(fn {allow, deny}, {allow_acc, deny_acc} ->
+          {allow_acc ||| allow, deny_acc ||| deny}
         end)
 
+      allow_perms = allow |> Permission.from_bitset()
+      deny_perms = deny |> Permission.from_bitset()
+
       guild_perms
-      |> MapSet.difference(deny)
-      |> MapSet.union(allow)
+      |> Enum.reject(&(&1 in deny_perms))
+      |> Enum.concat(allow_perms)
+      |> Enum.dedup()
     end
   end
 
--- a/lib/nostrum/struct/guild/member.ex
+++ b/lib/nostrum/struct/guild/member.ex
@@ -21,8 +21,10 @@ defmodule Nostrum.Struct.Guild.Member do
   ```
   """
 
-  alias Nostrum.Struct.{Guild, Permission, Snowflake, User}
+  alias Nostrum.Struct.{Channel, Guild, Permission, Snowflake, User}
   alias Nostrum.Util
+
+  require Logger
 
   defstruct [
     :user,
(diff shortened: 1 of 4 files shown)


lib/nostrum/struct/guild/member.ex
```diff
@@ -21,6 +21,8 @@ defmodule Nostrum.Struct.Guild.Member do
   ```
   """
 
+  alias Nostrum.Struct.Guild
```
Kraigie: These ideally should be grouped

lib/nostrum/struct/permission.ex
```diff
@@ -0,0 +1,115 @@
...
+          | :deafen_members
+          | :move_members
+          | :use_vad
+
+  @type t ::
+          general_permission
+          | text_permission
+          | voice_permission
+
+  @type permission_set :: MapSet.t(t)
```
Kraigie: I don't think this is correct, but maybe I'm not understanding something. We're saying that it'll be a `MapSet` of `t`, but wouldn't this be able to include all three of the types mentioned above (general, text, and voice)?

If I'm correct, maybe it won't be worth separating these permission atoms into their own fields?

Also some things like `:view_channel` looks like it applies to both text and voice channels.

lib/nostrum/struct/permission.ex
```diff
@@ -0,0 +1,115 @@
...
+      if band(bitset, bitvalue) == bitvalue do
+        acc ++ [perm]
+      else
+        acc
+      end
+    end)
+    |> MapSet.new()
+  end
+
+  defp permission_to_bitvalue_map do
```
Kraigie: Do you think this would be better suited as a module attribute?

lib/nostrum/struct/permission.ex
skippi: Permissions are collective I think. A set of permissions can contain all the permissions. I made separate general, text, and voice permission types to organize the module like the discord client. However, I found out later that there wasn't much reason to do so :\

Reason being that I'm going to make a type for each permission. This would give documentation to each permission and makes it easy to read. It also follows how we've been documenting our modules for some time. Plus, having a type for each possible permission would allow us to link to the permission in the rest of the documentation.

skippi: Speaking of which, would a MapSet be ideal? The MapSet has some nice functions that work well for what the user would need to compare permissions (`member?`, `difference/2`, `disjoint?/2`, `subset?/2`). However, the user would need to create a new MapSet for each time they want to make a permission list. Might be good to make a `new_set/1` then?

Let me know your thoughts.

lib/nostrum/struct/permission.ex
skippi: It looks like using a module attribute results in a 2 second improvement against 1,000,000 `from_bitset/1` calls in the worst case scenario.

Addressed in 4466136.

lib/nostrum/struct/guild/member.ex
skippi: Addressed in edfaaa7.

lib/nostrum/struct/guild/member.ex
```diff
@@ -21,10 +21,11 @@ defmodule Nostrum.Struct.Guild.Member do
   ```
   """
 
-  alias Nostrum.Struct.Snowflake
-  alias Nostrum.Struct.User
+  alias Nostrum.Struct.{Channel, Guild, Permission, Snowflake, User}
   alias Nostrum.Util
 
+  require Logger
```
Kraigie: I don't see this being used anywhere.

test/nostrum/struct/guild/member_test.exs
```diff
@@ -1,8 +1,10 @@
 defmodule Nostrum.Struct.MemberTest do
   use ExUnit.Case, async: true
 
-  alias Nostrum.Struct.Guild.Member
-  alias Nostrum.Struct.User
+  alias Nostrum.Struct.{Channel, Guild, Permission, User, Overwrite}
+  alias Nostrum.Struct.Guild.{Member, Role}
+
+  require Logger
```
Kraigie: I don't see this being used anywhere.

lib/nostrum/struct/permission.ex
```diff
@@ -0,0 +1,182 @@
...
+
+  ## Examples
+
+  ```Elixir
+  iex> Nostrum.Struct.Permission.to_bit(:administrator)
+  8
+  ```
+  """
+  @spec to_bit(t) :: bit
+  def to_bit(permission), do: @permission_to_bit_map[permission]
```
Kraigie: We have a guard for `is_permission`, don't you think we should be using it here?

lib/nostrum/struct/permission.ex
```diff
@@ -0,0 +1,182 @@
...
+  if :administrator in member_perms do
+    IO.puts("This user has the administrator permission.")
+  end
+  ```
+  """
+
+  use Bitwise
+
+  @type bit :: non_neg_integer
+  @type bitset :: non_neg_integer
```
Kraigie: Do you think a *very* short set of docs could be useful for these? Maybe with a link to some relevant data?

lib/nostrum/struct/guild/member.ex
```diff
@@ -21,10 +21,11 @@ defmodule Nostrum.Struct.Guild.Member do
   ```
   """
 
-  alias Nostrum.Struct.Snowflake
-  alias Nostrum.Struct.User
+  alias Nostrum.Struct.{Channel, Guild, Permission, Snowflake, User}
```
Kraigie: 💯

lib/nostrum/struct/permission.ex
```diff
@@ -0,0 +1,182 @@
...
+  Converts the given enumerable of permissions to a bitset.
+
+  ## Examples
+
+  ```Elixir
+  iex> Nostrum.Struct.Permission.to_bitset([:administrator, :create_instant_invite])
+  9
+  ```
+  """
+  @spec to_bitset(Enum.t()) :: bitset
```
Kraigie: Do you think `Enum.t()` is appropriate here? Why not just `[t]`? Or `Enum.t(t)`. I'm not sure if the latter is syntactically correct.

lib/nostrum/struct/permission.ex
```diff
@@ -0,0 +1,182 @@
...
+  Converts the given bit to a permission.
+
+  ## Examples
+
+  ```Elixir
+  iex> Nostrum.Struct.Permission.from_bit(0x04000000)
+  :change_nickname
+  ```
+  """
+  @spec from_bit(bit) :: t
```
Kraigie: And what of the case where they pass in a bit that is not handled?

lib/nostrum/struct/permission.ex
```diff
@@ -0,0 +1,182 @@
...
+  [:change_nickname]
+
+  iex> Nostrum.Struct.Permission.from_bitset(0x08000002)
+  [:kick_members, :manage_nicknames]
+
+  iex> Nostrum.Struct.Permission.from_bitset(0)
+  []
+  ```
+  """
+  @spec from_bitset(bitset) :: [t]
```
Kraigie: This doesn't handle the empty list scenario.

skippi: The spec is correct here I think; it states that the function will return a "list with any number of type elements". [source](https://hexdocs.pm/elixir/typespecs.html)

Kraigie: Yep, you're correct.

lib/nostrum/struct/guild/member.ex
skippi: Addressed in f92695f.

test/nostrum/struct/guild/member_test.exs
skippi: Addressed in f92695f.

lib/nostrum/struct/permission.ex
skippi: Addressed in 3943749.

lib/nostrum/struct/permission.ex
skippi: I've added in a set of docs here at 86b8184.

lib/nostrum/struct/permission.ex
skippi: Dialyzer doesn't seem to complain so... changed in 57ef3c3.

lib/nostrum/struct/permission.ex
skippi: I've split up the function to one that returns ok error tuples and one that raises an error on failure in 0930894.

skippi:
Ready for more code review feedback.

PR #88, opened 2018-06-26
Allow more than 100 messages in bulk delete

jchristgit:
Previously `Api.bulk_delete_messages/3` would error (from the Discord API) when given more than 100 messages at once. With this PR, the message list will be chunked into blocks of 100 and sent off to the API chunk by chunk.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -741,7 +741,7 @@ defmodule Nostrum.Api do
   def bulk_delete_messages(channel_id, messages, filter \\ true)
 
   def bulk_delete_messages(channel_id, messages, false),
-    do: request(:post, Constants.channel_bulk_delete(channel_id), %{messages: messages})
+    do: send_chunked_delete(channel_id, messages)
 
   def bulk_delete_messages(channel_id, messages, true) do
     filtered_messages =
@@ -749,7 +749,24 @@ defmodule Nostrum.Api do
         (message_id >>> 22) + 1_420_070_400_000 > Util.now() - 14 * 24 * 60 * 60 * 1000
       end)
 
-    request(:post, Constants.channel_bulk_delete(channel_id), %{messages: filtered_messages})
+    send_chunked_delete(channel_id, filtered_messages)
+  end
+
+  @spec send_chunked_delete(
+    Nostrum.Struct.Snowflake.t(),
+    [Nostrum.Struct.Message.id()]
+  ) :: error | {:ok}
+  def send_chunked_delete(channel_id, messages) do
+    messages
+    |> Enum.chunk_every(100)
+    |> Stream.map(fn message_chunk ->
+      request(
+        :post,
+        Constants.channel_bulk_delete(channel_id),
+        %{messages: message_chunk}
+      )
+    end)
+    |> Enum.find({:ok}, &match?({:error, _}, &1))
   end
 
   @doc """
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -754,6 +754,7 @@ defmodule Nostrum.Api do
     snowflake_two_weeks_ago =
       DateTime.utc_now()
       |> DateTime.to_unix()
+      # 60 seconds * 60 * 24 * 14 = 14 days / 2 weeks
       |> Kernel.-(60 * 60 * 24 * 14)
       |> DateTime.from_unix!()
       |> Snowflake.from_datetime!()
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -734,7 +734,9 @@ defmodule Nostrum.Api do
   `messages` is a list of `Nostrum.Struct.Message.id` that you wish to delete.
   When given more than 100 messages, this function will chunk the given message
   list into blocks of 100 and send them off to the API. It will stop deleting
-  on the first error that occurs.
+  on the first error that occurs. Keep in mind that deleting thousands of
+  messages will take a pretty long time and it may be proper to just delete
+  the channel you want to bulk delete in and recreate it.
 
   This method can only delete messages sent within the last two weeks.
   `Filter` is an optional parameter that specifies whether messages sent over
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -749,10 +749,17 @@ defmodule Nostrum.Api do
     do: send_chunked_delete(messages, channel_id)
 
   def bulk_delete_messages(channel_id, messages, true) do
+    alias Nostrum.Struct.Snowflake
+
+    snowflake_two_weeks_ago =
+      DateTime.utc_now()
+      |> DateTime.to_unix()
+      |> Kernel.-(60 * 60 * 24 * 14)
+      |> DateTime.from_unix!()
+      |> Snowflake.from_datetime!()
+
     messages
-    |> Stream.filter(fn message_id ->
-      (message_id >>> 22) + 1_420_070_400_000 > Util.now() - 14 * 24 * 60 * 60 * 1000
-    end)
+    |> Stream.filter(&(&1 > snowflake_two_weeks_ago))
     |> send_chunked_delete(channel_id)
   end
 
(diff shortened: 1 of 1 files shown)


lib/nostrum/api.ex
```diff
@@ -734,15 +737,32 @@ defmodule Nostrum.Api do
...
 
-    request(:post, Constants.channel_bulk_delete(channel_id), %{messages: filtered_messages})
+    send_chunked_delete(channel_id, filtered_messages)
+  end
+
+  @spec send_chunked_delete(
+          Nostrum.Struct.Snowflake.t(),
+          [Nostrum.Struct.Message.id()]
+        ) :: error | {:ok}
+  def send_chunked_delete(channel_id, messages) do
```
Kraigie: We should make this private since the user shouldn't really ever call it.

lib/nostrum/api.ex
```diff
@@ -734,15 +737,32 @@ defmodule Nostrum.Api do
...
+    do: send_chunked_delete(channel_id, messages)
 
   def bulk_delete_messages(channel_id, messages, true) do
     filtered_messages =
       Enum.filter(messages, fn message_id ->
         (message_id >>> 22) + 1_420_070_400_000 > Util.now() - 14 * 24 * 60 * 60 * 1000
       end)
 
-    request(:post, Constants.channel_bulk_delete(channel_id), %{messages: filtered_messages})
+    send_chunked_delete(channel_id, filtered_messages)
```
Kraigie: If we rewrote `send_chunked_delete/2` to accept `filtered_messages` as the first parameters we could easily pipe all of this, IMO it would be much cleaner.

lib/nostrum/api.ex
```diff
@@ -725,6 +725,9 @@ defmodule Nostrum.Api do
   Deletes multiple messages from a channel.
 
   `messages` is a list of `Nostrum.Struct.Message.id` that you wish to delete.
+  When given more than 100 messages, this function will chunk the given message
+  list into blocks of 100 and send them off to the API. It will stop deleting
+  on the first error that occurs.
```
Kraigie: We should have a note here that deleting thousands of messages will take a pretty long time, and it some cases it might be more proper to just delete the channel and recreate it.

lib/nostrum/api.ex
```diff
@@ -734,15 +737,32 @@ defmodule Nostrum.Api do
...
+    send_chunked_delete(channel_id, filtered_messages)
+  end
+
+  @spec send_chunked_delete(
+          Nostrum.Struct.Snowflake.t(),
+          [Nostrum.Struct.Message.id()]
+        ) :: error | {:ok}
+  def send_chunked_delete(channel_id, messages) do
+    messages
+    |> Enum.chunk_every(100)
```
Kraigie: It does matter if this is a stream function, I think it should be.

lib/nostrum/api.ex
```diff
@@ -734,15 +737,32 @@ defmodule Nostrum.Api do
   def bulk_delete_messages(channel_id, messages, filter \\ true)
 
   def bulk_delete_messages(channel_id, messages, false),
-    do: request(:post, Constants.channel_bulk_delete(channel_id), %{messages: messages})
+    do: send_chunked_delete(channel_id, messages)
 
   def bulk_delete_messages(channel_id, messages, true) do
     filtered_messages =
       Enum.filter(messages, fn message_id ->
```
Kraigie: We could move this into the stream pipe in the `send_chunked_delete` function. This is something we should probably do, but will require you to rewrite these functions just a bit.

lib/nostrum/api.ex
Kraigie: This does fall out of line with how we do other functions in this module, but like I said elsewhere it should probably be private anyways.

lib/nostrum/api.ex
z64: I'm not sure if this has also been considered:

Only messages within the past two weeks can be deleted. Using this method (with large `messages`) may incur batches of history that mix snowflakes that are both within and outside of this window, so it may fail to bulk delete messages that can, in fact, be deleted.

In `discordrb` we synthesize a reference snowflake value to filter the users input. ([implementation](https://github.com/meew0/discordrb/blob/master/lib/discordrb/data.rb#L1997-L2015))

So:
- Do you expect the user to always pass suitable snowflakes?
- Do you want to ignore them internally (or make it optionable)

Not only for long-reaching bulk-deletes, but users attempting to prune "stale" channels (the latest message is already >2 weeks old) may get bitten by this.

---

[Relevant docs](https://discordapp.com/developers/docs/resources/channel#bulk-delete-messages)

Example API error:
```
$ curl \
  -H 'Content-Type: application/json' \
  -H 'Authorization: Bot TOKEN' \
  --data '{"messages":["461221076626833430", "427681027134324736", "427681350116704268"]}' \
  -X POST https://discordapp.com/api/v6/channels/246283902652645376/messages/bulk-delete
```
```
{"code": 50034, "message": "You can only bulk delete messages that are under 14 days old."}
```
`46*` is less than a day old, and `42*` are about 3 months old.

lib/nostrum/api.ex
jchristgit: Made private in f3d8061f4d75fa892046d273f19ed29624251c09.

lib/nostrum/api.ex
jchristgit: Changed to `Stream.chunk_every(100)` in f3d8061f4d75fa892046d273f19ed29624251c09.

lib/nostrum/api.ex
jchristgit: Changed to take the messages as the first argument & pipe the messages in instead in f3d8061f4d75fa892046d273f19ed29624251c09.

lib/nostrum/api.ex
jchristgit: Added in b0cb940401061ec4114a2497f4108caca5c1579e

Kraigie:
Gonna' put this on hole until #90 is implemented, so we can do things a little more cleanly.

lib/nostrum/api.ex
```diff
@@ -741,15 +746,31 @@ defmodule Nostrum.Api do
...
-    do: request(:post, Constants.channel_bulk_delete(channel_id), %{messages: messages})
+    do: send_chunked_delete(messages, channel_id)
 
   def bulk_delete_messages(channel_id, messages, true) do
-    filtered_messages =
-      Enum.filter(messages, fn message_id ->
-        (message_id >>> 22) + 1_420_070_400_000 > Util.now() - 14 * 24 * 60 * 60 * 1000
-      end)
+    messages
+    |> Enum.filter(fn message_id ->
```
Kraigie: We should change this to `Stream.filter` so we have an entire stream pipeline.

lib/nostrum/api.ex
```diff
@@ -741,15 +746,31 @@ defmodule Nostrum.Api do
...
+    do: send_chunked_delete(messages, channel_id)
 
   def bulk_delete_messages(channel_id, messages, true) do
-    filtered_messages =
-      Enum.filter(messages, fn message_id ->
-        (message_id >>> 22) + 1_420_070_400_000 > Util.now() - 14 * 24 * 60 * 60 * 1000
-      end)
+    messages
+    |> Stream.filter(fn message_id ->
+      (message_id >>> 22) + 1_420_070_400_000 > Util.now() - 14 * 24 * 60 * 60 * 1000
```
Kraigie: Now that #90 is implemented, we should create a snowflake here by taking the current date and pushing it back two weeks and comparing against that.

lib/nostrum/api.ex
```diff
@@ -749,10 +749,17 @@ defmodule Nostrum.Api do
     do: send_chunked_delete(messages, channel_id)
 
   def bulk_delete_messages(channel_id, messages, true) do
+    alias Nostrum.Struct.Snowflake
+
+    snowflake_two_weeks_ago =
+      DateTime.utc_now()
+      |> DateTime.to_unix()
+      |> Kernel.-(60 * 60 * 24 * 14)
```
Kraigie: Could we add a comment here that this is supposed to represent 2 weeks?

jchristgit: Added in 861ab4afc1679ad212cb849a017c8995886a3554.

Kraigie:
Thanks!

Issue #89, opened 2018-06-26
Implement Snowflake.from_time


z64:
Generating snowflake values has a few use cases:

- Fetching channel history before/after/around a certain time, when an existing message ID isn't known
- Filtering messages out of a collection older than two weeks for bulk deletion
- If search is ever exposed to bots, the search API uses snowflakes as well for its time filters
- For sending `nonce` in creating messages

Kraigie:
Implemented with #90

PR #90, opened 2018-06-26
Added `Snowflake.from_datetime/1`

skippi:
This introduces the feature request from #89.

I've also cleaned up the Snowflake module a little bit while I was here.

### DIFF
--- a/lib/nostrum/struct/snowflake.ex
+++ b/lib/nostrum/struct/snowflake.ex
@@ -158,7 +158,7 @@ defmodule Nostrum.Struct.Snowflake do
   #DateTime<2016-05-05 21:04:13.203Z>
   ```
   """
-  @spec creation_time(t) :: DateTime.t() | no_return
+  @spec creation_time(t) :: DateTime.t()
   def creation_time(snowflake) when is_snowflake(snowflake) do
     use Bitwise
 
--- a/lib/nostrum/struct/snowflake.ex
+++ b/lib/nostrum/struct/snowflake.ex
@@ -106,8 +106,8 @@ defmodule Nostrum.Struct.Snowflake do
   @doc """
   Converts the given `datetime` into a snowflake.
 
-  If `datetime` occured before the discord epoch, the function will return a
-  snowflake of `0`.
+  If `datetime` occured before the discord epoch, the function will return
+  `:error`.
 
   The converted snowflake's last 22 bits will be zeroed out due to missing data.
 
@@ -116,14 +116,14 @@ defmodule Nostrum.Struct.Snowflake do
   ```Elixir
   iex> {:ok, dt, _} = DateTime.from_iso8601("2016-05-05T21:04:13.203Z")
   iex> Nostrum.Struct.Snowflake.from_datetime(dt)
-  177888205536755712
+  {:ok, 177888205536755712}
 
   iex> {:ok, dt, _} = DateTime.from_iso8601("1998-12-25T00:00:00.000Z")
   iex> Nostrum.Struct.Snowflake.from_datetime(dt)
-  0
+  :error
   ```
   """
-  @spec from_datetime(DateTime.t()) :: t
+  @spec from_datetime(DateTime.t()) :: {:ok, t} | :error
   def from_datetime(%DateTime{} = datetime) do
     use Bitwise
 
@@ -131,9 +131,20 @@ defmodule Nostrum.Struct.Snowflake do
     discord_time_ms = unix_time_ms - @discord_epoch
 
     if discord_time_ms >= 0 do
-      discord_time_ms <<< 22
+      {:ok, discord_time_ms <<< 22}
     else
-      0
+      :error
+    end
+  end
+
+  @doc """
+  Same as `from_datetime/1`, except it raises an `ArgumentError` on failure.
+  """
+  @spec from_datetime!(DateTime.t()) :: t | no_return
+  def from_datetime!(datetime) do
+    case from_datetime(datetime) do
+      {:ok, snowflake} -> snowflake
+      :error -> raise(ArgumentError, "invalid datetime #{inspect(datetime)}")
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/nostrum/struct/snowflake.ex
```diff
@@ -19,6 +19,8 @@ defmodule Nostrum.Struct.Snowflake do
   """
   @type t :: 0..0xFFFFFFFFFFFFFFFF
 
+  @discord_epoch 1_420_070_400_000
```
Kraigie: I feel like we should maybe move this to the constants module, just in case we need to use this anywhere else, or if users want to use it for some reason.

lib/nostrum/struct/snowflake.ex
```diff
@@ -101,6 +103,40 @@ defmodule Nostrum.Struct.Snowflake do
...
+  def from_datetime(%DateTime{} = datetime) do
+    use Bitwise
+
+    unix_time_ms = datetime |> DateTime.to_unix(:milliseconds)
+    discord_time_ms = unix_time_ms - @discord_epoch
+
+    if discord_time_ms >= 0 do
+      discord_time_ms <<< 22
+    else
+      0
```
Kraigie: Do we want to go down this route, or do we want to return an error tuple?

lib/nostrum/struct/snowflake.ex
```diff
@@ -101,6 +103,40 @@ defmodule Nostrum.Struct.Snowflake do
...
+  iex> {:ok, dt, _} = DateTime.from_iso8601("1998-12-25T00:00:00.000Z")
+  iex> Nostrum.Struct.Snowflake.from_datetime(dt)
+  0
+  ```
+  """
+  @spec from_datetime(DateTime.t()) :: t
+  def from_datetime(%DateTime{} = datetime) do
+    use Bitwise
+
+    unix_time_ms = datetime |> DateTime.to_unix(:milliseconds)
```
Kraigie: There's no real reason to pipe here

lib/nostrum/struct/snowflake.ex
```diff
@@ -115,7 +151,7 @@ defmodule Nostrum.Struct.Snowflake do
   def creation_time(snowflake) when is_snowflake(snowflake) do
     use Bitwise
 
-    time_elapsed_ms = (snowflake >>> 22) + 1_420_070_400_000
+    time_elapsed_ms = (snowflake >>> 22) + @discord_epoch
 
     DateTime.from_unix!(time_elapsed_ms, :milliseconds)
```
Kraigie: Looking at this now, the spec doesn't have `no_return` listed as a possible return value.

skippi: I think we talked about this before, but this function actually cannot fail; pretty much impossible since the `is_snowflake/1` guard prevents inputting numbers that can make the function raise an error.

Kraigie: Dialyzer won't agree with you I think 🙃

lib/nostrum/struct/snowflake.ex
skippi: Fixed in 267c68b.

lib/nostrum/struct/snowflake.ex
skippi: Error tuples sound good. Changed in 0421d14.

lib/nostrum/struct/snowflake.ex
skippi: Changed in 6e090fa.

lib/nostrum/struct/snowflake.ex
skippi: I've ran dialyzer a couple of times. Not warning us about anything so far. If you can run it on your machine and see if a warning pops up, we'll definitely change it. And if a warning does pop up, we'll need to explain to the user that the function can't fail and that the no_return is to make dialyzer be quiet. :\

Kraigie: [Well...](https://github.com/elixir-lang/elixir/blob/v1.6.6/lib/elixir/lib/calendar/datetime.ex#L165)

Also, we're doing checks on our own, but if something changes behind the scenes we'll want to pass any errors through. Under certain circumstances it probably can fail for w/e reason, even though with our code it's very unlikely. I think it's important to have that as part of the spec because it's not under our direct control.

skippi: Fixed in 9acfcbe.

lib/nostrum/struct/snowflake.ex
```diff
@@ -111,12 +158,13 @@ defmodule Nostrum.Struct.Snowflake do
...
-  @spec creation_time(t) :: DateTime.t()
+  @spec creation_time(t) :: DateTime.t() | no_return
   def creation_time(snowflake) when is_snowflake(snowflake) do
     use Bitwise
 
-    time_elapsed_ms = (snowflake >>> 22) + 1_420_070_400_000
+    time_elapsed_ms = (snowflake >>> 22) + Constants.discord_epoch()
 
-    DateTime.from_unix!(time_elapsed_ms, :milliseconds)
+    {:ok, datetime} = DateTime.from_unix(time_elapsed_ms, :milliseconds)
```
Kraigie: By changing this from a banged function, we no longer need the `no_return`. I think the function was fine before, it just needed the `no_return`.

skippi: I've removed the `no_return` in a6375f3.

skippi: As per our talk in the discord channel, I've ran the math on the function. It can't fail. However, I'm also adding in some tests to make sure it stays that way. This ensures that f we ever do change the implementation, we'll know if the function can break.

See 0fd0cc8.

skippi:
Ready for more code review.

Kraigie:
Thanks!

Issue #91, opened 2018-06-27
User/Nick State Issues


barkerja:
I am noticing when a nickname is reset on a server (meaning the `nick` is `nil`), the `GUILD_MEMBER_UPDATE` event payload received still contains the old nick in the new member payload.

Kraigie:
Taken care of with #97, thanks for the report!

PR #97, opened 2018-06-28
Fix Members not updating new `nil` fields in cache

skippi:
This fixes the #91 issue.

I've also included some extra cleanups while I was here.


Kraigie:
Gonna' put this on hold, pending design decisions from #98

skippi:
I think we should focus on the design decisions here, since this is a more immediate issue. With that being said, I think 43f2b1d covers most of the concerns you might've had with this PR. Check it again when you get the chance. I've included a use module for the struct construction logic.

Kraigie:
I'm super hesitant to include the changes in this PR as they really fall out of line with the issue at hand. I think we should just solve this issue first and worry about integrating those changes properly elsewhere. 

Let me know your thoughts.

skippi:
This is ready for code review.

Kraigie:
Thanks!

PR #99, opened 2018-06-28
fix(deps): add :inets to extra applications

ajimre:


Kraigie:
Taken care of with #112

PR #102, opened 2018-07-19
Fix incorrect key in `Overwrite` map

jchristgit:
The Discord API sends the overwrite type as `type`, but `Overwrite` was expecting it to be `name`.


Kraigie:
Thanks!

PR #105, opened 2018-08-01
Fix for `Nostrum.Cache.ChannelCache.lookup/1`.

Andris0:
After cached guild channel indexing, guilds now hold channels in a map instead of a list and so channel lookup needs to be adjusted accordingly.


Kraigie:
Thanks!

PR #106, opened 2018-08-01
Fix old code examples & permission checking using `Enum.find`

jchristgit:
For reference, these were found using:
image


Kraigie:
Thanks!

PR #107, opened 2018-08-05
Change heartbeat messages to :debug instead of :info

queer:
Generally it isn't useful to have a constant `[info]  HEARTBEAT_ACK` in the logs, so this makes it be debug level so it can be configured away.


Kraigie:
Thanks!

PR #112, opened 2018-11-18
Add dependency to :inets

kevinkjt2000:
While packaging an application that depended on nostrum, I came across a runtime error that claimed the built-in Erlang httpd_util module was missing.  Adding :inets as an extra_application fixes the error, because distillery will bundle the httpd_util module while making a release.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -20,7 +20,7 @@ defmodule Nostrum.Mixfile do
 
   def application do
     [
-      extra_applications: [:logger],
+      extra_applications: [:logger, :inets],
       mod: {Nostrum.Application, []}
     ]
   end


kevinkjt2000:
Nevermind, #99 has this.

kevinkjt2000:
Closer look at #99 shows that there are merge conflicts that are unresolved, so I will leave this open.

Kraigie:
Was just about to merge the other one, but this will save me some work. Thanks!

Just FYI, this wasn't addressed because I wasn't sure how necessary/correct this approach was, but since there's been two PRs now trying to take care of it, I don't think there's any harm.

PR #113, opened 2018-11-21
add timestamp typespecs.

ppraisethesun:
Poison implements Encode for Date, Time, NaiveDateTime, DateTime

### DIFF
--- a/lib/nostrum/struct/embed.ex
+++ b/lib/nostrum/struct/embed.ex
@@ -75,7 +75,7 @@ defmodule Nostrum.Struct.Embed do
   @type url :: String.t() | nil
 
   @typedoc "Timestamp of embed content"
-  @type timestamp :: String.t() | nil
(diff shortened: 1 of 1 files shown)


Kraigie:
I'm not sure if this is necessary, but feel free to prove me wrong.

When we get an embed from discord, we currently never parse the timestamp. We leave it as a ISO8601 string. Similarly when we send an embed, I'm not sure what kind of verification discord does on their end for the timestamps.

I know the Date module has a method for converting to ISO8601, but I also know people have had problems with the format in the past on Discord's side. I think it may be worth exploring whether the elixir native methods work well with Discord, and then I could see a reason to add these types.

Let me know your thoughts!

ppraisethesun:
Not really critical. I just get this annoying compiler warning. I guess i will just stringify date before putting it into embed

Kraigie:
You get a compiler warning for a missing type spec? Is this dialyzer complaining or something else?

jchristgit:
Hi, we are closing a bunch of old PRs due to inactivity. Feel free to re-open if you still have this problem!

PR #114, opened 2018-11-22
Avoid empty map error during permission fetch

kevinkjt2000:
This is the stacktrace I was getting from calling `guild_channel_permissions`:
```
[
  {Enum, :reduce, 2, [file: 'lib/enum.ex', line: 1869]},
  {Nostrum.Struct.Guild.Member, :guild_channel_permissions, 3,
   [file: 'lib/nostrum/struct/guild/member.ex', line: 164]},
  {Bowser, :_check_set_perms, 1, [file: 'lib/bowser.ex', line: 46]},
  {Bowser, :set_command, 2, [file: 'lib/bowser.ex', line: 68]},
  {Bowser, :handle_event, 1, [file: 'lib/bowser.ex', line: 161]},
  {Task.Supervised, :do_apply, 2, [file: 'lib/task/supervised.ex', line: 89]},
  {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 249]}
]
%Enum.EmptyError{message: "empty error"}
```

I'm not entirely sure that `{0, 0}` is a good starting value?  At least the stacktrace is gone now and the correct permissions seem to be returned by the function.  Although, I do not have any complicated set of permissions on the discord server that I was testing this out on.

### DIFF
--- a/lib/nostrum/struct/guild/member.ex
+++ b/lib/nostrum/struct/guild/member.ex
@@ -161,7 +161,7 @@ defmodule Nostrum.Struct.Guild.Member do
         channel.permission_overwrites
         |> Enum.filter(&(&1.id in overwrite_ids))
         |> Enum.map(fn overwrite -> {overwrite.allow, overwrite.deny} end)
-        |> Enum.reduce(fn {allow, deny}, {allow_acc, deny_acc} ->
+        |> Enum.reduce({0, 0}, fn {allow, deny}, {allow_acc, deny_acc} ->
           {allow_acc ||| allow, deny_acc ||| deny}
         end)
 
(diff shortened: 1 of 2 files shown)


kevinkjt2000:
My reasoning was that the main operation of the anonymous function being passed to reduce is bitwise-or, so it seems like `{0, 0}` would be the only value that would allow bitwise-or to accumulate values.

Kraigie:
It does seem like permission overwrites can be empty which I believe is the problem. I think before we go forward with this, we'd want to add some test cases for this case, but it seems like this would work OK. Let me know if you feel comfortable or not adding the tests, if not I can handle it sometime after the American holiday is over.

kevinkjt2000:
I'll give adding test code a shot after the holiday is over, assuming that there is existing testing for the `guild_channel_permissions` function already.

Kraigie:
Thanks!

PR #115, opened 2018-12-17
Adjustment for Nostrum.Api.get_channel_messages!/3

Andris0:
Added a default value for argument `locator` so that function's argument signature would match the non-banged counterpart.


Kraigie:
Thanks!

PR #116, opened 2018-12-25
Default value for icon_url

Awlexus:
Make nil the default value for icon_url for put_footer/3

Let's be real, icon_url is going to be nil most of the time.


Kraigie:
Thanks!

PR #117, opened 2019-01-01
Fix Credo and Format errors, bump Elixir 1.7.4

daekano:
Fix failing Travis master build

* Enum.map/2 |> Enum.into/2 --> Enum.into/3
* Added some Poison.Encoder aliases
* Ordering of aliases
* Bump Elixir version to 1.7.4

### DIFF
--- a/lib/nostrum/struct/message.ex
+++ b/lib/nostrum/struct/message.ex
@@ -3,13 +3,8 @@ defmodule Nostrum.Struct.Message do
   Struct representing a Discord message.
   """
 
-  alias Nostrum.Struct.Embed
-  alias Nostrum.Struct.Message.Activity
-  alias Nostrum.Struct.Message.Application
-  alias Nostrum.Struct.Message.Attachment
-  alias Nostrum.Struct.Message.Reaction
-  alias Nostrum.Struct.Snowflake
-  alias Nostrum.Struct.User
+  alias Nostrum.Struct.{Embed, Snowflake, User}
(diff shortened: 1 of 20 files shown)


lib/nostrum/struct/message.ex
```diff
@@ -4,12 +4,12 @@ defmodule Nostrum.Struct.Message do
   """
 
   alias Nostrum.Struct.Embed
-  alias Nostrum.Struct.User
   alias Nostrum.Struct.Message.Activity
   alias Nostrum.Struct.Message.Application
   alias Nostrum.Struct.Message.Attachment
   alias Nostrum.Struct.Message.Reaction
   alias Nostrum.Struct.Snowflake
+  alias Nostrum.Struct.User
```
Kraigie: Small nitpick, but everywhere else we group similar alias together, this and some of the other struct modules being an exception. I think we should go ahead and change that while we're in here. If you feel like it you can go through and change it for any other offending modules, but if you don't then just this module would be great!

daekano: Do you mean something like the following:

```
Nostrum.Struct.{Embed, Snowflake, User}
Nostrum.Struct.Message.{Activity, Application, Attachment, Reaction}
```
Or

```
Nostrum.Struct.{Embed, Message.Activity, Message.Application, Message.Attachment, Message.Reaction, Snowflake, User}
```

Not sure if the last form is even possible.

Kraigie: I was implying the former. IMO the latter looks really messy.

Kraigie:
Thanks!

PR #118, opened 2019-01-02
Improve Gateway Error Handling

defnorep:
`Util.get_new_gateway_url/0` was not matching the cases of `Api.request/3`
correctly, which resulted in `CaseClauseError` being thrown when real
errors were occurring beneath.

Added a less restrictive case that bubbles up an ok/error tuple to
`Util.gateway/0` and allows it to make a decision on how to handle the error.

`Util.gateway/0` now raises the returned error message if `Util.get_new_gateway_url/0` fails.

### DIFF
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -145,16 +145,18 @@ defmodule Nostrum.Util do
   @spec gateway() :: String.t()
   def gateway do
     case :ets.lookup(:gateway_url, "url") do
-      [] -> get_new_gateway_url()
-      [{"url", url, shards}] -> {url, shards}
+      [{"url", url, shards}] ->
+        {url, shards}
+      [] -> 
+        case get_new_gateway_url() do
+          {:ok, url, shards} -> {url, shards}
+          {:error, message} -> raise(message)
+        end
     end
   end
 
   defp get_new_gateway_url do
     case Api.request(:get, Constants.gateway_bot(), "") do
-      {:error, %{status_code: code, message: message}} ->
-        raise(Nostrum.Error.ApiError, status_code: code, message: message)
-
       {:ok, body} ->
         body = Poison.decode!(body)
 
@@ -162,7 +164,10 @@ defmodule Nostrum.Util do
         shards = if body["shards"], do: body["shards"], else: 1
 
         :ets.insert(:gateway_url, {"url", url, shards})
-        {url, shards}
+
+        {:ok, url, shards}
+      
+      {:error, message} -> {:error, message}
     end
   end
 
@@ -219,7 +224,7 @@ defmodule Nostrum.Util do
   end
 
   # Handles the case where the given term is already indexed
-  def cast(values, {:index, index_by, type}) when is_map(values), do: values
+  def cast(values, {:index, _index_by, _type}) when is_map(values), do: values
   def cast(values, {:index, index_by, type}) when is_list(values) do
     values
     |> Enum.map(&{&1 |> get_in(index_by) |> cast(Snowflake), cast(&1, type)})
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -152,9 +152,14 @@ defmodule Nostrum.Util do
 
   defp get_new_gateway_url do
     case Api.request(:get, Constants.gateway_bot(), "") do
+      # Handle deeper API error.
+      # For example, not including a token can result in a somewhat opaque 401 unauthorized error
+      {:error, %Nostrum.Error.ApiError{response: %{message: message}}} ->
+        raise(message)
+
       {:error, %{status_code: code, message: message}} ->
         raise(Nostrum.Error.ApiError, status_code: code, message: message)
-
+  
       {:ok, body} ->
         body = Poison.decode!(body)
 
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -154,7 +154,9 @@ defmodule Nostrum.Util do
     case Api.request(:get, Constants.gateway_bot(), "") do
       # Handle deeper API error.
       # For example, not including a token can result in a somewhat opaque 401 unauthorized error
-      {:error, %Nostrum.Error.ApiError{response: %{message: message}}} ->
+      {:error, error}} ->
+        raise(error)
+      
         raise(message)
 
       {:error, %{status_code: code, message: message}} ->
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -145,18 +145,16 @@ defmodule Nostrum.Util do
   @spec gateway() :: String.t()
   def gateway do
     case :ets.lookup(:gateway_url, "url") do
-      [{"url", url, shards}] ->
-        {url, shards}
-      [] -> 
-        case get_new_gateway_url() do
-          {:ok, url, shards} -> {url, shards}
-          {:error, message} -> raise(message)
-        end
+      [] -> get_new_gateway_url()
+      [{"url", url, shards}] -> {url, shards}
     end
   end
 
   defp get_new_gateway_url do
     case Api.request(:get, Constants.gateway_bot(), "") do
+      {:error, %{status_code: code, message: message}} ->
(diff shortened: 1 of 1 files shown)


daekano:
^ Expected to fail first, other branch still needs to be merged and I wanted to discuss these changes.

lib/nostrum/util.ex
```diff
@@ -145,24 +145,29 @@ defmodule Nostrum.Util do
...
 
         url = body["url"]
         shards = if body["shards"], do: body["shards"], else: 1
 
         :ets.insert(:gateway_url, {"url", url, shards})
-        {url, shards}
+
+        {:ok, url, shards}
+      
+      {:error, message} -> {:error, message}
```
Kraigie: I'm not a super big fan of a case statement like this. Rather I'd be more inclined to fix the case statement we had before to handle an error struct instead of the map. 

Your approach is a little more clear so if you want to go that way maybe we can find a middle ground or just go with your method. Let me know your thoughts on the matter!

daekano: Are you talking about the case statement in `get_new_gateway_url`? What specifically are you not a fan of?

I waffled on the nested case statement in `gateway`, trying out some with statements, but I couldn't quite express multiple steps in a failure path the same way that I could in the success path. Maybe I just have to forget "success" and "failure" and invert the two...

Kraigie: Yeah, I was talking about the case statement in `get_new_gateway_url` that has a case that looks like `{:error, message} -> {:error, message}`. In my opinion it seems kind of silly to handle the case like this just to pass it up and raise an error one step higher.

The intended outcome of this function is to return the number of shards and gateway url, and the user will never interact with this function unless we reach the case where their token is invalid and we throw an error. In this case I think it's ok to throw the error at the lowest level. We can just match on the `Nostrum.Error.ApiError`

Let me know if you feel otherwise and we can investigate some other way to handle this!

daekano: Gotcha. That makes sense to me.

lib/nostrum/util.ex
```diff
@@ -152,9 +152,14 @@ defmodule Nostrum.Util do
 
   defp get_new_gateway_url do
     case Api.request(:get, Constants.gateway_bot(), "") do
+      # Handle deeper API error.
+      # For example, not including a token can result in a somewhat opaque 401 unauthorized error
+      {:error, %Nostrum.Error.ApiError{response: %{message: message}}} ->
```
Kraigie: We should just match on the error and raise it here, no need to raise just the message as that'll give us a runtime error instead of an `ApiError`.

This will also handle the case where we get an error from `httpoison`, in which case it will probably be whatever it needs to be to be raised.

I guess the comment may need some revising as well in this case.

lib/nostrum/util.ex
```diff
@@ -152,9 +152,14 @@ defmodule Nostrum.Util do
...
+        raise(message)
+
       {:error, %{status_code: code, message: message}} ->
```
Kraigie: Like I said in the Discord channel this is out of date in regards to the [new](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/api/ratelimiter.ex#L154) return formats. I believe it can be built into the other error case.

lib/nostrum/util.ex
Kraigie:
```suggestion
      {:error, error}} ->
        raise(error)
      
```

lib/nostrum/util.ex
```diff
@@ -152,9 +152,14 @@ defmodule Nostrum.Util do
...
         raise(Nostrum.Error.ApiError, status_code: code, message: message)
-
+
```
Kraigie: Is this an added space? Be sure to run the formatter!

lib/nostrum/util.ex
daekano: Chalk that up to my lack of knowledge in Elixir error handling. It makes sense to raise the error module. However, since we are making an API call in this function, shouldn't we specifically handle the API failure modes as well as a generic one?

Kraigie: Maybe I'm not understanding what you're saying, but we're handling both cases of errors here. 

This handles the Nostrum API error by raising the error struct, and it also handles the generic httpoison error by raising the error struct that they return. Is there another error you're talking about?

Also sorry this took so long to get back to you on, I've been quite under the weather the past few days!

daekano: No problem, I've been a little busy lately anyway.

I'm referring to your suggested change, where we are replacing the `Nostrum.Error.ApiError` type with the less specific `error`. Shouldn't we continue to match on `Nostrum.Error.ApiError`?

Kraigie: The `error` term here would pattern match the `Nostrum.Error.ApiError` struct that is returned by us. It's good to be less specific because this will also pattern match on any httpoison error.

But maybe I'm still not understanding what you're trying to say? 🙃

daekano: Haha! I'm definitely not explaining it well enough.

I realize that `error` will match any Api-thrown error, as well as any error tuple that it matches against. When we talk about the responsibility of the function, it is responsible for calling our domain (the Api module). Errors that can be raised by that domain might be handled explicitly.

So, in addition to the generic error term, I'm wondering if we should also match the specific Api-thrown error term.

We might not have any specific rescue functionality for `Nostrum.Error.ApiError` now, but at least the function does respect the possible return values and exceptions raised from the module it is calling, and it makes it easier to diagnose what is happening.

Kraigie: Hmm, I see where you're coming from now. I fully agree with everything you're saying, but I'd like to make a counter argument.

I think your argument is that the errors that we handle here may have certain outcomes that we wish to handle explicitly, such as an invalid token? I'd argue that any errors we get here (such as a token being invalid) are not things we should try to recover from. In the case of connection errors we'd have to add some logic around this area to handle errors like that. This function is only called when trying to start up the bot and most errors we encounter then are fatal.

Let me know your thoughts and if I was correct!

daekano: > I think your argument is that the errors that we handle here may have certain outcomes that we wish to handle explicitly, such as an invalid token?

Kind of, yes. In languages like PHP, I usually try to avoid generic exception handling and prefer to rely on exceptions thrown from the facade's (or module's) domain explicitly.

> I'd argue that any errors we get here (such as a token being invalid) are not things we should try to recover from.

Recover, no, but it would be nice to be able to indicate to a developer or operator what the issue is and how they can fix or work around it. I think this is fairly aligned with error messaging in a lot of modern languages and tools. I definitely like the idea of "let it crash" that Erlang/Elixir subscribe to, but this is a different class of error or log message in my opinion.

In any case, I'm 100% okay with avoiding making a change like that right now.

Kraigie: We could match on the two different types of errors, but what are we going to do then? We can't bubble them up to some return value without a lot of work, and 99% of users probably wouldn't see them anyways. I don't think we could add anything here to the error that wouldn't be shown my raising the error. Just raising the error is really all we can do IMO.

Maybe we could match on the 401 just for the `ApiError` to say their token is invalid?
We could match on just the error struct itself? That's the only way I can think of that would give us some control over the errors here, but even then the only validation would be that it's an error type we expect which I don't think is helpful. 

For what to do now, I think raising just the message is maybe not so great. I don't know how Discord sources their status code messages, and the point of the error struct is to be raised. Also, we're only limiting the amount of information available to the user. I think we should go with the generic error handling for the moment just to resolve the issue, and if we think of a different way to handle it then we can come back in and change it. 

Let me know your thoughts, and sorry once again for taking so long to get back to you!

jchristgit:
Hi @daekano, are you still interested in working on this PR? I completely understand if you don't have the time available to check it out again, just drop us a short note and we'll take up on it.

jchristgit:
Hi @daekano, did you find some time to look into this? Otherwise I'd take over here to fix the remaining comments.

jchristgit:
On master:
```elixir
      {:error, %{status_code: 401}} ->
```
is already raising a more specific error. I'm nuking.

PR #121, opened 2019-01-22
Added missing field for Nostrum.Struct.Message

Andris0:
Updated Nostrum.Struct.Message for 1:1 field consistency with Message object sent by Discord.
Annotated the possibility for Nostrum.Struct.Guild.Member struct's field `user` being `nil`.
Message Update event will now provide a Nostrum.Struct.Message struct instead of a map, as event spec suggests.
Fixed warnings for having unused variables in Nostrum.Util.cast/2.

### DIFF
--- a/lib/nostrum/struct/guild/member.ex
+++ b/lib/nostrum/struct/guild/member.ex
@@ -38,8 +38,11 @@ defmodule Nostrum.Struct.Guild.Member do
     def to_string(member), do: @for.mention(member)
   end
 
-  @typedoc "The user struct"
-  @type user :: User.t()
+  @typedoc """
+  The user struct. This field can be `nil` if Member struct came as a partial Member object included
+  in a message received from a guild channel.
+  """
+  @type user :: User.t() | nil
 
   @typedoc "The nickname of the user"
   @type nick :: String.t() | nil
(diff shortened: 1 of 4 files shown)


lib/nostrum/struct/message.ex
```diff
@@ -3,6 +3,7 @@ defmodule Nostrum.Struct.Message do
   Struct representing a Discord message.
   """
 
+  alias Nostrum.Struct.Guild.Member
```
Kraigie: Can we move this below the alias below?

lib/nostrum/struct/guild/member.ex
```diff
@@ -38,8 +38,11 @@ defmodule Nostrum.Struct.Guild.Member do
     def to_string(member), do: @for.mention(member)
   end
 
-  @typedoc "The user struct"
-  @type user :: User.t()
+  @typedoc """
+  The user struct. This field can be `nil` if Member struct came as a partial Member object included
```
Kraigie: I think this sentence will read better if we include a `the` after "`nil` if".

lib/nostrum/struct/message.ex
```diff
@@ -100,6 +102,11 @@ defmodule Nostrum.Struct.Message do
   """
   @type application :: Application.t() | nil
 
+  @typedoc """
+  Partial Guild Member object received with Message Create event if message came from a guild channel.
```
Kraigie: Like before, I think a `the` would make it read better before "Message Create event"

Kraigie:
Thanks!

PR #122, opened 2019-01-26
Replaced deprecated time units (1.8.0)

Andris0:
With Elixir 1.8.0, calling functions from `DateTime` module with [deprecated time units](https://github.com/elixir-lang/elixir/blob/v1.8/CHANGELOG.md#4-hard-deprecations) will produce a warning on each function call.
```
warning: deprecated time unit: :milliseconds. A time unit should be :second, :millisecond, 
:microsecond, :nanosecond, or a positive integer
  (elixir) lib/calendar/iso.ex:717: Calendar.ISO.precision_for_unit/1
  (elixir) lib/calendar/iso.ex:708: Calendar.ISO.from_unix/2
  (elixir) lib/calendar/datetime.ex:118: DateTime.from_unix/3
  (elixir) lib/calendar/datetime.ex:166: DateTime.from_unix!/3
```

### DIFF
--- a/lib/nostrum/struct/snowflake.ex
+++ b/lib/nostrum/struct/snowflake.ex
@@ -127,7 +127,7 @@ defmodule Nostrum.Struct.Snowflake do
   def from_datetime(%DateTime{} = datetime) do
     use Bitwise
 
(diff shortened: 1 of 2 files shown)


Kraigie:
Thanks!

PR #123, opened 2019-02-03
Documentation: Display "Nostrum" as package name in sidebar, reference proper `Ratelimiter` callback

jchristgit:
There's another broken link but I'm not sure what that's supposed to be: 
```elixir
warning: documentation references Nostrum.Struct.Emoji.format_custom_emoji/2 but it doesn't exist or it's listed as @doc false (parsing state docs)
```

### DIFF
--- a/docs/static/API.md
+++ b/docs/static/API.md
@@ -59,5 +59,5 @@ The ratelimiter at a high level works something like this:
 ## Rest Only
 If you only want to use the REST portion of the provided API, the only process
 needed is the ratelimiter. This can be manually started by calling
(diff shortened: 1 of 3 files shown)


Kraigie:
The error you mentioned is from [here](https://github.com/Kraigie/nostrum/blob/2ff47a08c0dd1ec24b7ef4172b540691012a6d4a/docs/static/State.md#structs). We removed that exact method in favor of some other methods that were a little more useful and more descriptive. If you want you could link to one of those methods, or you could choose any other utility method in a struct module.

jchristgit:
Thanks for the info - I changed it to reference `Nostrum.Struct.Emoji.image_url/1` instead.

PR #124, opened 2019-02-06
Request for feedback: add initial audit log retrieval support

jchristgit:
This PR adds support for retrieving audit logs. Right now it's more or less just parsing the response from the API. However, the API is a mess. A slight mess. A big mess. Pick your poison.

Since I'm not sure if we should structure this nicer (e.g. should we parse each audit log event type into its own struct? what about new audit log events here). Would be happy for some feedback.

### DIFF
--- a/lib/nostrum/struct/guild/audit_log.ex
+++ b/lib/nostrum/struct/guild/audit_log.ex
@@ -9,13 +9,13 @@ defmodule Nostrum.Struct.Guild.AuditLog do
 
   defstruct [:entries, :users, :webhooks]
 
-  @typedoc "Entries of this guild's audit log."
+  @typedoc "Entries of this guild's audit log"
   @type entries :: [AuditLogEntry.t()]
 
-  @typedoc "Users found in the audit log."
+  @typedoc "Users found in the audit log"
   @type users :: [User.t()]
 
-  @typedoc "Webhooks found in the audit log."
+  @typedoc "Webhooks found in the audit log"
   @type webhooks :: [Webhook.t()]
 
   @type t :: %__MODULE__{
--- a/lib/nostrum/struct/guild/audit_log.ex
+++ b/lib/nostrum/struct/guild/audit_log.ex
@@ -29,9 +29,9 @@ defmodule Nostrum.Struct.Guild.AuditLog do
     new =
       map
       |> Map.new(fn {k, v} -> {Util.maybe_to_atom(k), v} end)
-      |> Map.update(:audit_log_entries, [], &Util.cast(&1, {:list, {:struct, AuditLogEntry}}))
-      |> Map.update(:users, [], &Util.cast(&1, {:list, {:struct, User}}))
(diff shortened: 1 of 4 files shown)


lib/nostrum/struct/guild/audit_log.ex
```diff
@@ -0,0 +1,38 @@
...
+  @type t :: %__MODULE__{
+          entries: entries,
+          users: users,
+          webhooks: webhooks
+        }
+
+  @doc false
+  def to_struct(map) do
+    struct(
+      __MODULE__,
```
Kraigie: Any particular reason you're doing the updates this way? Discord isn't exactly known to be super faithful about the fields they send. I'd suggest you check out how any other struct modules does this and see if it'd be better to do it similarly here, either just to be more consistent, or to handle any possible missing keys.

lib/nostrum/struct/guild/audit_log_entry.ex
```diff
@@ -0,0 +1,71 @@
...
+    changes: changes,
+    id: id,
+    options: options,
+    reason: reason,
+    target_id: target_id,
+    user_id: user_id
+  }
+
+  @doc false
+  def to_struct(map) do
```
Kraigie: Similarly as above. I think we should try to be consistent with our method of transferring these maps to structs.

lib/nostrum/struct/guild/audit_log.ex
```diff
@@ -0,0 +1,38 @@
...
+  Represents a guild's audit log.
+  """
+
+  alias Nostrum.Struct.Guild.AuditLogEntry
+  alias Nostrum.Struct.{User, Webhook}
+  alias Nostrum.Util
+
+  defstruct [:entries, :users, :webhooks]
+
+  @typedoc "Entries of this guild's audit log."
```
Kraigie: We don't typically have periods at the end of the typedoc definitions

Kraigie:
I don't think we should make structs for each type of event. They can match on the audit log event if they want to. It'd be a mess on our end to support all of that.

I'll take another look at the actual implementation later to make sure the types and such are OK.

Also, don't forget to run the formatter. 🙃

lib/nostrum/struct/guild/audit_log.ex
jchristgit: This seemed the most intuitive - I see the point about missing keys though, maybe I should handle that more gracefully. Also, I could just use `%__MODULE__` directly. For `audit_log_entries`, I'd have to rename the key in `Map.update`, which means temporarily saving the key, deleting the other key, and then putting in the new key, which I'm not a huge fan of. Should I go with that approach or stick with this one and be more graceful about missing keys?

lib/nostrum/struct/guild/audit_log.ex
jchristgit: Removed in 44473a6

Kraigie:
Thanks!

PR #125, opened 2019-02-11
Update HTTPoison to 1.5

jchristgit:
Gave this a quick test on my own bot and it seems largely compatible. We don't use huge parts of it anyways, and the changelog also looks rather harmless (https://github.com/edgurgel/httpoison/blob/master/CHANGELOG.md)

Using `defp` yielded warnings in `Nostrum.Api.Base`, so that was switched to `def`.


Kraigie:
I'll check this out locally to make sure I don't run into any problems, but it looks OK to me.

Kraigie:
Thanks!

PR #126, opened 2019-02-16
Dializer Fixes for `Nostrum.Util`

skippi:
Changes:
- Deleted `Util.producers/0`. From what I read on the discord, this is unused. If I misunderstood, give me a heads up.
- Fixed typespecs for `Util.gateway/0` and `Util.fullsweep_after/0`. These are purely static analysis fixes for dialyzer.

### DIFF
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -136,13 +136,13 @@ defmodule Nostrum.Util do
     end
   end
 
-  @doc """
-  Returns the gateway url for current websocket connections.
+  @doc ~S"""
+  Returns the gateway url and shard count for current websocket connections.
 
   If by chance no gateway connection has been made, will fetch the url to use and store it
   for future use.
   """
-  @spec gateway() :: String.t()
+  @spec gateway() :: {String.t(), integer}
   def gateway do
     case :ets.lookup(:gateway_url, "url") do
       [] -> get_new_gateway_url()
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -219,7 +219,7 @@ defmodule Nostrum.Util do
   end
 
   @doc false
-  @spec fullsweep_after() :: non_neg_integer
+  @spec fullsweep_after() :: {:fullsweep_after, non_neg_integer}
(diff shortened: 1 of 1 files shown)


lib/nostrum/util.ex
```diff
@@ -4,29 +4,10 @@ defmodule Nostrum.Util do
   """
 
   alias Nostrum.{Api, Constants}
-  alias Nostrum.Shard.Stage.Producer
   alias Nostrum.Struct.Snowflake
 
   require Logger
 
-  @doc """
```
Kraigie: I'm okay with removing this, but if we're going to we should add a note somewhere so those creating their own custom pipelines know what they're supposed to subscribe to.

lib/nostrum/util.ex
```diff
@@ -136,13 +117,13 @@ defmodule Nostrum.Util do
     end
   end
 
-  @doc """
-  Returns the gateway url for current websocket connections.
+  @doc ~S"""
```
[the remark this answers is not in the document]
skippi: Force of habit. Fixed in 67ce3a1f7a107ac87128a0c3aa0137268f1fba8e.

lib/nostrum/util.ex
skippi: I'm not sure on the procedure for adding a customer consumer. It looks like the directions in the README docs are outdated. If you can list out the procedure, I can quickly follow up with a fix. Might be a documentation change or reverting this commit depending on what is said.

Kraigie: The [documentation](https://kraigie.github.io/nostrum/consumers.html#implementing-custom-consumers) looks up to date with me, besides there being an error (`To do so, simply add custom_consumer: false to your config.` -> `costum_consumer: true`). The bigger issue is that the implementation no longer works as a lot of the processes and supervision trees have changed since it was implemented. Also in the original implementation we weren't super thorough in making sure we didn't start processes that we didn't need.

I'd say for now lets leave this removed, and when the feature is re-implemented we can worry about it then.

Kraigie:
Thanks!

PR #131, opened 2019-03-02
add Modify Current User Nick

ndac-todoroki:
This adds an method to modify current nickname in that guild.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -1571,7 +1571,7 @@ defmodule Nostrum.Api do
   @doc """
   Modifies the nickname of the current user in a guild.
 
-  If successful, returns `{:ok}`. Otherwise, returns a `t:Nostrum.Api.error/0`.
+  If successful, returns `{:ok, %{nick: nick}}`. Otherwise, returns a `t:Nostrum.Api.error/0`.
 
   ## Options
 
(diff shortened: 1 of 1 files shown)


Kraigie:
Could we add a banged version of this method as well? Besides that it looks good to go to me!

ndac-todoroki:
Did that :)

BTW, sorry for leaving the other issues for a long time. I was busy with school and came back using Nostrum recently. I'll look for those issues again, since Nostrum had improved a lot since then :)

lib/nostrum/api.ex
```diff
@@ -1568,6 +1568,37 @@ defmodule Nostrum.Api do
...
+  @spec modify_current_user_nick(Guild.id(), options) :: error | {:ok, %{nick: String.t()}}
+  def modify_current_user_nick(guild_id, options \\ %{}) do
+    request(:patch, Constants.guild_me_nick(guild_id), options)
+    |> handle_request_with_decode()
+  end
+
+  @doc """
+  Same as `modify_current_user_nick/2`, but raises `Nostrum.Error.ApiError` in case of failure.
+  """
+  @spec modify_current_user_nick!(Guild.id(), options) :: error | {:ok, %{nick: String.t()}}
```
Kraigie: I don't believe this spec is correct.
```suggestion
  @spec modify_current_user_nick!(Guild.id(), options) :: no_return | %{nick: String.t()}
```

ndac-todoroki: You're right!

ndac-todoroki:
Being with your friends is one of the best things you could do in your life :)

I've been reading my old code to find out what I wanted to do, and been reading Nostrum's new code. I'll close those issues, but am already working for the fixes so will create a new PR! Just to let you know :)

Kraigie:
🙃 ❤️ 🙉

PR #137, opened 2019-05-20
Gateway Struct for Message Delete Event

skippi:
Introduces a Gateway struct for representing Message Delete events. This PR is loosely based off of some Gateway discussion in the discord. ~~I'm 10 months late, but better now than never~~

Let's iron out this PR as much as possible, as it will likely set a precedent for future Gateway structs.

### DIFF
--- a/lib/nostrum/gateway/message_delete_event.ex
+++ b/lib/nostrum/gateway/message_delete_event.ex
@@ -1,6 +1,6 @@
 defmodule Nostrum.Gateway.MessageDeleteEvent do
   @moduledoc """
-  Struct representing a Message Delete Event.
+  Struct representing a Message Delete Event
   """
 
   alias Nostrum.Snowflake
@@ -11,14 +11,14 @@ defmodule Nostrum.Gateway.MessageDeleteEvent do
     :guild_id
   ]
 
-  @typedoc "Id of the deleted message."
+  @typedoc "Id of the deleted message"
   @type id :: Snowflake.t()
 
-  @typedoc "Channel id of the deleted message."
+  @typedoc "Channel id of the deleted message"
   @type channel_id :: Snowflake.t()
 
   @typedoc """
-  Guild id of the deleted message.
+  Guild id of the deleted message
 
   `nil` if a non-guild message was deleted.
   """
--- a/lib/nostrum/gateway/message_delete_event.ex
+++ b/lib/nostrum/gateway/message_delete_event.ex
@@ -11,8 +11,17 @@ defmodule Nostrum.Gateway.MessageDeleteEvent do
     :guild_id
   ]
 
+  @typedoc "Id of the deleted message."
   @type id :: Snowflake.t()
+
+  @typedoc "Channel id of the deleted message."
   @type channel_id :: Snowflake.t()
+
+  @typedoc """
+  Guild id of the deleted message.
+
+  `nil` if a non-guild message was deleted.
+  """
   @type guild_id :: Snowflake.t() | nil
 
   @type t :: %__MODULE__{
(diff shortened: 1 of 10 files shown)


lib/nostrum/gateway/message_delete_event.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Nostrum.Gateway.MessageDeleteEvent do
```
Kraigie: Do you think this is a proper path for this module? Do you think this shouldn't be under `lib/nostrum/struct/`?

lib/nostrum/gateway/message_delete_event.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  alias Nostrum.Snowflake
+
+  defstruct [
+    :id,
+    :channel_id,
+    :guild_id
+  ]
+
+  @typedoc "Id of the deleted message."
```
Kraigie: We typically don't use periods at the end of our typedocs.

test/nostrum/shard/dispatch_test.exs
```diff
@@ -0,0 +1,31 @@
...
+
+  alias Nostrum.Shard.Dispatch
+  alias Nostrum.Gateway.MessageDeleteEvent
+
+  describe "handle_event/1" do
+    test "returns `%MessageDeleteEvent{}` given Message Delete Event" do
+      result =
+        Dispatch.handle_event(
+          :MESSAGE_DELETE,
+          %{
```
Kraigie: In the future we may want to use a payload like this multiple times. We should consider making this more accessible. It would also allow us to pretty easily throw some real examples into the mix if we wanted to make the tests more robust.

lib/nostrum/gateway/message_delete_event.ex
skippi: Resolved in 283604a.

test/nostrum/shard/dispatch_test.exs
skippi: Resolved in 7c53040. Let me know if I misunderstood the review.

lib/nostrum/gateway/message_delete_event.ex
skippi: Not a proper place. I placed the module in that directory out of habit. Resolved in b4f4faf.

skippi:
Ready for code review.

skippi:
Resolved concerns from the discord. Ready for code review.

mix.exs
```diff
@@ -18,6 +19,9 @@ defmodule Nostrum.Mixfile do
     ]
   end
 
+  defp elixirc_paths(:test), do: ["lib", "test"]
```
Kraigie: Since we moved the stubs module into the test directory do you think we should get rid of this?

test/nostrum/shard/dispatch_test.exs
```diff
@@ -0,0 +1,22 @@
+defmodule Nostrum.Shard.DispatchTest do
+  use ExUnit.Case, async: true
+
+  alias Nostrum.Shard.Dispatch
+  alias Nostrum.Struct.Event.MessageDelete
+
+  describe "handle_event/1" do
+    test "returns `%MessageDelete{}` given Message Delete Event" do
```
Kraigie: I don't think there's much sense in trying to format the struct here in the code markdown, we could just have it reference the message delete struct with plain good ol' fashion english.

skippi: Resolved in e3a0ce7.

mix.exs
skippi: I don't think we can get rid of this. Attempting to remove these changes results in the tests not compiling (because the `Stubs` module needs to be compiled). If you have an alternative solution, let me know.

PR #138, opened 2019-05-20
Enabled credo `--strict` flag for travis

skippi:
Enables more credo checks to improve code uniformity. ~~Also prevents me from incorrectly ordering aliases.~~


Kraigie:
Thanks!

PR #139, opened 2019-05-22
Give unique IDs to per-scheduler consumers.

jchristgit:
As seen in https://github.com/jchristgit/bolt/commit/5f5586eaf3a64c1858c0023f4f73aed1842c1c57, when creating multiple consumer processes, the supervisor needs to have a unique ID (any term) to identify them.


Kraigie:
Thanks!

PR #146, opened 2019-05-30
Pluggable cache backends

jchristgit:
This merge request implements support for pluggable caches on the user cache.

The current cache has been moved to `Nostrum.Cache.UserCache.ETS` and is used by default.

There is one remaining issue which is that the `bulk_create` function currently takes a list of guild members. To keep things generic, we should probably just pass this via `Stream.map` in the dispatcher, but I'm curious for other ideas.

### DIFF
--- a/lib/nostrum/cache/user_cache.ex
+++ b/lib/nostrum/cache/user_cache.ex
@@ -1,7 +1,7 @@
 defmodule Nostrum.Cache.UserCache do
   @default_cache_implementation Nostrum.Cache.UserCache.ETS
   @moduledoc """
-  Cache behaviour for users.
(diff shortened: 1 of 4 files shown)


tobbbles:
I think a really nice first stepping-stone to having plugable caches will be to introduce the behaviour that the existing {Channel,User,Guild}Cache modules conform to; and then in future refactor those out into an ETS specific pluggable cache

jchristgit:
Blocked on #214

jchristgit:
Happy second Birthday, my sweet child. 🥳🎉🎉🎉

jchristgit:
I will reopen this with the branch on this repository.

PR #147, opened 2019-06-05
Documentation corrections

LazyShpee:
- bad arity references
- inconsistent banged methods
- documented get_application_info/1
- kwl instead of map for edit_channel_permissions/2
- embed instead of list of embed (in accordance to discord's doc)

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -795,18 +795,18 @@ defmodule Nostrum.Api do
 
   Role or user to overwrite is specified by `channel_id` and `overwrite_id`.
 
-  `permission_info` is a map with the following required keys:
+  `permission_info` is a map with the following keys:
+   * `type` - Required; `member` if editing a user, `role` if editing a role.
    * `allow` - Bitwise value of allowed permissions.
    * `deny` - Bitwise value of denied permissions.
-   * `type` - `member` if editing a user, `role` if editing a role.
   """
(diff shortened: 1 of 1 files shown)


lib/nostrum/api.ex
```diff
@@ -799,35 +795,35 @@ defmodule Nostrum.Api do
...
    * `deny` - Bitwise value of denied permissions.
    * `type` - `member` if editing a user, `role` if editing a role.
   """
   @spec edit_channel_permissions(
           integer,
           integer,
-          allow: integer,
-          deny: integer,
-          type: String.t()
+          %{
```
Kraigie: We should include the `required` or `optional` type specs here for map keys to denote whether these keys are required or not. We don't do this currently in this module, but we should.

lib/nostrum/api.ex
```diff
@@ -799,35 +795,35 @@ defmodule Nostrum.Api do
...
-  Raises `Nostrum.Error.ApiError` if error occurs while making the rest call.
+  Same as `edit_channel_permissions/3`, but raises `Nostrum.Error.ApiError` in case of failure.
   """
   @spec edit_channel_permissions!(
           integer,
           integer,
-          allow: integer,
-          deny: integer,
-          type: String.t()
+          %{
```
Kraigie: Same here regarding the map keys in type specs.

Kraigie:
Thanks!

PR #148, opened 2019-06-10
Gateway Struct for Message Delete Bulk Event

skippi:
Introduces a Gateway struct for representing Message Delete Bulk events.

### DIFF
--- a/lib/nostrum/consumer.ex
+++ b/lib/nostrum/consumer.ex
@@ -22,7 +22,7 @@ defmodule Nostrum.Consumer do
 
   alias Nostrum.Shard.Stage.Cache
   alias Nostrum.Struct.{Channel, WSState}
-  alias Nostrum.Struct.Event.MessageDelete
+  alias Nostrum.Struct.Event.{MessageDelete, MessageDeleteBulk}
   alias Nostrum.Util
 
   @doc """
@@ -131,9 +131,8 @@ defmodule Nostrum.Consumer do
            {guild_id :: integer, old_role :: Nostrum.Struct.Guild.Role.t() | nil,
             new_role :: Nostrum.Struct.Guild.Role.t()}, WSState.t()}
   @type message_create :: {:MESSAGE_CREATE, {message :: Nostrum.Struct.Message.t()}, WSState.t()}
-  @type message_delete :: {:MESSAGE_DELETE, {message :: MessageDelete.t()}, WSState.t()}
-  @type message_delete_bulk ::
-          {:MESSAGE_DELETE_BULK, {updated_messages :: [Nostrum.Struct.Message.t()]}, WSState.t()}
+  @type message_delete :: {:MESSAGE_DELETE, {MessageDelete.t()}, WSState.t()}
+  @type message_delete_bulk :: {:MESSAGE_DELETE_BULK, {MessageDeleteBulk.t()}, WSState.t()}
   @type message_update ::
           {:MESSAGE_UPDATE, {updated_message :: Nostrum.Struct.Message.t()}, WSState.t()}
   @type message_reaction_add :: {:MESSAGE_REACTION_ADD, {map}, WSState.t()}
--- a/lib/nostrum/consumer.ex
+++ b/lib/nostrum/consumer.ex
@@ -65,8 +65,8 @@ defmodule Nostrum.Consumer do
           | {:max_seconds, non_neg_integer()}
           | {:subscribe_to, [GenStage.stage() | {GenStage.stage(), keyword()}]}
 
-  @type channel_create :: {:CHANNEL_CREATE, {Channel.t()}, WSState.t()}
-  @type channel_delete :: {:CHANNEL_DELETE, {Channel.t()}, WSState.t()}
+  @type channel_create :: {:CHANNEL_CREATE, Channel.t(), WSState.t()}
+  @type channel_delete :: {:CHANNEL_DELETE, Channel.t(), WSState.t()}
   @typedoc """
   Dispatched when a channel is updated.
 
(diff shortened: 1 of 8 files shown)


lib/nostrum/struct/event/message_delete_bulk.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  @typedoc """
+  Guild id of the deleted message
+
+  `nil` if a non-guild message was deleted.
+  """
+  @type guild_id :: Snowflake.t() | nil
+
+  @typedoc "Ids of the deleted messages"
+  @type ids :: [Snowflake.t(), ...]
```
Kraigie: I think the `...` might be a little confusing here, it almost seems like we're implying there can be `any` type items in here, and not just snowflakes.

lib/nostrum/consumer.ex
```diff
@@ -131,9 +131,8 @@ defmodule Nostrum.Consumer do
            {guild_id :: integer, old_role :: Nostrum.Struct.Guild.Role.t() | nil,
             new_role :: Nostrum.Struct.Guild.Role.t()}, WSState.t()}
   @type message_create :: {:MESSAGE_CREATE, {message :: Nostrum.Struct.Message.t()}, WSState.t()}
-  @type message_delete :: {:MESSAGE_DELETE, {message :: MessageDelete.t()}, WSState.t()}
-  @type message_delete_bulk ::
-          {:MESSAGE_DELETE_BULK, {updated_messages :: [Nostrum.Struct.Message.t()]}, WSState.t()}
+  @type message_delete :: {:MESSAGE_DELETE, {MessageDelete.t()}, WSState.t()}
```
Kraigie: I didn't think of this before, but if we're including everything in structs we should be able to drop the tuple. This would be a breaking change, but I think we should go ahead with it regardless. :^)

lib/nostrum/struct/event/message_delete_bulk.ex
skippi: No confusion from my perspective. According to the official docs, `[Snowflake.t(), ...]` is syntactical sugar for `nonempty_list(Snowflake.t())`. The delete message bulk event should contain at least one message id, so this typespec looks appropriate.

lib/nostrum/consumer.ex
skippi: Will do. I'll create an issue for making sure the tuple is removed in future structs.

skippi: Just remembered something. If we are introducing breaking changes, then we can take this opportunity to make a less verbose message passing format. Here is what we have right now:
```elixir
{event_atom, {payload}, WSState.t()}
# Example:
{:MESSAGE_DELETE, {%MessageDelete{...}}, %WSState{...}}
```
The struct type and `event_atom` tack on extra information. In addition to removing the tuple, we can reasonably reduce the event format to the following:
```elixir
{payload, WSState.t()}
# Example:
{%MessageDelete{...}, %WSState{...}}
```
This is an easy change that can fit within the scope of this PR. Let me know what you think.

lib/nostrum/struct/event/message_delete_bulk.ex
Kraigie: Yeah, you're completely correct.

lib/nostrum/consumer.ex
Kraigie: I think thematically this is a good change. My worries are that it makes things a little less explicit. As this `handle_event` is the defacto way that users interact with events from discord I think it's nice to be able to see that atom that refers directly to the event as opposed to a struct type. This is however just my opinion - if we're moving to structs everywhere then this might be the play to make. Let me know what you think and we can see if any others want to chime in.

As for fitting it into this pr, I think we should hold off for now. I'm okay with breaking changes for now, but I think we should just take care of all of these in one fell swoop after all the event types are sorted out.

skippi: Then we'll hold off on this change for now. I'll move on with removing the tuple.

skippi: Resolved in 2d487ca.

skippi:
This is ready for more code review.

Kraigie:
We should update the example in the README as well

test/nostrum/shard/dispatch_test.exs
```diff
@@ -18,5 +18,20 @@ defmodule Nostrum.Shard.DispatchTest do
...
+      result =
+        Dispatch.handle_event(
+          :MESSAGE_DELETE_BULK,
+          Stubs.gateway_message_delete_bulk_payload(),
+          %{}
+        )
+
+      expected =
+        {:MESSAGE_DELETE_BULK,
+         struct(MessageDeleteBulk, Stubs.gateway_message_delete_bulk_payload()), %{}}
```
Kraigie: It feels a little weird to me to be just throwing the event payload stub into the struct here. I know it's correct since this event isn't doing a whole lot, but I think it can be done better.

How would you feel about changing this to be like the `MessageDelete` event, but instead of the hardcoded keys we can access the fields individually when creating the expected struct? In the future where keys from payloads can be missing or added to events we won't have to worry about how to handle that.
```suggestion
         struct(MessageDeleteBulk, %{
           channel_id: Stubs.gateway_message_delete_bulk_payload().channel_id,
           guild_id: Stubs.gateway_message_delete_bulk_payload().guild_id,
           ids: Stubs.gateway_message_delete_bulk_payload().ids
         }),
         %{}
       }
```
I think this makes it clearer where the keys are coming from. I'd like to hear your thoughts.

skippi: If a payload needs to be updated due to missing fields or deprecated fields, then I feel it'd be appropriate to update the payload stub.

I don't think readers need to know where the keys are coming from. If my intuition is correct, the test checks that it converts the given payload to its struct counterpart. No part of that intuition requires knowing what keys are included in the payload.

skippi: On a slightly related note...

The payload stub doesn't cover every situation appropriately. For example, the `guild_id` in the bulk delete's stub can be missing from the payload. The current payload includes a `guild_id`.

It might be appropriate to introduce generators that can correctly produce that kind of data. Something like `StreamData` offers this. That'd allow us to incorporate tests that cover a larger variety of data samples.

Kraigie: I think the tests here should test that a given payload results in a proper event sent to the user. There's a lot of nuance here that I'm trying to plan ahead for.

Like for example, when a guild is created we do a bit of indexing on certain fields. In that case our expected result would need to be crafted a bit carefully, unless we have the fields indexed in the initial payload, but that kind of defeats the purpose IMO.

As for the `StreamData` inclusion and property testing, I think we should move this to it's own PR as it might be out of scope here.

skippi: Marking this as resolved. Cleared confusion on discord.

skippi:
Ready for more code review.

Kraigie:
Thanks!

PR #150, opened 2019-06-28
Use gun instead of Websockex.

jchristgit:
Boom, working websockets on OTP 22.

To do:
- [x] Proper reconnections on remote close


jos-b:
Rock solid PR jchristgit, can't wait 👍

lib/nostrum/shard/session.ex
```diff
@@ -81,59 +110,76 @@ defmodule Nostrum.Shard.Session do
...
+        "attempting reconnect"
     end)
 
-    :timer.cancel(state.heartbeat_ref)
+    {:ok, :cancel} = :timer.cancel(state.heartbeat_ref)
+    :ok = :gun.close(conn)
 
-    {:reconnect, state}
+    # Drop the `:gun_down` message from the mailbox.
+    receive do
```
Kraigie: We shouldn't do this here. If we're inside of a callback I don't think there's any guarantee we'll get sent a message by gun here and it could be causing the deadlock issue you were talking about. We should instead add another `handle_info` for a gun_down message and let it come in naturally.  

For both of these we can return `:noreply` (and maybe a helpful logger error message) at which point the ws connection should close and gun will try to reconnect. Gun will send a `gun_up` message as stated [here](https://ninenines.eu/docs/en/gun/1.0/manual/gun_up/) if it reconnects successfully and then we can upgrade to ws again and send a resume to Discord.

lib/nostrum/shard/session.ex
```diff
@@ -81,59 +110,76 @@ defmodule Nostrum.Shard.Session do
...
-     %{state | heartbeat_ref: ref, heartbeat_ack: false, last_heartbeat_send: DateTime.utc_now()}}
+    {:noreply, %{state | conn: worker}}
   end
 
-  def handle_disconnect(%{reason: reason}, state) when is_tuple(reason) do
+  def handle_info({:gun_ws, conn, _stream, {:close, errno, reason}}, state) do
     Logger.warn(fn ->
-      "websocket disconnected with reason #{inspect(reason)}, attempting reconnect"
+      "websocket disconnected with code #{errno}, reason #{inspect(reason)}, " <>
+        "attempting reconnect"
```
Kraigie: Maybe we should move this "attempting reconnect" portion elsewhere if we're not going to actually try reconnecting in this callback.

lib/nostrum/shard/session.ex
```diff
@@ -81,59 +110,76 @@ defmodule Nostrum.Shard.Session do
...
-    Logger.warn("heartbeat_ack not received in time, disconnecting")
-    {:close, state}
-  end
+    {:ok, :cancel} = :timer.cancel(state.heartbeat_ref)
+    :ok = :gun.close(conn)
 
-  def handle_cast(:heartbeat, state) do
-    {:ok, ref} =
-      :timer.apply_after(state.heartbeat_interval, WebSockex, :cast, [state.conn_pid, :heartbeat])
+    worker = connect(state.gateway)
```
Kraigie: I didn't see this handler before the prior review comments, but I'll leave them there regardless for posterity. Like I said elsewhere, we shouldn't try to reconnect the ws in this callback. Let it restart itself and handle it properly.

lib/nostrum/shard/session.ex
```diff
@@ -22,50 +27,74 @@ defmodule Nostrum.Shard.Session do
...
   end
 
   def request_guild_members(pid, guild_id, limit \\ 0) do
     payload = Payload.request_members_payload(guild_id, limit)
-    WebSockex.cast(pid, {:request_guild_members, payload})
+    GenServer.cast(pid, {:request_guild_members, payload})
   end
 
   def start_link([gateway, shard_num]) do
+    GenServer.start_link(__MODULE__, [gateway, shard_num])
```
Kraigie: We should add the option of setting the `fullsweep_after` start option here. It's a `spawn_opt` as you can see [here](https://hexdocs.pm/elixir/Process.html#t:spawn_opt/0) that users can choose to use so their bot uses less memory.

You can see how we implemented it before [here](https://github.com/Kraigie/nostrum/blob/ffb73b62d68a55b6adf8298a257a42734a6b4368/lib/nostrum/shard/session.ex#L48). Even then it wasn't doing anything back then, it will work now.

lib/nostrum/util.ex
```diff
@@ -138,7 +138,7 @@ defmodule Nostrum.Util do
       {:ok, body} ->
         body = Poison.decode!(body)
 
-        url = body["url"]
+        "wss://" <> url = body["url"]
```
Kraigie: What's the point of this?

jchristgit: Strip the leading `wss://`

lib/nostrum/shard/session.ex
```diff
@@ -22,50 +27,77 @@ defmodule Nostrum.Shard.Session do
...
-      state,
-      spawn_opt: [Util.fullsweep_after()]
-    )
+    {:noreply, state}
   end
 
-  def handle_connect(conn, state) do
-    Logger.metadata(shard: state.shard_num)
+  @spec connect(String.t()) :: pid()
+  defp connect(gateway) do
```
Kraigie: I'm not sure this needs to be in its own function, but if you really want it to I'm fine with that. We're only calling it in one place.

Kraigie:
Once all the issues are addressed I'll merge this PR. As for the problems you provided in the discord channel, `:data_error`'s from `:zlib` are something we dealt with in the past when using. Gonna' have to look into it more before we can really fix it.

jos-b:
Awesome PR Johannes, good job!

lib/nostrum/shard/session.ex
```diff
@@ -22,50 +27,77 @@ defmodule Nostrum.Shard.Session do
...
+    {:ok, worker} = :gun.open(:binary.bin_to_list(gateway), 443, %{protocols: [:http]})
+    {:ok, :http} = :gun.await_up(worker, @timeout_connect)
+    stream = :gun.ws_upgrade(worker, @gateway_qs)
```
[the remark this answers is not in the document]
jchristgit: I tend to agree, but if gun 2.0 is released, then there's a built-in function to do this, and I would prefer to keep all the connection interaction in top-level then (which would mean replacing `await_ws_upgraded/2` with the procedure from gun)

lib/nostrum/shard/session.ex
jchristgit: Leftover from old reconnection code. Dropped like I drop my ubers

lib/nostrum/util.ex
Kraigie: :+1:

lib/nostrum/shard/session.ex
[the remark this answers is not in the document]
Kraigie: :+1:

Kraigie:
While you're here can you fix the failing build? I apparently am not good with the alphabet.

lib/nostrum/shard/session.ex
```diff
@@ -32,45 +37,65 @@ defmodule Nostrum.Shard.Session do
...
-      __MODULE__,
-      state,
-      spawn_opt: [Util.fullsweep_after()]
-    )
+    {:noreply, state}
   end
 
-  def handle_connect(conn, state) do
-    Logger.metadata(shard: state.shard_num)
+  defp await_ws_upgraded(worker, stream) do
```
Kraigie: Can we change this to be more present tense and less past tense?

lib/nostrum/shard/session.ex
```diff
@@ -86,63 +111,63 @@ defmodule Nostrum.Shard.Session do
...
+        {:gun_down, _conn, _proto, _reason, _killed_streams, _unprocessed_streams},
+        state
+      ) do
+    {:noreply, state}
+  end
+
+  def handle_info({:gun_up, worker, _proto}, state) do
+    stream = :gun.ws_upgrade(worker, @gateway_qs)
+    await_ws_upgraded(worker, stream)
+    Logger.warn("Reconnected after connection broke.")
```
Kraigie: I think periods at the end of log messages are a little weird.

lib/nostrum/shard/session.ex
```diff
@@ -86,63 +111,63 @@ defmodule Nostrum.Shard.Session do
...
+        {:noreply, new_state}
 
       new_state ->
-        {:ok, new_state}
+        {:noreply, new_state}
     end
   end
 
+  def handle_info({:gun_ws, _conn, _stream, {:close, errno, reason}}, state) do
+    Logger.info("Shard websocket closed (errno #{errno}, reason #{inspect(reason)}).")
```
Kraigie: Another weird period fella' here.

lib/nostrum/shard/session.ex
```diff
@@ -86,63 +111,63 @@ defmodule Nostrum.Shard.Session do
...
-    {:reply, {:binary, payload}, state}
+    :ok = :gun.ws_send(state.conn, {:binary, payload})
+    {:noreply, state}
   end
 
   def handle_cast(:heartbeat, %{heartbeat_ack: false} = state) do
     Logger.warn("heartbeat_ack not received in time, disconnecting")
-    {:close, state}
+    :ok = :gun.close(state.conn)
+    {:stop, {:heartbeat_ack_timeout, state.shard_num}, state}
```
Kraigie: I'm not sure we should stop the process here because then we can't resume through discord. We should just cancel the timer ref here and let it restart gracefully.

lib/nostrum/shard/session.ex
```diff
@@ -86,63 +111,63 @@ defmodule Nostrum.Shard.Session do
...
   end
 
   def handle_cast(:heartbeat, state) do
```
Kraigie: We need to keep in mind how this will work. Right now if we successfully resume we will have two heartbeats because we don't cancel the timer on disconnect and for every hello we start heartbeating again. 

We should either:
 * Before we start a new timer, check if one currently exists, and if so use that one. This is potentially problematic if things line up weirdly as I'm not sure if discord expects the first heartbeat immediately or after the interval upon connection.

* Cancel the timer on disconnect and our logic should restart it when we get another hello message from discord.

lib/nostrum/shard/session.ex
jchristgit: Yes

lib/nostrum/shard/session.ex
jchristgit: But that's what I'm doing, huh?

I think that if we receive this we might not have a timer.

lib/nostrum/shard/session.ex
jchristgit: We don't resume successfully right now though...

lib/nostrum/shard/session.ex
jchristgit: I think your emojis in your review comment are weird, and such is the amassment of exclamation marks in your comment. Erlang software runs like a German car, it's this thing you can always rely on, and in order to stand behind that, the log messages in my applications should read like "Sir, I, the Erlang runtime, have once again done something where everything else would tremble in fear, I have reconnected in a message handler and will now continue to serve your application without any issues and proceed by processing my entire message queue without problems." and not like "yup this is 🐐 i think i reconnected". Therefore, I heavily disagree with this suggestion. This also applies to the comment below.

lib/nostrum/shard/session.ex
Kraigie: I mean, in a way we are cancelling the timer ref because we're stopping the gen_server process. I was trying to say that we should not stop the process and instead just cancel the timer ref and let gun try to reconnect for us, and in doing so start another heartbeat.

lib/nostrum/shard/session.ex
Kraigie: Because of the `:data_error` issue? While that may be the case I think your current implementation is still wrong.

lib/nostrum/shard/session.ex
jchristgit: :shipit:

jchristgit:
Wrong button

lib/nostrum/shard/session.ex
```diff
@@ -22,55 +27,75 @@ defmodule Nostrum.Shard.Session do
...
       shard_num: shard_num,
       gateway: gateway <> @gateway_qs,
       last_heartbeat_ack: DateTime.utc_now(),
-      heartbeat_ack: true
+      heartbeat_ack: true,
+      zlib_ctx: zlib_context
     }
 
-    Connector.block_until_connect()
+    Logger.debug(fn -> "Websocket connection up on worker #{inspect(worker)}." end)
```
Kraigie: Can we remove the period here? And capitalize?

Kraigie:
:woman_shrugging:

skwerlman:
:eyes:

PR #151, opened 2019-08-01
fix (almost) every dialyzer warning

skwerlman:
this is as done as im going to get it. barring me having broke something ofc


skwerlman:
still needs typespecs for `Struct.Guild.Integration` and `Struct.User.Connection` but while i try to get examples for those, the rest of the pr should be good to review

skwerlman:
i havent had too much time to work on this, and wont for a while, so im going to revert the changes to guildsupervisor since thats the main thing holding this back, and update this to work on master. should be mergable after that

skwerlman:
Should be good to merge, but double checking that it works wouldnt hurt lol

Remaining errors:

- [ ] Unknown type: Nostrum.Struct.Guild.Integration.t/0.
    Needs a type to be defined, but i dont know what it should look like
- [ ] Unknown type: Nostrum.Struct.User.Connection.t/0.
    Needs a type to be defined, but i dont know what it should look like
- [ ] lib/nostrum/cache/guild/guild_supervisor.ex:12:no_return Function init/1 has no local return.
    Caused by deprecation of the :simple_one_to_one strategy
- [ ] lib/nostrum/cache/guild/guild_supervisor.ex:13:call The function call will not succeed.
    Caused by deprecation of the :simple_one_to_one strategy

this is down from 46 errors on master to just 4

config/.credo.exs
```diff
@@ -41,10 +41,10 @@
...
         {Credo.Check.Warning.IExPry},
         {Credo.Check.Warning.IoInspect},
-        {Credo.Check.Warning.NameRedeclarationByAssignment},
-        {Credo.Check.Warning.NameRedeclarationByCase},
-        {Credo.Check.Warning.NameRedeclarationByDef},
-        {Credo.Check.Warning.NameRedeclarationByFn},
+        # {Credo.Check.Warning.NameRedeclarationByAssignment},
+        # {Credo.Check.Warning.NameRedeclarationByCase},
+        # {Credo.Check.Warning.NameRedeclarationByDef},
+        # {Credo.Check.Warning.NameRedeclarationByFn},
```
Kraigie: What are we getting rid of here?

lib/nostrum/cache/channel_cache.ex
```diff
@@ -20,6 +20,15 @@ defmodule Nostrum.Cache.ChannelCache do
 
   import Nostrum.Snowflake, only: [is_snowflake: 1]
 
+  # dialyzer seems to think lookup will never fail,
```
Kraigie: I think I want to look into this. If we're going to take care of these lets not add any more debt to the pile!

lib/nostrum/shard/session.ex
```diff
@@ -41,7 +41,7 @@ defmodule Nostrum.Shard.Session do
   end
 
   def start_link([gateway, shard_num]) do
-    GenServer.start_link(__MODULE__, [gateway, shard_num], spawn_opt: [Util.fullsweep_after()])
+    GenServer.start_link(__MODULE__, [gateway, shard_num])
```
Kraigie: I'm pretty skeptical of this change. I took a look at the GenServer start options [here](https://hexdocs.pm/elixir/1.9.1/GenServer.html#t:option/0) and nothing immediately looks amiss. I think we should look into this more as this is effectively removing functionality.

Kraigie:
> Should be good to merge, but double checking that it works wouldnt hurt lol
> 
> Remaining errors:
> 
> * [ ]  Unknown type: Nostrum.Struct.Guild.Integration.t/0.
>   Needs a type to be defined, but i dont know what it should look like

It should look something like this if you want to implement it. Or we could save it for another PR.
https://discordapp.com/developers/docs/resources/guild#integration-object

> * [ ]  Unknown type: Nostrum.Struct.User.Connection.t/0.
>   Needs a type to be defined, but i dont know what it should look like

https://discordapp.com/developers/docs/resources/user#connection-object

> * [ ]  lib/nostrum/cache/guild/guild_supervisor.ex:12:no_return Function init/1 has no local return.
>   Caused by deprecation of the :simple_one_to_one strategy
> * [ ]  lib/nostrum/cache/guild/guild_supervisor.ex:13:call The function call will not succeed.
>   Caused by deprecation of the :simple_one_to_one strategy
> 
> this is down from 46 errors on master to just 4

config/.credo.exs
skwerlman: None of these exist in credo anymore since they duplicate compiler functionality: https://github.com/rrrene/credo/issues/276

lib/nostrum/shard/session.ex
skwerlman: As i said, i havent looked at this pr for a while, but iirc there was some error that kept spawn_opt from being accepted as valid. I can double check this change in a bit tho, since the docs do suggest that it should work

skwerlman: ah, found it; just removing the list from around the call to `Util.fullsweep_after` is enough

lib/nostrum/cache/guild_cache.ex
```diff
@@ -24,6 +24,7 @@ defmodule Nostrum.Cache.GuildCache do
   @type reason ::
           :id_not_found
           | :id_not_found_on_guild_lookup
+          | :channel_not_found
```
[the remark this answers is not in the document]
skwerlman: this can happen because `GuildCache.select_by` accepts a selector function (which can return anything) and passes on the result as-is
https://github.com/Kraigie/nostrum/blob/6c99bb8af945d8ba5dd22058baaf090c06132088/lib/nostrum/cache/guild_cache.ex#L22
https://github.com/Kraigie/nostrum/blob/6c99bb8af945d8ba5dd22058baaf090c06132088/lib/nostrum/cache/guild_cache.ex#L146
The specific place it happens is here:
https://github.com/Kraigie/nostrum/blob/603da771b0171c71e49109dfb72f26a912037380/lib/nostrum/cache/channel_cache.ex#L96-L98
An alternate (and maybe more correct) fix would be to make a new `@type selector_return`, modify `selector` to require returning that type, and allow `select_by` to return that as well

lib/nostrum/consumer.ex
```diff
@@ -247,17 +246,15 @@ defmodule Nostrum.Consumer do
...
+      name: name
     )
   end
 
   def start_link(mod, options),
     do:
       ConsumerSupervisor.start_link(
         __MODULE__,
-        [mod, options],
-        spawn_opt: [Util.fullsweep_after()]
```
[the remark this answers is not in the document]
skwerlman: i will get to these when i have linux access again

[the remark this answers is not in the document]
skwerlman: I was able to add back this opt to the shard cache, but i cant here because ConsumerSupervisor.start_link doesnt accept most of the genserver options: https://hexdocs.pm/gen_stage/ConsumerSupervisor.html#t:option/0

lib/nostrum/cache/guild_cache.ex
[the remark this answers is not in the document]
Kraigie: I don't think your idea is bad but I have another idea I want to run by you.

Why don't we instead move the return out of the anonymous function. This way we're not introducing another potential return value of `{:error, :channel_not_found}` which we have to document in `GuildCache.select_by/2`

lib/nostrum/consumer.ex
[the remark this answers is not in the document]
Kraigie: That makes sense, thanks!

skwerlman:
hi, apologies for the long delay. finally have time to look at this again

Kraigie:
Cool, I had another look over it and seems besides the two open comments everything looks good to me!

skwerlman:
assuming the channel_cache fix is satisfactory, this should be good to merge :)

Kraigie:
I'd be happy to merge if you don't mind fixing the conflicts!

Kraigie:
Thanks!

PR #159, opened 2019-09-20
Fix Api.create_message! typespec

ThatsNoMoon:
`Api.create_message` treats the options parameter as the message content if it is a string. This is represented in the typespec for `Api.create_message`. `Api.create_message!` simply passes the options parameter to `Api.create_message`, so it should have the same typespec for this parameter.


Kraigie:
Good catch. Thanks!

PR #160, opened 2019-10-09
Strip minor parameters from `/reactions` endpoint.

### LINKED ISSUE #156: Ratelimiter not handling adding reaction properly
Adding multiple reactions one after another gives a `%Nostrum.Error.ApiError{response: %{global: false, message: "You are being rate limited.", retry_after: 108}, status_code: 429}` instead of being queued by Nostrum.Api.Ratelimiter and handled for you.

BlindJoker:
This should be enough to fix #156, though probably not the best way to do it.

While the new `X-RateLimit-Bucket` should've helped in this situation, in practice we'd still need to strip out the minor parameters, otherwise something like the following would happen:

> We queue a request to `/channels/123456789/messages/_id/reactions/⬅/@me`;
> We haven't seen it before -- our rate limiter says it's alright;
> Discord replies with an OK, we store the bucket for that route;
> We queue a request to `/channels/123456789/messages/_id/reactions/➡/@me`;
> We haven't seen it before -- our rate limiter says it's alright;
> Discord replies with 429.

Because of that, though handling the new header is worthwhile, I consider those two things separate issues.

### DIFF
--- a/lib/nostrum/api/ratelimiter.ex
+++ b/lib/nostrum/api/ratelimiter.ex
@@ -147,7 +147,8 @@ defmodule Nostrum.Api.Ratelimiter do
     if String.ends_with?(endpoint, "/messages/_id") and method == :delete do
       "delete:" <> endpoint
     else
-      endpoint
+      # Strips `emoji` and `user_id` suffix off the reaction endpoints.
+      String.replace(endpoint, ~r[/reactions/.+/.+$]iu, "/reactions")
     end
   end
 
--- a/test/nostrum/api/ratelimit_test.exs
+++ b/test/nostrum/api/ratelimit_test.exs
@@ -43,6 +43,18 @@ defmodule Nostrum.Api.RatelimitTest do
     assert result == expected
   end
 
+  test "reaction endpoint" do
+    expected = "/channels/#{@test_channel}/messages/_id/reactions"
+
+    result =
+      Nostrum.Api.Ratelimiter.get_endpoint(
+        "/channels/#{@test_channel}/messages/#{@test_message}/reactions/⬅/@me",
+        :get
+      )
+
+    assert result == expected
+  end
+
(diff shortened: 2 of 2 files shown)


jchristgit (changes requested):
I tested this with bolt's paginator, but I think something is going wrong - I've added `IO.inspect`s for the ratelimiter endpoint:

```elixir
endpoint: "/channels/455589923626549259/messages/_id/reactions"
endpoint: "/channels/455589923626549259/messages/_id/reactions"

22:41:13.688 [error] Task #PID<0.500.0> started from #PID<0.482.0> terminating
** (MatchError) no match of right hand side value: {:error, %Nostrum.Error.ApiError{response: %{global: false, message: "You are being rate limited.", retry_after: 75}, status_code: 429}}
    (bolt 0.11.2) lib/bolt/paginator.ex:59: Bolt.Paginator.paginate_over/3
    (bolt 0.11.2) lib/bolt/consumer/message_create.ex:15: Bolt.Consumer.MessageCreate.handle/1
    (elixir 1.11.2) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (stdlib 3.14) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
Function: #Function<0.130613716/0 in Bolt.Consumer.start_link/1>
    Args: []
```

jchristgit:
Hi @ghost, are you still interested in working on this issue? I understand if you're not, feel free to drop a short note & we'll take up on it.

jchristgit:
There's a deeper issue with our ratelimiter here, per our internal dev discussion yesterday. @jb3 also discovered problems related to `add_guild_member_role` ratelimiting.

jchristgit:
I'm going to close this for now as it does not fix the issue. Joe and me will have to document the bigger ratelimiter issue.

Issue #164, opened 2020-03-15
Nostrum.Api.leave_guild/1 fails with 400: Bad Request


Meow:
`Nostrum.Api.leave_guild/1` fails to leave the guild, no matter what the argument passed is (string / integer).

**Expected Behavior:**
```elixir
Nostrum.Api.leave_guild(guild_snowflake)
```
would leave the guild as specified by the `guild_snowflake` variable.

**Actual Behavior:**
Fails with `400: Bad Request` response, according to the error inspection.

Here is the actual code of the command in my own bot:
```elixir
  def handle(msg, _, %Arguments{parsed: args}) do
    guild_id =
      args
      |> Enum.at(0)

    case Api.leave_guild(guild_id) do
      {:ok} ->
        Api.create_message(
          msg.channel_id,
          "I left the following guild: #{guild_id |> to_string()}"
        )

      {:error, err} ->
        Api.create_message(
          msg.channel_id,
          "Failed to leave guild: '#{guild_id |> to_string()}'\n#{err.response.message}"
        )
    end
  end
```
(`args` is a list of strings)

Converting the string to integer via `String.to_integer/1` does not yield any different results.

The same function (which calls the same API endpoint) in discord.py library (which I used to verify it was not a Discord issue), on the other hand, does work as expected.

Andris0:
Fixed with 401443001bc2ca6227d8d189dad72400b80ffdf7.

PR #165, opened 2020-03-15
Leaving guild fails with 400: Bad Request (#164)

Andris0:
Default headers [`[{"content-type", "application/json"}]`](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/api.ex#L2833) added by `Nostrum.Api.request/4` causes `Nostrum.Api.leave_guild/1` to fail with `400: Bad Request` response.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -2427,7 +2427,13 @@ defmodule Nostrum.Api do
   """
   @spec leave_guild(integer) :: error | {:ok}
   def leave_guild(guild_id) do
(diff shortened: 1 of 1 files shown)


Kraigie:
Any insight into why this is the case? Do other delete methods fail? Regardless, thanks for the quick fix!

Andris0:
Tested 7/8 other delete methods (except `Nostrum.Api.delete_guild_integrations/2`), work just fine.

Kraigie:
Cool, thanks!

PR #166, opened 2020-05-14
Add flag support to Nostrum for user public_flags

jos-b:
Hello Craig, hope you are safe.

I've implemented flags onto the user struct to allow people to see who is staff.

```elixir
iex> {:ok, %{public_flags: flags}} = Nostrum.Cache.UserCache.get(165023948638126080)
...
iex> flags
%Nostrum.Struct.Flags{
  bug_hunter_level_1: false,
  bug_hunter_level_2: false,
  early_supporter: true,
  hypesquad_balance: true,
  hypesquad_bravery: false,
  hypesquad_brilliance: false,
  hypesquad_events: false,
  partner: true,
  staff: false,
  system: false,
  team_user: false,
  verified_bot: false,
  verified_developer: true
}
iex> Nostrum.Struct.Flags.to_integer(flags)
131842
```

Thanks!

### DIFF
--- a/lib/nostrum/struct/flags.ex
+++ b/lib/nostrum/struct/flags.ex
@@ -1,0 +1,194 @@
+defmodule Nostrum.Struct.Flags do
+  @moduledoc """
+  Struct representing the flags a user account can have
+  """
+
+  import Bitwise
+
+  defstruct staff: false,
+            partner: false,
+            hypesquad_events: false,
+            bug_hunter_level_1: false,
+            hypesquad_bravery: false,
+            hypesquad_brilliance: false,
+            hypesquad_balance: false,
+            early_supporter: false,
+            team_user: false,
+            system: false,
+            bug_hunter_level_2: false,
+            verified_bot: false,
+            verified_developer: false
+
+  @typedoc """
+  Discord Employee
+  """
+  @type staff :: bool
+
+  @typedoc """
+  Discord Partner
+  """
+  @type partner :: bool
+
+  @typedoc """
+  HypeSquad Events
+  """
+  @type hypesquad_events :: bool
+
+  @typedoc """
+  Bug Hunter (Level 1)
+  """
+  @type bug_hunter_level_1 :: bool
+
+  @typedoc """
+  HypeSquad Bravery
+  """
+  @type hypesquad_bravery :: bool
+
+  @typedoc """
+  HypeSquad Brilliance
+  """
+  @type hypesquad_brilliance :: bool
+
+  @typedoc """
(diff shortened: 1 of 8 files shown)


lib/nostrum/struct/user.ex
```diff
@@ -63,6 +64,9 @@ defmodule Nostrum.Struct.User do
   @typedoc "The user's email"
   @type email :: String.t() | nil
 
+  @typedoc "The user's public flags"
+  @type public_flags :: Flags.t() | nil
```
skippi, line 68: Although the `nil` here matches discord's documentation, it doesn't quite match with the PR's code, which suggests that this value will never be `nil` (if I'm not mistaken). I'd remove the `nil` here to simplify pattern matching.

lib/nostrum/struct/flags.ex
```diff
@@ -0,0 +1,194 @@
...
+  iex> Nostrum.Struct.Flags.to_integer(my_flags)
+  131842
+  ```
+  """
+  @spec to_integer(t) :: integer()
+  def to_integer(flag_struct) do
+    booleans =
+      flag_struct
+      |> Map.to_list()
+      |> List.delete({:__struct__, Nostrum.Struct.Flags})
```
skippi, line 185: I'd recommend using `Map.from_struct/1`.

lib/nostrum/struct/user.ex
jos-b: Addressed in aff5d4ca76ec14983946c67521b8f03850f0a78c

lib/nostrum/struct/flags.ex
jos-b: Addressed in 2d4d439c6b77c8f2c6179a9b43a94184e93b9a28

lib/nostrum/struct/user.ex
```diff
@@ -153,6 +158,7 @@ defmodule Nostrum.Struct.User do
       map
       |> Map.new(fn {k, v} -> {Util.maybe_to_atom(k), v} end)
       |> Map.update(:id, nil, &Util.cast(&1, Snowflake))
+      |> Map.update(:public_flags, 0, &Flags.from_integer(&1))
```
Kraigie, line 161: If we set this to 0 here as a default then the spec will be wrong. I think we should leave the integer as it is, and if users want to act on the flags then they can use your function to do so.

jos-b: Craig! Thank you for the review. Addressed in 3cdff15fc4f6658d239630b8d669ffdc458fdeb8

PR #172, opened 2020-07-30
Updated outdated Supervisor example

latesh:


examples/event_consumer.ex
```diff
@@ -1,10 +1,17 @@
...
+  
+  def start_link(args) do
+    Supervisor.start_link(__MODULE__, args, name: __MODULE__)
+  end
 
-    children = [ExampleConsumer]
+  @impl true
+  def init(_init_arg) do
+    children = [
+      {ExampleConsumer, []}
```
Kraigie, line 11: Do you think we should maybe change this line to just be `ExampleConsumer`? IMO it's more idiomatic and could be less confusing for those who are new to supervision trees (since they're the most likely people to use this kind of example).

latesh: Yes, that makes sense as the consumer doesn't accept any init args anyway. Also ```children = {ExampleConsumer, []}``` is equal to ```children = [ExampleConsumer]```

Kraigie:
Thanks!

PR #173, opened 2020-08-06
Fixes Logger deprecation warning.

barkerja:
As of Elixir 1.10, any values other than strings, lists or maps passed to Logger results in a deprecation warning.

Example: warning: passing :PRESENCE_UPDATE to Logger is deprecated, expected a binary or an iolist


Kraigie:
Thanks!

PR #176, opened 2020-08-24
Update configs

Awlexus:
Adds a way to add secret configs during testing.

I also realized that Nostrum was still using `Mix.Config`, which have been deprecated, so I updated that as well


Kraigie:
Thanks!

PR #177, opened 2020-08-25
Updated `cache.ex` example

Andris0:
The title is indeed accurate, can confirm, I witnessed it.

### DIFF
--- a/examples/cache.ex
+++ b/examples/cache.ex
@@ -23,15 +23,14 @@ defmodule ExampleConsumer do
     Consumer.start_link(__MODULE__)
   end
 
-  # We only need to write specific event handlers for the events we are interested in,
-  # which in this case, will be the handling of new Discord messages,
-  # for other cases, there is a catch all handler at the end of the module.
+  # We only need to write event handlers for the events we are interested in,
+  # the rest will go to the catch-all case to be ignored.
   def handle_event({:MESSAGE_CREATE, message, _ws_state}) do
     ExampleCommands.command(message)
   end
 
-  # The catch all event case that takes and ignores all events you are not
-  # interested in, if you do not have this, or have not defined literally
+  # The catch-all event case that takes the rest of the events.
+  # If you do not have this, or have not defined literally
   # every event case yourself (what an absolute madlad if you have),
   # the consumer will crash, not having any function signature to match upon.
   # By the way, the return atom stands for "no-op", shorthand for "no operation",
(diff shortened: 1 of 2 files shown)


examples/cache.ex
```diff
@@ -1,74 +1,125 @@
...
-    Consumer.start_link(__MODULE__, :state)
+    Consumer.start_link(__MODULE__)
   end
 
-  def handle_event({:MESSAGE_CREATE, {msg}, ws_state}, state) do
-    # Let's do a little command 'parsing'
-    # We first grab the content, split it into an array and then convert the array to a tuple
-    args = msg.content |> String.split(" ") |> List.to_tuple()
+  # We only need to write specific event handlers for the events we are interested in,
+  # which in this case, will be the handling of new Discord messages,
```
Kraigie, line 27: I think this is a bit of a run on sentence

examples/cache.ex
```diff
@@ -1,74 +1,125 @@
...
   end
 
-  # Default event handler, if you don't include this, your consumer WILL crash if
-  # you don't have a method definition for each event type.
-  def handle_event(_, state) do
-    {:ok, state}
+  # The command that will be handling our message.
+  # Pattern matching's pretty neat by the way, a nice way to define a shape for the data we accept
+  # and have it reject the rest, also gets rid of a good chunk of all that pesky condition checking
+  # that you would normally see, trying for that data shape that we need, destructuring's also pretty neat.
```
Kraigie, line 67: Bit of a run-on here as well

examples/cache.ex
```diff
@@ -1,74 +1,125 @@
...
+  def command(%{
+        content: @prefix <> "userinfo " <> message_user_id,
+        channel_id: channel_id,
+        guild_id: guild_id
+      })
+      when guild_id != nil do
+    # Our happy path for sending a complete userinfo response.
+    # Pretty standard stuff, maybe apart from additionally wrapping both sides in a tuple
+    # with an extra identifier, that's just to make failure cases more obvious.
+    with {{user_id, _binary}, :parse_id} when is_snowflake(user_id) <-
```
Kraigie, line 79: I don't think this is a bad example, but given this will probably be used by beginners mostly (I think?) something a little more simple might be better. I think seeing all the curly braces might discourage some people.

Kraigie: Maybe we just need to get rid of the resource portion?

examples/cache.ex
```diff
@@ -1,74 +1,125 @@
...
+      {{:error, reason}, resource} ->
+        Api.create_message(channel_id, "Unable to retrieve #{resource}: #{inspect(reason)}")
+    end
   end
+
+  # Just like handling events, a fall through `command` case to catch and ignore unwanted cases,
+  # otherwise it will go boom, KABLAM, "OH GOD NO!", *insert mushroom cloud here*, "AAAAH, MY FACE,
+  # IT'S MELTING!", well not really, since we have supervision trees, fault tolerance and the absolute,
+  # massive powerhouse of awesomeness that the BEAM is and all that jazz, but you know, you catch my drift.
+  # Also, "noop", *giggles*.
```
Kraigie, line 123: I actually really enjoy the lighthearted tone throughout this example, but I think this paragraph overall might just be a *tad* too much. 😉

Kraigie:
Perfect, thanks!

PR #178, opened 2020-08-30
Application and adjacent module cleanup

Andris0:
+ A minor cleanup for Application and Dummy testing modules
+ Specific `raise` case for fetching gateway info with invalid token (first authenticated API request)
+ Integrity check for configured shard count


Kraigie:
Thanks!

Issue #167, opened 2020-05-25
Consumer.start_link types


amiralies:
I've a consumer (the example from readme).
Dialyzer says start_link has no local return

```elixir
  def start_link do
    Nostrum.Consumer.start_link(__MODULE__)
  end
```

jchristgit:
I can't reproduce this on master. I assume that it's probably been fixed by something else, but if not, could you re-open? I also had no Dialyzer complaint on my bot regarding this.

PR #179, opened 2020-09-14
Fix error on README.md example.

DarkCode01:
Update example on README.md.

Issue Closed: #167


DarkCode01:
Updated folder `examples` and create a simple project using cache and event consumer on project example.

jb3 (changes requested):
This feels like it could be a good improvement, but the logic is in the wrong place right now.

The `command`, `get_cached_with_fallback` and so on should all be inside the cached consumer, holding them at the root application level is bad practice and makes things hard to find.

Let's have a fully self-contained consumer without using the cache in `examples/event_consumer/lib/event_consumer/basic_consumer.ex` and one in `examples/event_consumer/lib/event_consumer/cache_consumer.ex` using the cache functions you've written.

We should also use better naming, let's switch to `EventConsumer.BasicConsumer` and `EventConsumer.CacheConsumer`. We can then move the application logic into `EventConsumer` instead of `EventConsumer.Application`, removing a file and making things simpler. In my eyes we should have 3 files:
- Root application 
- Cache consumer
- Basic consumer

I've also left a few minnor comments where I think improvements could be made.

jchristgit:
Hi @DarkCode01, are you still interested in working on this issue?

jchristgit:
Thank you for the PR! As we have closed the initial issue and there was no more activity here, I am closing this PR. Feel free to open an issue if you feel like we are missing something!

PR #181, opened 2020-10-01
Add gateway intents

jb3:
Adds support for gateway intents through a `gateway_intents` configuration option.

By default all intents are enabled, but passing the configuration option disables everything and only enables those passed.

There is an additional documentation page summarising how to specify gateway intents in the configuration.


Kraigie:
BIG! Thanks!

jb3:
Very epic, Craig! Thank you! :shipit:

PR #182, opened 2020-10-14
Add convenience keys for gateway_intents config

BrandtHill:
Yes

### DIFF
--- a/lib/nostrum/shard/intents.ex
+++ b/lib/nostrum/shard/intents.ex
@@ -32,7 +32,7 @@ defmodule Nostrum.Shard.Intents do
   @spec get_enabled_intents :: integer()
   def get_enabled_intents do
     # If no intents are passed in config, default to all being enabled.
-    enabled_intents = Application.get_env(:nostrum, :gateway_intents, :all)
+    enabled_intents = Application.get_env(:nostrum, :gateway_intents, :nonprivileged)
 
     case enabled_intents do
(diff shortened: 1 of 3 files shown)


lib/nostrum/shard/intents.ex
```diff
@@ -26,13 +31,18 @@ defmodule Nostrum.Shard.Intents do
...
   @spec get_enabled_intents :: integer()
   def get_enabled_intents do
-    enabled_intents = Application.get_env(:nostrum, :gateway_intents)
-
     # If no intents are passed in config, default to all being enabled.
-    if enabled_intents do
-      get_intent_value(enabled_intents)
-    else
-      get_intent_value(Keyword.keys(intent_values()))
+    enabled_intents = Application.get_env(:nostrum, :gateway_intents, :all)
```
Kraigie, line 35: How do we feel about making the default be `:nonprivileged` so people who don't know about this setting still will have a working bot?

BrandtHill: I would prefer that as it's minimally invasive for existing bots' configs. I just kept it as all to preserve the original author's intention.

Kraigie: 👍 lets change it then

BrandtHill: oui

Kraigie:
Thanks!

PR #183, opened 2020-10-14
Add support for allowed mentions

Awlexus:
Adds `:allowed_mentions` as an option to `create_message`.

~~Shouldn't be merged until Discord API V8 is used by default.~~ Idk why that didn't work during development

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -163,8 +163,26 @@ defmodule Nostrum.Api do
       * `:name` (string) - the name of the file
       * `:body` (string) - binary you wish to send
     * `:embed` (`t:Nostrum.Struct.Embed.t/0`) - embedded rich content
+    * `:allowed_mentions` - See "Allowed mentions" below
 
     At least one of the following is required: `:content`, `:file`, `:embed`.
+
+  ## Allowed mentions
+
+  With this option you can control when content from a message should trigger a ping.
+  Consider using this option when you are going to display user_generated content.
+
+  ### Allowed values
+    * `:all` (default) - Ping everything as usual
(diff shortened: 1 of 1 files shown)


lib/nostrum/api.ex
```diff
@@ -163,9 +163,27 @@ defmodule Nostrum.Api do
...
+    * `:all` (default) - Ping everything as usual
+    * `:none` - Nobody will be pinged
+    * `:everyone` - Allows to ping @here and @everone
+    * `:user` - Allows to ping users
+    * `:roles` - Allows to ping roles
+    * `{:user, list}` - Allows to ping list of users. Can contain up to 100 ids of users.
+    * `{:role, list}` - Allows to ping list of roles. Can contain up to 100 ids of roles.
+    * list - a list containing the values above.
+
+  Passing a list will merge the settings settings provided
```
Kraigie, line 185: `settings settings` ❓

lib/nostrum/api.ex
```diff
@@ -2938,4 +2960,44 @@ defmodule Nostrum.Api do
 
     {:ok, convert}
   end
+
+  def prepare_allowed_mentions(options) do
```
Kraigie, line 2964: Might as well make this `defp` as well IMO

Kraigie (changes requested):

lib/nostrum/api.ex
Awlexus: Sorry, was testing the function and forgot to make it private again

lib/nostrum/api.ex
Awlexus: :confounded:

Kraigie:
Thanks!

PR #184, opened 2020-10-23
ci: migrate to github actions

skippi:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
+name: ci
```
jb3, line 1: This looks cleaner in the GitHub UI with capitals, we should also use a more descriptive name.
```suggestion
name: Test & Lint
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,30 @@
+name: docs
```
jb3, line 1:
```suggestion
name: Publish Docs
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+on: [push]
+jobs:
+  ubuntu:
```
jb3, line 4:
```suggestion
  test_and_lint:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        elixir-version: ['1.9', '1.10']
```
jb3, line 8: Maybe we should be testing on Elixir v1.11, @Kraigie may have an opinion on this. It'll run in parallel so theoretically no major time difference (and we can create up to 256 jobs :smile:)
```suggestion
        elixir-version: ['1.9', '1.10', '1.11']
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-elixir@v1
+        with:
+          otp-version: '22.0'
```
jb3, line 13: It may be worth running checks for OTP 23 as well as 22, which we can add to the matrix, again no major time differences, and also something Craig will have thoughts on. (This may be me overcomplicating things!)

.github/workflows/docs.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  push:
+    branches:
+      - master
+jobs:
+  gh-pages:
```
jb3, line 7:
```suggestion
  docs:
```

jb3 (changes requested):
This is looking fly, a nicely setup GitHub Actions setup.

I've added some comments, however some of them may be me overcomplicating things!

.github/workflows/ci.yml
Kraigie: 👍

jb3 (approved):

Kraigie:
Thanks!

PR #190, opened 2020-10-24
fix: update_status now handles status as :atom

mad42:
Updated `Nostrum.Api.update_status` to reflect `@type status :: :dnd | :idle | :online | :invisible`


Kraigie:
Thanks!

PR #193, opened 2020-10-27
Improve CI speed

jb3:
> Elixir has a non-scary syntax and combines the good features of Ruby and Erlang. It's not Erlang and it's not Ruby and it has ideas of its own.
— Joe Armstrong

This PR drops Elixir 1.9 support.

It also caches the build directory for each matrix job (OTP version + Elixir version). This should hopefully decrease the CI time slightly. 💨


Kraigie:
Thanks!

PR #194, opened 2020-10-27
Add util functions for determining latency of shard connections

BrandtHill:
A super very important feature

### DIFF
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -245,12 +245,12 @@ defmodule Nostrum.Util do
   @doc """
   Gets the latency of the shard connection from a `Nostrum.Struct.WSState.t()` struct.
 
-  Returns the latency in milliseconds as an integer, returning 0 if unknown.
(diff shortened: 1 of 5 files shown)


lib/nostrum/util.ex
```diff
@@ -240,6 +242,38 @@ defmodule Nostrum.Util do
      )}
   end
 
+  @doc """
+  Gets the latency of the shard connection from a `Nostrum.Struct.WSState.t()` struct.
+
+  Returns the latency in milliseconds as an integer, returning 0 if unknown.
+  """
+  @spec get_shard_latency(WSState.t()) :: non_neg_integer
```
Kraigie, line 250: Do you think we should instead return something like `nil`? I think if we return 0 it could be misconstrued as there being actually 0 latency.

Kraigie (commented):
One small question, LGTM otherwise!

Kraigie:
Thanks!

PR #195, opened 2020-10-31
Nostrum.Cache.GuildCache.all/0 fails with EXIT :noproc

jvantuyl:
Title says it all.  The code references the `GuildSupervisor` without its full prefix (as if it was aliased) instead of the full path `Nostrum.Cache.GuildCache.all`.

This compiles because nothing at compile time knows that `GuildSupervisor` won't be a registered name.  At run time, however, the `GuildSupervisor` is named with its full path, so `Supervisor.which_children/1` can't find the process and exits with the expected `:noproc` status.

This fix adds an alias.


Kraigie:
Thanks!

PR #201, opened 2020-11-25
Turn embed into a behaviour

Awlexus:
Basic implementation. If this approach is acceptable I'll add documentation about the usage, before merging.

### DIFF
--- a/lib/nostrum/struct/embed.ex
+++ b/lib/nostrum/struct/embed.ex
@@ -119,12 +119,12 @@ defmodule Nostrum.Struct.Embed do
         }
 
   @callback author(struct) :: author()
-  @callback color(struct) :: color()
+  @callback color(struct) :: integer() | nil
   @callback fields(struct) :: fields()
   @callback description(struct) :: description()
   @callback footer(struct) :: footer()
-  @callback image(struct) :: image()
-  @callback thumbnail(struct) :: thumbnail()
+  @callback image(struct) :: url()
+  @callback thumbnail(struct) :: url()
   @callback timestamp(struct) :: timestamp
   @callback title(struct) :: title()
   @callback url(struct) :: url()
--- a/lib/nostrum/struct/embed.ex
+++ b/lib/nostrum/struct/embed.ex
@@ -296,6 +296,10 @@ defmodule Nostrum.Struct.Embed do
   ```
   """
   @spec put_image(t, Image.url()) :: t
+  def put_image(%__MODULE__{} = embed, nil) do
+    %__MODULE__{embed | image: nil}
+  end
+
   def put_image(%__MODULE__{} = embed, url) do
     image = %Image{
       url: url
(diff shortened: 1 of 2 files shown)


lib/nostrum/struct/embed.ex
```diff
@@ -381,6 +422,46 @@ defmodule Nostrum.Struct.Embed do
     put_field(%__MODULE__{embed | fields: []}, name, value, inline)
   end
 
+  @doc """
+  Create an embed from a struct that implements the protocol `Nostrum.Struct.Embed.Protocol`.
```
Kraigie, line 426: I don't think I'm 100% happy with the wording here. Structs don't necessarily implement a protocol per se, right? It's the module that implements the behavior. I obviously get what you're trying to say but it's not 100% correct and if someone is unfamiliar with behaviors I could see there being issues.

Also, this still mentions a protocol when we're using a behavior now.

Kraigie (commented):
Thanks for the PR! I definitely don't hate this idea and it doesn't seem like it'd cause any harm if it were to be added. Additionally there must be a use case if you're suggesting adding this capability. However, I do have some feedback about it!

I'll be the first to admit I haven't looked closely at behaviors in the past or how they're implemented, so feel free to correct me if I'm wrong on anything. To me it feels like this implementation is too closely coupled to the actual struct module itself, and I'm not sure how we'd fix that.

Take the author field for example. The `author` that their implementation returns needs to have the correct fields (`name`, `url`, `icon_url`). Do you think this is clear? The only indicator is the type spec which now that I'm thinking about it might be enough? 

Additionally, if these fields don't exist the user will not get a useful error. For example, if their `author/1` method doesn't return a term with the `url` field they'll get an error that `:url` doesn't exist in their term. IMO this would be unfriendly inside of a function doing as much as `from/1` is.

lib/nostrum/struct/embed.ex
Awlexus: My bad. I wrote this while I still relied on protocols and forgot to update it 🙏

Awlexus:
Elixir LS should pick up on type mismatching in this case and emit a warning, but we can always add strict pattern matching and error messages to further help devs

Kraigie:
Sorry for the late response, I'm happy to move forward with this. I do think I would like to include the mentioned pattern matching and/or better error messages!

Awlexus:
Added some pattern matching for the callbacks that return structs (author, footer and fields). I wasn't sure whether wanted a check for all other fields as well, so I omitted it for now.

Kraigie:
This looks good! I think some short examples would be nice, but besides that I'd be happy to merge it!

Awlexus:
While adding some short examples with with doc tests I noticed that I call [`put_image/2` and `put_thumbnail/2`](https://github.com/Kraigie/nostrum/pull/201/files#diff-90555f3cc794f87ab01ec93dc1f8d798e23e4dac678aedfd8575632906e884cbR438-R439) with the value `nil` it creates structs with the `url`s set to `nil`. Should I handle this by adding function clauses to the mentioned functions that replace the current value with `nil` or should I just handle this in `from/1`, by checking if the values are `nil`?

jb3 (approved):
Looking good!

Kraigie:
Thanks!

PR #202, opened 2020-12-03
Fix Nostrum.Consumer options

Andris0:
When calling `Nostrum.Consumer.start_link/2` from your own consumer, some option combinations will make it fail to start.

Currently, you can either pass the `name` option for the ConsumerSupervisor
```elixir
# First pattern match
def start_link(mod, [name: name] = options) do
  ConsumerSupervisor.start_link(
    __MODULE__,
    [mod, Keyword.drop(options, [:name])],
    name: name
  )
end
```

Or pass in the rest of the [options](https://kraigie.github.io/nostrum/Nostrum.Consumer.html#t:option/0) for the event handler module
```elixir
# Second pattern match
def start_link(mod, options),
  do:
    ConsumerSupervisor.start_link(
      __MODULE__,
      [mod, options]
    )
```

If `name` and some of the other options get passed together to the `start_link` function, it will fail the first pattern match and put all given options as the event handler module options, making the consumer fail to start with
`** (EXIT) {:bad_opts, "unknown options [name: <whatever_name_was_passed>]"}`

For example, defining the `start_link` function of your own consumer like this is currently not possible
```elixir
def start_link do
  Consumer.start_link(__MODULE__, name: __MODULE__, max_restarts: 0)
end
```

### DIFF
--- a/lib/nostrum/consumer.ex
+++ b/lib/nostrum/consumer.ex
@@ -240,28 +240,21 @@ defmodule Nostrum.Consumer do
   `mod` is the name of the module where you define your event callbacks, which should probably be
   the current module which you can get with `__MODULE__`.
 
-  `options` is a list of general process options. See `t:Nostrum.Consumer.options/0` for more info.
+  `opts` is a list of general process options. See `t:Nostrum.Consumer.options/0` for more info.
   """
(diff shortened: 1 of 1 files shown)


Kraigie:
Yeah this looks like me being a noob a while ago. Thanks!

PR #203, opened 2020-12-05
Add support for reading message references

alexkuang:
Adding support for `message_reference` and `referenced_message` for message objects: https://discord.com/developers/docs/resources/channel#message-object

Per contributing guidelines I ran `mix credo --strict`.  I got some warnings about TODOs, but they are unrelated to my changes.

### DIFF
--- a/lib/nostrum/struct/message.ex
+++ b/lib/nostrum/struct/message.ex
@@ -5,7 +5,7 @@ defmodule Nostrum.Struct.Message do
 
   alias Nostrum.Struct.{Embed, User}
   alias Nostrum.Struct.Guild.Member
-  alias Nostrum.Struct.Message.{Activity, Application, Attachment, Reaction}
+  alias Nostrum.Struct.Message.{Activity, Application, Attachment, Reaction, Reference}
   alias Nostrum.{Snowflake, Util}
 
   defstruct [
@@ -23,9 +23,11 @@ defmodule Nostrum.Struct.Message do
     :mention_everyone,
     :mention_roles,
     :mentions,
+    :message_reference,
     :nonce,
     :pinned,
     :reactions,
+    :referenced_message,
     :timestamp,
     :tts,
     :type,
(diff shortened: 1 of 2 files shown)


lib/nostrum/struct/message.ex
```diff
@@ -107,6 +109,20 @@ defmodule Nostrum.Struct.Message do
   """
   @type member :: Member.t() | nil
 
+  @typedoc """
+  reference data sent with crossposted messages and replies
```
Kraigie, line 113: For consistency could we capitalize the first words of these sentences?

lib/nostrum/struct/message.ex
```diff
@@ -122,6 +138,7 @@ defmodule Nostrum.Struct.Message do
           mention_everyone: mention_everyone,
           mention_roles: mention_roles,
           mentions: mentions,
+          message_reference: message_reference,
```
Kraigie, line 141: Is this missing the new `referenced_message` type?

lib/nostrum/struct/message/reference.ex
```diff
@@ -0,0 +1,40 @@
...
+  @typedoc "id of the originating message"
+  @type message_id :: Snowflake.t()
+
+  @typedoc "id of the originating message's channel"
+  @type channel_id :: Snowflake.t()
+
+  @typedoc "id of the originating message's guild"
+  @type guild_id :: Snowflake.t()
+
+  @type t :: %__MODULE__{
```
Kraigie, line 23: How do we feel about actually using the types from these modules. Like `Nostrum.Struct.Message.id` for example?

Kraigie (commented):
Looks good, just a couple nitpicks!

lib/nostrum/struct/message.ex
alexkuang: Ah, sure.  I just pasted these from the discord docs.

lib/nostrum/struct/message/reference.ex
alexkuang: That makes sense to me.

alexkuang:
@Kraigie Thanks for the feedback.  I added changes addressing your points.  Let me know if I need anything else to get this merged.

lib/nostrum/struct/message/reference.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Nostrum.Struct.Message.Reference do
+  @moduledoc """
+  Struct representing a discord message reference.
+  """
+
+  alias Nostrum.{Snowflake, Util}
+  alias Nostrum.Struct.Message.{Channel, Guild, Message}
```
Kraigie, line 7: I think this needs changed to something like this, right?
```suggestion
  alias Nostrum.Struct.{Channel, Guild, Message}
```

alexkuang: Yep, my bad.

Kraigie:
Thanks!

PR #204, opened 2020-12-06
Add new public guild fields

oliver-ni:
I implemented the new fields `rules_channel_id` and `public_updates_channel_id` that Discord recently added for public (discoverable) guilds.

Works properly in my testing. Per the contributing guidelines, I ran `mix credo --strict`, but the warnings that popped up didn't seem to be related to my changes.

### DIFF
--- a/lib/nostrum/struct/guild.ex
+++ b/lib/nostrum/struct/guild.ex
@@ -53,13 +53,13 @@ defmodule Nostrum.Struct.Guild do
   @type splash :: String.t() | nil
 
   @typedoc "The id of the guild owner"
-  @type owner_id :: Snowflake.t()
+  @type owner_id :: User.id()
 
   @typedoc "The id of the voice region"
   @type region :: String.t()
 
   @typedoc "The id of the guild's afk channel"
-  @type afk_channel_id :: Snowflake.t() | nil
+  @type afk_channel_id :: Channel.id() | nil
 
   @typedoc "The time someone must be afk before being moved"
   @type afk_timeout :: integer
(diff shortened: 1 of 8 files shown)


lib/nostrum/struct/guild.ex
```diff
@@ -113,6 +115,18 @@ defmodule Nostrum.Struct.Guild do
   """
   @type system_channel_id :: Snowflake.t() | nil
 
+  @typedoc """
+  The id of the channel that is used for rules. This is only available to guilds that
+  contain ``PUBLIC`` in `t:features/0`.
+  """
+  @type rules_channel_id :: Snowflake.t() | nil
```
Kraigie, line 122: While this is correct, I think it's more helpful if we link to the remote types in the actual `Nostrum.Struct.Channel` module itself. This way we signify in the type that this is an `id` coming from a channel and the user is only one click away from going to check out the type and any intricacies it may have!

lib/nostrum/struct/guild.ex
```diff
@@ -113,6 +115,18 @@ defmodule Nostrum.Struct.Guild do
...
+
+  @typedoc """
+  The id of the channel where admins and moderators receive notices from Discord. This
+  is only available to guilds that contain ``PUBLIC`` in `t:features/0`.
+  """
+  @type public_updates_channel_id :: Snowflake.t() | nil
```
Kraigie, line 128: Same as above! ^^

Kraigie (approved):
Looks good, just one small nitpick!

lib/nostrum/struct/guild.ex
oliver-ni: OK, I changed it everywhere I could find

Kraigie:
Thanks!

PR #206, opened 2020-12-19
Fix logging issue when not all gateway intents present

BrandtHill:


jb3 (approved):
Smart! I like it.

Kraigie:
Thanks!

PR #208, opened 2020-12-23
Member caching improvements

jb3:
Resolves some hiccups from the way that members are sent from the gateway when using intents.

We no longer receive the members in the `GUILD_CREATE` event unless we enable the presences intent, we can however still request them directly with every guild that we get a `GUILD_CREATE` object for.

We already have the logic in place for this since it's the same as the old logic for handling large guilds, it's just that now we trigger it under the large threshold providing that the `request_guild_members` config option is set to true.

N.B. Some weird commits showed up in this PR, from my quick view I can't see that they will make any changes but I'd advise squashing when merging in order to maintain a clean Git history.


lib/nostrum/shard/dispatch.ex
```diff
@@ -111,8 +113,15 @@ defmodule Nostrum.Shard.Dispatch do
       :ets.insert(:channel_guild_map, {channel.id, guild.id})
     end)
 
-    if guild.member_count >= @large_threshold and
-         Application.get_env(:nostrum, :request_guild_members, false) do
+    has_members = (Intents.get_enabled_intents() &&& 1 <<< 1) == 1 <<< 1
```
Kraigie, line 116: Do you think we should create some helper in the `Intents` module to prevent having to include these messy details whenever we want to check intents?

Kraigie (commented):
Just a small question, but looks good otherwise!

lib/nostrum/shard/dispatch.ex
jb3: Resolved with introduction of `Intents.has_intent?/1` in 08e886864b38bb2f096e486abf328c6d00374f0e

jchristgit (approved):
:shipit:

Kraigie:
Thanks!

PR #209, opened 2020-12-28
Use `options` type instead of `keyword | map`.

### LINKED ISSUE #49: Replacing `keyword | map` with the `options` type in the Api module
This is just a reminder for our recent Api module changes. With the recent PRs ( #48 #46 ) changing up the Api, we may forget to do this in places where it is applicable.

## Known Inconsistencies

* User related Api functions

jchristgit:
Only two remainders found. Fixed here.

Closes #49.


Kraigie:
Thanks!

PR #217, opened 2021-01-12
Make `recon` dependency optional.

jchristgit:
This is not required by users of the library, but can be useful when we are
debugging the library ourselves.


jchristgit:
@jb3 This was signed off by Craig in our development chat, feel free to merge yourself if it looks good.

PR #221, opened 2021-02-01
Add options to Nostrum.Voice.play.

BrandtHill:
Adds options to play function to select if you want real-time processing (performance tradeoffs), selecting a different start position, and selecting a custom duration.

### DIFF
--- a/lib/nostrum/struct/voice_state.ex
+++ b/lib/nostrum/struct/voice_state.ex
@@ -69,7 +69,9 @@ defmodule Nostrum.Struct.VoiceState do
     end
 
     unless is_nil(v.session_pid) do
-      Session.close_connection(v.session_pid)
+      if Process.alive?(v.session_pid) do
+        Session.close_connection(v.session_pid)
+      end
     end
 
     :ok
--- a/lib/nostrum/voice.ex
+++ b/lib/nostrum/voice.ex
@@ -87,8 +87,17 @@ defmodule Nostrum.Voice do
       - `:url` Input will be [any url that `ffmpeg` can read](https://www.ffmpeg.org/ffmpeg-protocols.html).
       - `:pipe` Input will be data that is piped to stdin of `ffmpeg`.
       - `:ytdl` Input will be url for `youtube-dl`, which gets automatically piped to `ffmpeg`.
+    - `options` - See options section below.
+
 
   Returns `{:error, reason}` if unable to play or a sound is playing, else `:ok`.
+
+  ## Options
+    - `:start_pos` (string) - The start position of the audio to be played. Defaults to beginning.
+    - `:duration` (string) - The duration to of the audio to be played . Defaults to entire duration.
+    - `:realtime` (boolean) - Make ffmpeg process the input in realtime instead of as fast as possible. Defaults to true.
+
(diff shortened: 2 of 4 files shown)


jchristgit (approved):

lib/nostrum/struct/voice_state.ex
```diff
@@ -69,7 +69,9 @@ defmodule Nostrum.Struct.VoiceState do
     end
 
     unless is_nil(v.session_pid) do
-      Session.close_connection(v.session_pid)
+      if Process.alive?(v.session_pid) do
```
jchristgit, line 72: This is new. For when is this guard introduced? When `ffmpeg` died?

jchristgit (approved):
I can't say anything about the voice session changes, but apart from that it looks good.

jchristgit:
I do have one small suggestion though: Could we have the `volume` option just put its value into the `filter` option? That way the volume setting would not be overwritten.

lib/nostrum/voice.ex
```diff
@@ -87,15 +87,38 @@ defmodule Nostrum.Voice do
...
+    - `options` - See options section below.
+
 
   Returns `{:error, reason}` if unable to play or a sound is playing, else `:ok`.
 
+  ## Options
+    - `:start_pos` (string) - The start position of the audio to be played. Defaults to beginning.
+    - `:duration` (string) - The duration to of the audio to be played . Defaults to entire duration.
+    - `:realtime` (boolean) - Make ffmpeg process the input in realtime instead of as fast as possible. Defaults to true.
+    - `:volume` (number) - The output volume of the audio. Default volume is 1.0.
```
Kraigie, line 99: Do you think there's any value in normalizing this number to be more equivalent to how most people think about volume? I'm not too familiar with ffmpeg but saying `1.0` or `0.5` is less friendly than your typical `0-100` scale. Not sure what values we would have the `max` and `min` be but food for thought!

Kraigie (approved):
One comment, but nothing blocking. Thanks for all the voice work you're doing!

lib/nostrum/struct/voice_state.ex
BrandtHill: This is for when the voice gateway connection may have already closed. This change and changing a call to a cast in the Session module are to prevent an error that can occur when leaving a voice channel.

lib/nostrum/voice.ex
BrandtHill: The reason I have it this way is that there are no limits to the `ffmpeg` values. 0.75 would slightly lower volume, 2.0 would add some light saturation, 1000 would hyper distort it (think bass-boosted memes), -1.0 would reverse polarity, etc. This means you have all the flexibility of `ffmpeg`, and someone writing a bot could choose how to use the values, e.g. limit values from 0 to 100 and map them to 0 to 1.0.

BrandtHill:
I've added the volume option to the end of the filter chain as per @jchristgit's recommendation.
I've added more explanation in the docs for the acceptable volume values and their expected results.

jchristgit (approved):
:shipit:

jchristgit:
Thanks!

jb3:
Thanks!

SamOphis:
Thanks!

PR #222, opened 2021-02-02
Pull `:GUILD_{AVAILABLE,CREATE}` payload out of tuple.

### LINKED ISSUE #220: GUILD_AVAILABLE/GUILD_CREATE wrong type spec/akward event/unecessary encapsuled in tuple
the type definition for the events say u should get a guild struct
```elixir
  @type guild_create :: {:GUILD_CREATE, new_guild :: Nostrum.Struct.Guild.t(), WSState.t()}
  @type guild_available :: {:GUILD_AVAILABLE, new_guild :: Nostrum.Struct.Guild.t(), WSState.t()}
```

but the event actually gives u a tuple with the guild struct inside which makes no sense as far as i can see
https://github.com/Kraigie/nostrum/blob/22d30ffe6a14ee3d6f7c0779a39e3e0b1e813454/lib/nostrum/shard/dispatch.ex#L115

so its actually {EVENT,{GUILD},WSSTATE}
```elixir
[error] Task #PID<0.948.0> started from #PID<0.468.0> terminating
** (ArgumentError) argument error
    :erlang.apply({%Nostrum.Struct.Guild{afk_channel_id: nil, a
```

jchristgit:
The tuple sent here only had a single element, whilst being typespecced as
directly passing through the guild struct.

Fixes #220.

### DIFF
--- a/lib/nostrum/shard/dispatch.ex
+++ b/lib/nostrum/shard/dispatch.ex
@@ -134,7 +134,7 @@ defmodule Nostrum.Shard.Dispatch do
         :noop
 
       {:ok, g} ->
-        {check_new_or_unavailable(g.id), {g}, state}
+        {check_new_or_unavailable(g.id), g, state}
     end
   end
 


Kraigie:
LGTM! 🚀

I think alternatively we could implement the struct for this event like we have done with [message deletes](https://github.com/Kraigie/nostrum/blob/9069ac2db2b0272b2f86d375c21df3a9fc66c6f2/lib/nostrum/struct/event/message_delete.ex). We only have a couple of them but I think overall it's a better process. Up to you though!

Kraigie (approved):

jchristgit:
> I think alternatively we could implement the struct for this event like we have done with message deletes. We only have a couple of them but I think overall it's a better process. Up to you though!

Ohh, that's a good idea. In the case of these events though, we always have the complete guild object, don't we? I'm not sure if it would make sense to do that here. I'll merge as-is, but we can revisit it later

jchristgit:
Hex release

Kraigie:
TBH I think it's less about what data is there than it is about being consistent. I think it's nice to have the separate module doc that describes when the event happens and such. But yeah lets revisit it later!

PR #223, opened 2021-02-06
Move guild cache supervisor below individual shard supervisors.

jchristgit:
Per-shard caching - it's happening.

image

This moves the `GuildSupervisor` responsible for managing the guild gen servers per-guild below individual shards.

Sadly I don't have any bot that requires more than one shard to test this with. I would be very happy if somebody could test multiple shards with this.


jchristgit:
I'm a bit confused on how to manage the guild cache test that is failing. Since the shard supervisor also contains the shard session connecting to Discord, I believe this would require a valid token to test with. Or should we mock the supervisor states?

jchristgit:
I have removed the failing test to make the test suite pass

jchristgit:
Guild cache will be moved to ETS, so I'm closing this.

PR #227, opened 2021-02-22
Fix bug where voice supervisor stopped when voice connection drops

BrandtHill:


lib/nostrum/voice/session.ex
```diff
@@ -108,7 +108,13 @@ defmodule Nostrum.Voice.Session do
 
   def handle_info({:gun_ws, _conn, _stream, {:close, errno, reason}}, state) do
     Logger.info("Voice websocket closed (errno #{errno}, reason #{inspect(reason)})")
-    Voice.remove_voice(state.guild_id)
+
+    if errno == 4006 do
+      channel_id = Voice.get_channel_id(state.guild_id)
+      Voice.leave_channel(state.guild_id)
+      Voice.join_channel(state.guild_id, channel_id)
+    end
```
jchristgit, line 116: The `gun_error` change looks good to me. Could you add a comment here explaining why we rejoin on `4006`? Apart from that, I'm not very knowledged in our voice code, so someone else should take a look here.

jchristgit (approved):

lib/nostrum/voice/session.ex
BrandtHill: `4006` means the session is no longer valid so forcing new session.

jchristgit (approved):

Kraigie:
Thanks!

PR #228, opened 2021-02-25
Guns up

BrandtHill:
Make gun always retry shard and voice connections

### DIFF
--- a/lib/nostrum/shard/session.ex
+++ b/lib/nostrum/shard/session.ex
@@ -15,6 +15,8 @@ defmodule Nostrum.Shard.Session do
   @timeout_connect 10_000
   # Maximum time the websocket upgrade may take, in milliseconds.
   @timeout_ws_upgrade 10_000
+  # Options to pass to :gun.open/3
+  @gun_opts %{protocols: [:http], retry: 1_000_000}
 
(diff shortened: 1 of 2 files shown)


queer:
:gun: pew pew

jchristgit (changes requested):
Very interesting.

But isn't this against OTPs design principles? Wouldn't this hang the nostrum application forever if we don't manage to start the shard session?

BrandtHill:
The shard session genserver will already retry to establish a connection forever.
Currently, the shard's `gun` process will stop trying to reconnect if connectivity is lost long enough for `gun`'s default of `5` retries to be expended. In this case, every part of Nostrum will stay alive, including the shard session genserver, but no events will be received on that shard as the websocket connection is dead (I should think this is not intended behavior).

@Kraigie found [this snippet for genserver gun usage](https://github.com/GenericJam/websocket-client/blob/master/lib/websocket_client/websocket_client.ex#L176) which may work for auto reconnects (what happens to the original supervised gun process when we call `ws_upgrade` at the `gun_down` event? idk). This is really a bugfix more than an enhancement.

jchristgit (approved):
I see. Thank you for the explanation

Kraigie:
Thanks!

PR #231, opened 2021-03-06
Switch porcelain dependency to git

BrandtHill:
Porcelain isn't being actively developed, and v2.0.3 in hex.pm is nearly 5 years old. Master branch on GitHub fixes all deprecation warnings.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -93,7 +93,7 @@ defmodule Nostrum.Mixfile do
       {:poison, "~> 3.0"},
       {:gun, "~> 1.3"},
       {:kcl, "~> 1.3"},
-      {:porcelain, "~> 2.0"},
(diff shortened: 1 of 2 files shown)


jchristgit (commented):
I am kind of 50/50 on this, since I'm not sure whether `git` is a requirement for using hex in the first place, therefore some people may not have it installed on their workstations.

I can understand your point on deprecation warnings, however, if the library is not actively developed, should we not switch to another library instead?

Kraigie:
Throwing an additional wrench into the plan here - I'm not sure you can push a package to hex with a git dependency. I know this was the case in the past at least.

BrandtHill:
I’m not too concerned with this, it’s just to fix the deprecation warnings. If alco ever creates a new hex release the problem goes away.

PR #240, opened 2021-05-18
Bump `:gun` to 2.0.0-rc.1.

### LINKED ISSUE #236: Update (or relax) dependencies
Hello everyone! Just tried the lib today by creating a brand new Phoenix app and I got the following error:

```
Failed to use "cowlib" (version 2.9.1) because
  cowboy (version 2.8.0) requires ~> 2.9.1
  gun (versions 1.3.0 to 1.3.2) requires ~> 2.6.0
  mix.lock specifies 2.9.1


Failed to use "cowlib" (version 2.9.1) because
  cowboy (version 2.8.0) requires ~> 2.9.1
  gun (version 1.3.3) requires ~> 2.7.0
  mix.lock specifies 2.9.1
```

Do you think we could have this updated?
 
PS.: Out of curiosity: Is the `gun` lib really necessary if using `httpoison`?

jchristgit:
This prevents dependency problems when trying to use nostrum with e.g. Phoenix,
as outlined in #236. It also sets the foundation for using `:gun` as our one
and only HTTP client, instead of relying on an extra library here.

Fixes #236.


mix.exs
```diff
@@ -91,7 +91,7 @@ defmodule Nostrum.Mixfile do
     [
       {:httpoison, "~> 1.7"},
       {:poison, "~> 3.0"},
-      {:gun, "~> 1.3"},
+      {:gun, github: "ninenines/gun", tag: "2.0.0-rc.1"},
```
Kraigie, line 94: I'm not sure we can do a hex release with a GH dependency. If we can't we should try and see if one of the other available versions from hex would work. It looks like hex is currently hosting the rc version you have here.

Kraigie (approved):
One small comment we should probably look into, LGTM otherwise!

mix.exs
jchristgit: Craig, you are absolutely right. I've pinned this to the hex pre-release instead.

jchristgit:
Thanks!

PR #241, opened 2021-05-19
Simplify await_up logic.

jchristgit:
Gun 2.0 allows us to do this.

### DIFF
--- a/lib/nostrum/shard/session.ex
+++ b/lib/nostrum/shard/session.ex
@@ -61,7 +61,7 @@ defmodule Nostrum.Shard.Session do
     {:ok, worker} = :gun.open(:binary.bin_to_list(gateway), 443, @gun_opts)
     {:ok, :http} = :gun.await_up(worker, @timeout_connect)
     stream = :gun.ws_upgrade(worker, @gateway_qs)
-    await_ws_upgrade(worker, stream)
+    {:upgrade, ["websocket"], _} = :gun.await(worker, stream, @timeout_ws_upgrade)
 
     zlib_context = :zlib.open()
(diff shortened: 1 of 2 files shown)


lib/nostrum/shard/session.ex
```diff
@@ -61,7 +61,7 @@ defmodule Nostrum.Shard.Session do
     {:ok, worker} = :gun.open(:binary.bin_to_list(gateway), 443, @gun_opts)
     {:ok, :http} = :gun.await_up(worker, @timeout_connect)
     stream = :gun.ws_upgrade(worker, @gateway_qs)
-    await_ws_upgrade(worker, stream)
+    {:upgrade, ["websocket"], _} = :gun.await(worker, stream, @timeout_ws_upgrade)
```
Kraigie, line 64: Are there alternative return values for this function now we should handle? Previously we had a whole section of code that provided detailed errors when things went awry.

jchristgit: Only `:gun_error` and a timeout, which will cause an exit with badmatch.

jb3 (approved):
Tested & works fine, good stuff.

Only comment is that instead of exiting out with badmatches we might want to try catch those and reshape them into descriptive nostrum errors about failures with Discord Gateway, but this is optional (no matter what the error is it'd be fatal).

jchristgit:
@Kraigie How do we continue here? Do we close, since we will likely have to go back to gun 2.0 to allow making a new release, or do we merge?

jchristgit:
I am closing this for now. The simplification is nice, but we don't want to block a new nostrum release on a new `:gun` release.

PR #243, opened 2021-05-27
Fix rate limiting broken by casing of x-ratelimit-xxx headers

RealVidy:
It looks like discord is now sending lowercased rate limiting headers (e.g. "x-ratelimit-global" instead of "X-RateLimit-Global"), which breaks the current code.

With this PR, I am making sure this doesn't happen again by forcing all processed headers to be lowercased no matter what. I am also adding a `headers_to_keep` MapSet to improve filtering speed. Indeed, we now get all our headers in one pass compared to 5 with the previous `List.keyfind` solution.

Sampled headers within the `HTTPoison.Response` in `Nostrum.Api.Ratelimiter.handle_headers`:

```
  {"Date", "Thu, 27 May 2021 16:52:09 GMT"},
  {"Content-Type", "text/html; charset=utf-8"},
  {"Content-Length", "0"},
...
  {"x-ratelimit-bucket", "c7f060a5d9790cb6f53000110a7b7edf"},
  {"x-ratelimit-limit", "10"},
  {"x-ratelimit-remaining", "5"},
  {"x-ratelimit-reset", "1622134339"},
  {"x-ratelimit-reset-after", "9"},
```

### DIFF
--- a/lib/nostrum/api/ratelimiter.ex
+++ b/lib/nostrum/api/ratelimiter.ex
@@ -77,16 +77,22 @@ defmodule Nostrum.Api.Ratelimiter do
   defp handle_headers({:error, reason}, _route), do: {:error, reason}
 
   defp handle_headers({:ok, %HTTPoison.Response{headers: headers}} = response, route) do
-    global_limit = headers |> List.keyfind("X-RateLimit-Global", 0)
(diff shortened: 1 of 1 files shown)


jchristgit (approved):
Tested this locally. Thank you very much!

jchristgit:
Thanks!

PR #244, opened 2021-05-27
Fix `gun_down` crash caused by gun 2.0 migration

RealVidy:
The `UnprocessedStreams` parameter was removed in gun 2.0.

[gun_down in 1.3](https://ninenines.eu/docs/en/gun/1.3/manual/gun_down/)
[gun_down in 2.0](https://ninenines.eu/docs/en/gun/2.0/manual/gun_down/)


jchristgit (approved):
Excellent, thanks for fixing this!

PR #246, opened 2021-05-27
Remove overly cautious "sanity wait" for rate limiting

RealVidy:
I think we all want stability in a library like Nostrum so I understand why a sanity wait was added in the first place. 

However, 500ms seems like a very long time to wait in addition to an accurate retry-after. Furthermore, the time it took for the original request to come back with the rate limiting header in addition to the time it will take for the retried request to reach discord servers (let's say 50ms for each at best) should be more than enough to guarantee we are well over the retry-after limitation.

If you are not comfortable removing it entirely, then let us please consider lowering it to something like 100ms.


Kraigie (approved):
Yeah, this looks good to me. If there is a problem down the line we can reassess it then!

PR #247, opened 2021-05-28
Fix a bug that caused 404 messages when trying to bulk delete messages

Awlexus:


jchristgit (approved):
amazing change, thank you @Awlexus

jchristgit:
Thanks!

PR #254, opened 2021-06-24
Do not cast all incoming keys to atoms.

### LINKED ISSUE #252: Memory leak, unsafe conversion of strings to atoms
I am writing an interaction handler with an option type 6 of user. I noticed there's a potential DoS issue here:
```elixir
%Nostrum.Struct.Interaction{
  channel_id: 410902161292263426,
  data: %{
    id: 856724084574781462,
    name: "balance",
    options: [%{name: "user", type: 6, value: "170221514245668831"}],
    resolved: %{
      members: %{
        "170221514245668831": %{
          avatar: nil,
          is_pending: false,
          joined_at: "2018-02-07T20:58:29.506000+00:00",
          nick: nil,
          pending: false,
          permissions: "137438953471",
          premium_since: nil,
          roles: [snip]
        }
      },
      users: %{
        "170221514245668831": %{
          avatar: "a_be0f95eecc54134bafb8e6019a7edad6",
          discriminator: "3333",
          id: 170221514245668831,
          public_flags: 576,
          username: "Hoss"
        }
      }
    }
  },
  guild_id: 410902161292263424,
  id: 856736335565815819,
  member: %Nostrum.Struct.Guild.Member{},
  token: "snip",
  type: 2,
  version: 1
}
```

It took me a while to notice, but the keys for the users and members maps inside of the resolved key are atoms instead of strings as they should be.

jchristgit:
This is only adjusted in the session and voice session dispatcher. That said,
it should be compatible with the current code: atoms are transformed in most
`to_struct` methods via `maybe_to_atom` anyways. In order to fix an atom leak,
the following changes are introduced additionally:

- Adding new keys from the API to interaction structs.
- Adding the `ApplicationCommandInteractionDataResolved` struct.

This may or may not break some things. I will test it with
[bolt](https://github.com/jchristgit/bolt) to look out for any problems, but if
I inferred everything correctly, it would work just fine with existing code.

Note that the new way of converting incoming data to structs also allows us to
slowly phase out `maybe_to_atom`, with the drawback that we need to specify
parsed keys manually.

Fixes #252.

### DIFF
--- a/lib/nostrum/shard/event.ex
+++ b/lib/nostrum/shard/event.ex
@@ -3,13 +3,10 @@ defmodule Nostrum.Shard.Event do
 
   alias Nostrum.Shard.Payload
   alias Nostrum.Shard.Stage.Producer
-  alias Nostrum.Util
 
   require Logger
 
   def handle(:dispatch, payload, state) do
-    payload = Util.safe_atom_map(payload)
-
     if Application.get_env(:nostrum, :log_dispatch_events),
       do: payload.t |> inspect() |> Logger.debug()
 
--- a/lib/nostrum/cache/guild/guild_server.ex
+++ b/lib/nostrum/cache/guild/guild_server.ex
@@ -202,7 +202,7 @@ defmodule Nostrum.Cache.Guild.GuildServer do
 
   def handle_call({:update, :voice_state, guild_id, vsu}, _from, %{voice_states: vs_list} = state) do
     # Remove heavy and duplicate data from voice state
-    vsu = vsu |> Map.delete(:member)
+    vsu = vsu |> Map.delete("member")
     vs_list = vs_list |> Enum.reject(fn v -> v.user_id == vsu.user_id end)
 
     new_voice_states =
--- a/lib/nostrum/consumer.ex
+++ b/lib/nostrum/consumer.ex
@@ -114,7 +114,7 @@ defmodule Nostrum.Consumer do
            {guild_id :: integer, old_emojis :: [Nostrum.Struct.Emoji.t()],
             new_emojis :: [Nostrum.Struct.Emoji.t()]}, WSState.t()}
   @type guild_integrations_update ::
-          {:GUILD_INTEGERATIONS_UPDATE, GuildIntegrationsUpdate.t(), WSState.t()}
+          {:GUILD_INTEGRATIONS_UPDATE, GuildIntegrationsUpdate.t(), WSState.t()}
   @type guild_member_add ::
           {:GUILD_MEMBER_ADD,
            {guild_id :: integer, new_member :: Nostrum.Struct.Guild.Member.t()}, WSState.t()}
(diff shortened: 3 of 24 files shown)


lib/nostrum/shard/event.ex
```diff
@@ -3,13 +3,10 @@ defmodule Nostrum.Shard.Event do
 
   alias Nostrum.Shard.Payload
   alias Nostrum.Shard.Stage.Producer
-  alias Nostrum.Util
 
   require Logger
 
   def handle(:dispatch, payload, state) do
-    payload = Util.safe_atom_map(payload)
```
Kraigie, line 11: IIRC this will definitely at the very least cause breaking changes.

Anyone who attempts to access a key by via the atom syntax (`guild.id` for example) will be at risk of having issues appear. When I was implementing this functionality I was looking at the guild response I think and in that struct alone there were differing keys.

I think if we wanted to do this piecemeal we'd need to go into the event handlers (on our side) and change them at a high level there to convert to atoms, but add a header for this problematic payload for now that does not convert them. Thoughts?

Kraigie: I took another look and I see what you're talking about now with this probably not being an issue because of how we build the structs. If we do offer any events not as a struct though it would still be an issue, and I believe we do this in a lot of places.

BrandtHill (commented):
Voice gateway uses JSON only so this might break things

lib/nostrum/shard/event.ex
jchristgit: Yes, there are quite a few, via `lib/nostrum/shard/dispatch.ex`. Should I convert those to structs?

Kraigie: I think ideally they should be yes. But I also know it's a lot of work. :(

jchristgit:
> Voice gateway uses JSON only so this might break things

Yes, I wasn't sure whether I should touch the voice parts in the first place, but I did it anyways because I like causing some terror. Do we want to transform those into structs?

lib/nostrum/shard/event.ex
jchristgit: Yes, it is. "Good things take time", or something like that.

We could also `safe_atom_map` those instead. But I vouch for the clean solution.

BrandtHill:
> > Voice gateway uses JSON only so this might break things
> 
> Yes, I wasn't sure whether I should touch the voice parts in the first place, but I did it anyways because I like causing some terror. Do we want to transform those into structs?

I have a pull request #255  which keeps voice gateway payloads as JSON maps with string keys and all access of those payloads are using string accessors. So once that's merged the Voice Session module should be wholly taken care of with respect to these changes.

lib/nostrum/shard/event.ex
jchristgit: This should be taken care of by https://github.com/Kraigie/nostrum/commit/1674954530b7ef96428aa0484d1ec99d9ade714c.

jb3 (approved):

Kraigie (approved):
This is PEAK elixir code. 10/10. Would review again. Actually - I think I will. Every night before I go to bed and every morning when I wake up. I'm a better person having reviewed this. 

:shipit:

jchristgit:
thank you craig

jchristgit:
Thanks!

PR #255, opened 2021-06-26
Use json maps in voice session instead of casting to atom maps.

BrandtHill:


jchristgit (approved):
Stellar. Thank you very much.

Kraigie (approved):
LGTM, I think eventually we'd want structs to encapsulate the data, but this is fine for now since users never really see this,

jchristgit:
Thanks!

PR #260, opened 2021-07-17
Allow user to override the user caching method.

jchristgit:
By default, we will still resort to ETS. However, users can overwrite
the cache implementation with their own, to cache in other datastores,
for example `:mnesia` or Redis.


jb3 (approved):

jchristgit:
@jb3 I've been testing this on my live bot for quite some time and apart from the occasional shard session crash i think it is entirely stable. anything else would be a felony because this is Erlang and just let it crash is the way professionals handle their errors here please just test it on your own and you will observe that it is in fact fully functional.

david-cooke:
> @jb3 I've been testing this on my live bot for quite some time and apart from the occasional shard session crash i think it is entirely stable. anything else would be a felony because this is Erlang and just let it crash is the way professionals handle their errors here please just test it on your own and you will observe that it is in fact fully functional.

I thought bolt never crashed

jchristgit:
Thanks for the review 🙂

PR #261, opened 2021-07-17
Reword some parts of the README, link documentation.

jchristgit:
Take it from the third language English guy.


Kraigie (approved):
LGTM. Stellar work. Wondering if we should change the `666` to be less satanic, but 🤷

jchristgit:
Craig, thanks! I think satanic is exactly what we want to go for here.

I rebased from master. Merging since it was only a conflict.

PR #262, opened 2021-07-17
Do not document `num_shards` setting.

jchristgit:
Attempt to decomplicate a bit. Nostrum is built on Erlang, and Erlang is
smart, so we do not need to document this configuration value, as it is
not needed. Save users some time.

### DIFF
--- a/docs/static/Intro.md
+++ b/docs/static/Intro.md
@@ -73,7 +73,7 @@ config :nostrum,
 
 The following fields are also supported:
 
- - `num_shards` - A fixed number of shards to run, or `:auto` to have Nostrum determine it automatically.
+ - `num_shards` - A fixed number of shards to run, or `:auto` to have Nostrum determine it automatically. Defaults to `:auto`.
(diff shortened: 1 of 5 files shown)


docs/static/Intro.md
```diff
@@ -68,15 +68,12 @@ The file should be located at `/config/config.exs`. To run Nostrum you need the
...
-  num_shards: 2 # The number of shards you want to run your bot under, or :auto.
+  token: 666  # The token of your bot as a string
 ```
-If you don't know what `num_shards` is or don't have your bot on a lot of guilds
-you can omit the field and it will default to 1. You can also set this option to
-`:auto` and Nostrum will automatically get the recommended number of shards.
 
 The following fields are also supported:
 
+ - `num_shards` - A fixed number of shards to run, or `:auto` to have Nostrum determine it automatically.
```
jb3, line 76:
```suggestion
 - `num_shards` - A fixed number of shards to run, or `:auto` to have Nostrum determine it automatically. Defaults to `:auto`.
```

jb3 (changes requested):
I like this change. We should still document the default though (as we do with other options listed under the additional fields).

jchristgit:
Thanks!

PR #267, opened 2021-07-18
Use `:gun` as single HTTP client.

jchristgit:


jchristgit:
I need some help with this. This is fully functional, **except** for the last one missing piece, which is multipart form requests for messages with files. I do not know how to make it work. Me and @jb3 have tried tons of things to make it work, but we can't figure out how to send a multipart request...

BrandtHill (commented):
Just a thought, perhaps `method` can converted to an uppercase string here, then in the API module the `request` calls can remain atoms like `:get`, `:post`, etc.

jchristgit:
> Just a thought, perhaps method can be converted to an uppercase string here, then in the API module the request calls can remain atoms like :get, :post, etc.

Yesss, that's a good idea, although I've already very work-intensively (search and replace) replaced all the calls here, ahh!

I think we should reconsider doing that once we got multipart requests working for sure. Sounds like a good idea.

lib/nostrum/api/ratelimiter.ex
```diff
@@ -33,7 +34,10 @@ defmodule Nostrum.Api.Ratelimiter do
 
   def init([]) do
     :ets.new(:ratelimit_buckets, [:set, :public, :named_table])
-    {:ok, []}
+    domain = to_charlist(Constants.domain())
+    {:ok, conn_pid} = :gun.open(domain, 443)
```
BrandtHill, line 38: Remember the rules of `:gun` safety. The option for `retry` must be set to `1_000_000_000` (`:infinity` doesn’t work) for `:gun` to stay up. It defaults to `5`.

jchristgit: You are absolutely right. 72ef871 makes gun invulnerable.

jchristgit:
Via Discord:

Here's what a multipart reqest looks like from HTTPoison/hackney

```elixir
[<<"-----------------------------nwzbqxuomekhzpow\r\ncontent-length: 222618\r\ncontent-type: image/png\r\ncontent-disposition: form-data; name=\"file\"; filename=\"falling_joe.png\"\r\n\r\n">>,
{:file,<<"/mnt/c/Users/Brandt/Pictures/falling_joe.png">>},
<<"\r\n">>,
<<"-----------------------------nwzbqxuomekhzpow\r\ncontent-length: 2\r\ncontent-type: application/octet-stream\r\ncontent-disposition: form-data; name=\"payload_json\"\r\n\r\n{}\r\n">>,
<<"-----------------------------nwzbqxuomekhzpow--\r\n">>]
```

```elixir
-----------------------------nwzbqxuomekhzpow
content-length: 222618
content-type: image/png
content-disposition: form-data; name="file"; filename="falling_joe.png"

(File content goes here)
-----------------------------nwzbqxuomekhzpow
content-length: 2
content-type: application/octet-stream
content-disposition: form-data; name="payload_json"

{}
-----------------------------nwzbqxuomekhzpow--
```

jchristgit:
I found one more bug while testing this:

```elixir
iex(2)> Nostrum.Api.get_channel_messages(455589923626549259, 5, {:before, 455589479713865749})
** (exit) exited in: GenServer.call(Ratelimiter, {:queue, %{body: "", headers: [{"content-type", "application/json"}], method: "GET", options: [params: [limit: 5, before: 455589479713865749]], route: "/channels/455589923626549259/messages"}, nil}, :infinity)
    ** (EXIT) an exception was raised:
        ** (BadMapError) expected a map, got: [params: [limit: 5, before: 455589479713865749]]
            (stdlib 3.15.1) maps.erl:517: :maps.get(:flow, [params: [limit: 5, before: 455589479713865749]], :infinity)
            (gun 2.0.0-rc.2) ~/workspace/bolt/deps/gun/src/gun.erl:636: :gun.request/6
            (nostrum 0.4.7) lib/nostrum/api/base.ex:11: Nostrum.Api.Base.request/6
            (nostrum 0.4.7) lib/nostrum/api/ratelimiter.ex:75: Nostrum.Api.Ratelimiter.do_request/2
            (nostrum 0.4.7) lib/nostrum/api/ratelimiter.ex:59: Nostrum.Api.Ratelimiter.handle_call/3
            (stdlib 3.15.1) gen_server.erl:721: :gen_server.try_handle_call/4 
            (stdlib 3.15.1) gen_server.erl:750: :gen_server.handle_msg/6
            (stdlib 3.15.1) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
    (elixir 1.12.2) lib/gen_server.ex:1024: GenServer.call/3
    (nostrum 0.4.7) lib/nostrum/api.ex:810: Nostrum.Api.get_channel_messages_call/3
    (nostrum 0.4.7) lib/nostrum/api.ex:774: Nostrum.Api.get_messages_sync/4
```

I think I'm passing around the options wrongly here.

jchristgit:
Okay, that bug seems to apply to everything that wants to pass HTTP query parameters, from what I understand.

The calls found by this vim regex may be problematic:

```vim
?request(".*:  
```

jchristgit:
Okay, so the current way we are passing `options` to `:gun.request` in `lib/nostrum/api/base.ex` is wrong, since those have type `:gun.req_opts()` and represent internal options. In that case, I will need to gobble out the `params` and URI-encode it, if it's present.

jchristgit:
Okay, here's something odd.

```elixir
13:25:05.537 [error] unknown error, original message: %Nostrum.Struct.Message{...}, error: {:error,
 {:stream_error,
  {:stream_error, :protocol_error,
   :"Content-length header received in a 204 response. (RFC7230 3.3.2)"}}}
```
This is _probably_ the response from `bulk_delete_messages`. I don't have a traceback :(

I guess Discord responds with a 204 when pruning here?

BrandtHill:
> I guess Discord responds with a 204 when pruning here?

Yup that’s what it is:

> Returns a 204 empty response on success. Fires a Message Delete Bulk Gateway event.

looks like a lot of stuff returns a 204

jchristgit:
So what are our options - from what I understand:

- not use gun
- ask gun to not validate this
- ask Discord to fix this

BrandtHill:
I am pro-`:gun`, I think Discord can be asked to fix their thing and `:gun` should have an option to ignore those error types. In the meantime a shim would be to have a `handle_info` that matches the error you see and if it’s this issue treat it as a normal response.

jchristgit:
> I am pro-:gun, I think Discord can be asked to fix their thing and :gun should have an option to ignore those error types. In the meantime a shim would be to have a handle_info that matches the error you see and if it’s this issue treat it as a normal response.

Agreed with the shim. The problem is that in the error case, we do not get any headers back, and the ratelimiter expects them to be there to determine how often we can continue sending requests. For reference, my simple but functional patch:
```diff
diff --git a/lib/nostrum/api/base.ex b/lib/nostrum/api/base.ex
index e8507d5..beec284 100644
--- a/lib/nostrum/api/base.ex
+++ b/lib/nostrum/api/base.ex
@@ -27,6 +27,11 @@ defmodule Nostrum.Api.Base do
         {:ok, body} = :gun.await_body(conn, stream)
         {:ok, {status, headers, body}}
 
+      # Make up a response. We don't control the Discord web servers, after all.
+      # https://github.com/Kraigie/nostrum/pull/267#issuecomment-903253950
+      {:error, {:stream_error, {:stream_error, :protocol_error, :"Content-length header received in a 204 response. (RFC7230 3.3.2)"}}} ->
+        {:ok, {204, [], ""}}
+
       {:error, _reason} = result ->
         result
     end
```

This results in a horrible ratelimiter crash:
``elixir
19:59:59.490 [error] GenServer Ratelimiter terminating
** (FunctionClauseError) no function clause matching in String.to_charlist/1
    (elixir 1.12.2) lib/string.ex:2335: String.to_charlist(nil)
    (nostrum 0.4.7) lib/nostrum/api/ratelimiter.ex:129: Nostrum.Api.Ratelimiter.date_string_to_unix/1
    (nostrum 0.4.7) lib/nostrum/api/ratelimiter.ex:100: Nostrum.Api.Ratelimiter.handle_headers/2
    (nostrum 0.4.7) lib/nostrum/api/ratelimiter.ex:84: Nostrum.Api.Ratelimiter.do_request/2
    (nostrum 0.4.7) lib/nostrum/api/ratelimiter.ex:59: Nostrum.Api.Ratelimiter.handle_call/3
```

I'm not sure how to deal with this properly. We need the ratelimiting information from Discord, but we only get it from the HTTP library _on success_...

bdanklin:
I have published forks of both libraries to hex which allow this to move forward with minimal friction.

```elixir
defp deps do
  [
    {:cowlib, "~> 2.11.1", hex: :remedy_cowlib, override: true},
    {:gun, "~> 2.0", hex: :remedy_gun},
  ]
end

jchristgit:
@bdanklin rock solid work. Thank you very much. With the patched libraries, this works without a problem.

Should we report this to the upstream cowlib repository?

PR #272, opened 2021-07-18
Parse interaction types properly and document types.

### LINKED ISSUE #235: Parse Interactions Options to Integers
Values 6, 7, and 8 of [`ApplicationCommandOptionType`](https://discord.com/developers/docs/interactions/slash-commands#applicationcommandoptiontype) represent integer IDs and should be parsed to integers instead of being left as strings. This would cause it to match the other ID fields in the `Interaction` struct, and make working with interactions easier.

jchristgit:
Discord sends other types than string here occasionally. That said, when
a snowflake arrives, we want to decode it from string to integer, which
makes looking it up in the `resolved` type easier. Document the new
types for the user.

Fixes #235.


jb3:
Thanks!

PR #275, opened 2021-07-18
Define `Nostrum.Struct.Guild.Integration`.

jchristgit:
This was previously only a stub, and caused dialyzer to throw up.


Kraigie (approved):
Stellar!

jchristgit:
Thanks!

PR #277, opened 2021-07-18
Ignore dialyzer warnings for functions that use Porcelain.

BrandtHill:
Porcelain process spawning functions can return an error tuple even though it's not specified in the type specs. Until [this PR](https://github.com/alco/porcelain/pull/54) is merged and Porcelain has another release, these added attributes will allow dialyzer to ignore the false positives.


jchristgit (approved):
In Joe's words, excellent.

jchristgit:
Thanks!

PR #278, opened 2021-07-28
Maintain status of self-deaf and self-mute for bot. 

BrandtHill:
Keep self_deaf and self_mute for reconnects.
Fix to_struct for voice event structs.

### DIFF
--- a/lib/nostrum/struct/event/voice_state.ex
+++ b/lib/nostrum/struct/event/voice_state.ex
@@ -92,19 +92,19 @@ defmodule Nostrum.Struct.Event.VoiceState do
   @doc false
   def to_struct(map) do
     %__MODULE__{
-      guild_id: map["guild_id"],
-      channel_id: map["channel_id"],
-      user_id: map["user_id"],
+      guild_id: map.guild_id,
+      channel_id: map.channel_id,
+      user_id: map.user_id,
       member: Util.cast(map["member"], {:struct, Member}),
-      session_id: map["session_id"],
-      deaf?: map["deaf"],
-      mute?: map["mute"],
-      self_deaf?: map["self_deaf"],
-      self_mute?: map["self_mute"],
-      self_stream?: map["self_stream"],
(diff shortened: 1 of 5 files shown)


lib/nostrum/struct/event/voice_state.ex
```diff
@@ -92,19 +92,19 @@ defmodule Nostrum.Struct.Event.VoiceState do
...
-      self_stream?: map["self_stream"],
-      self_video?: map["self_video"],
-      suppress?: map["suppress"],
-      request_to_speak_timestamp: convert_stamp(map["request_to_speak_timestamp"])
+      session_id: map.session_id,
+      deaf?: map.deaf,
+      mute?: map.mute,
+      self_deaf?: map.self_deaf,
+      self_mute?: map.self_mute,
+      self_stream?: map[:self_stream] || false,
```
jchristgit, line 104: [I'm not even going to question why this is optional but the other ones isn't (on Discord's side).....](https://discord.com/developers/docs/resources/voice#voice-state-object)

lib/nostrum/struct/event/voice_state.ex
```diff
@@ -92,19 +92,19 @@ defmodule Nostrum.Struct.Event.VoiceState do
   @doc false
   def to_struct(map) do
```
jchristgit, line 93: Thanks so much for testing the dev version and fixing this. It seems my recent changes here were quite the mess 🍅

jchristgit (approved):
Rock solid. Thank you so much!

lib/nostrum/struct/event/voice_state.ex
BrandtHill: Yeah it's also super very cool that all the keys to this struct are atoms except for `"member"` which is a string for no reason 🤷‍♂️

jchristgit: Consistency. You love to see it...

PR #279, opened 2021-07-28
Correct documentation as the voice_speaking_update returns a different WS struct type.

BrandtHill:


lib/nostrum/struct/voice_ws_state.ex
```diff
@@ -16,4 +18,65 @@ defmodule Nostrum.Struct.VoiceWSState do
     :heartbeat_interval,
     :heartbeat_ref
   ]
+
+  @typedoc "The guild id"
```
jchristgit, line 22: I think it would be nice if we could write this out like this:

```suggestion
  @typedoc "The guild id which this event applies to"
```

What do you think?

lib/nostrum/struct/voice_ws_state.ex
```diff
@@ -16,4 +18,65 @@ defmodule Nostrum.Struct.VoiceWSState do
...
+  @typedoc "PID of the connection process"
+  @type conn_pid :: pid
+
+  @typedoc "Stream reference for `:gun`"
+  @type stream :: reference()
+
+  @typedoc "Gateway URL"
+  @type gateway :: String.t()
+
+  @typedoc "Where the session has been identified"
```
jchristgit, line 43: I believe there's a typo here:
```suggestion
  @typedoc "Whether the session has been identified"
```

jchristgit (changes requested):
Thank you very much for this.

I have two minor nags here, otherwise this looks excellent!

BrandtHill:
I've fixed those things in the types 👍

jchristgit (approved):
Thank you, this looks rock solid!

PR #280, opened 2021-07-28
Voice listen

BrandtHill:
Add RTP listening to Voice API

### DIFF
--- a/lib/nostrum/voice.ex
+++ b/lib/nostrum/voice.ex
@@ -485,6 +485,7 @@ defmodule Nostrum.Voice do
 
   This function will block until the specified number of packets is received.
   """
+  @doc since: "0.5.0"
   @spec listen(Guild.id(), pos_integer) :: [{binary, binary}] | {:error, String.t()}
   def listen(guild_id, num_packets) do
     voice = get_voice(guild_id)
--- a/lib/nostrum/voice.ex
+++ b/lib/nostrum/voice.ex
@@ -474,6 +474,28 @@ defmodule Nostrum.Voice do
     end
   end
 
+  @doc """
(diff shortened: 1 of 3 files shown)


lib/nostrum/voice.ex
```diff
@@ -474,6 +474,29 @@ defmodule Nostrum.Voice do
...
+  """
+  @doc since: "0.5.0"
+  @spec listen(Guild.id(), pos_integer) :: [{binary, binary}] | {:error, String.t()}
+  def listen(guild_id, num_packets) do
+    voice = get_voice(guild_id)
+
+    if VoiceState.ready_for_rtp?(voice) do
+      Enum.map(1..num_packets, fn _ -> Audio.get_rtp_packet(voice) end)
+    else
+      {:error, "Must be connected to voice channel to listen for incoming data."}
```
jchristgit, line 496: Since we're in a library here, do you think it would make more sense to return an atom as the error, such as

```suggestion
      {:error, :not_connected}
```

I think it would make it easier for users to match on the error, should this function start returning other errors.

jchristgit (approved):
This looks really really cool, thank you for it! One small comment / idea and we're good to go.

lib/nostrum/voice.ex
BrandtHill: I kept it a string because the rest of the voice API returns a descriptive error message and wanted to conform to that without breaking the rest. Maybe the rest should be changed but could be breaking

jchristgit: Perfectly reasonable. Thanks!

jchristgit:
Rock solid.

jchristgit:
Thanks!

PR #283, opened 2021-08-03
Fix gateway payload key types.

jchristgit:
Thanks to The Major for reporting these.


jchristgit:
(I will merge this on my own, since it's a regression from the cache PR, and master is kind of broken without it at the moment)

PR #284, opened 2021-08-03
Trim voice state updates with correct key(s)

BrandtHill:
Small fix to trim the heavy voice state update which was lost during the move to ETS guild cache. Dropping both `"member"` and `:member` keys from the map in case Discord later decides to normalize and return `member` as an atom instead of a string.


jchristgit (approved):
Good catch, thank you!

PR #285, opened 2021-08-03
Use proper key type for channel update events.

jchristgit:
Thanks to snapcase for reporting this issue.


jchristgit:
(I will merge this on my own, since it's a regression from the cache PR, and master is kind of broken without it at the moment)

PR #286, opened 2021-08-04
Fix `MessageDelete.to_struct` for direct messages

RealVidy:
`:guild_id` is not set when a direct message is deleted. This change ensures that we don't crash when that happens.


jchristgit (approved):
Thank you!

PR #289, opened 2021-08-05
Gun url params

BrandtHill:
Create query string from params if present in options and convert options to map if it's a keyword list.

### DIFF
--- a/lib/nostrum/api/base.ex
+++ b/lib/nostrum/api/base.ex
@@ -5,14 +5,26 @@ defmodule Nostrum.Api.Base do
 
   import Nostrum.Constants, only: [base_route: 0]
 
+  def request(conn, method, route, body, raw_headers, options) when is_list(options) do
+    request(conn, method, route, body, raw_headers, Map.new(options))
+  end
+
   def request(conn, method, route, body, raw_headers, options) do
-    full_route = "#{base_route()}#{route}"
     headers = process_request_headers(raw_headers)
     # Convert method from atom to string for `:gun`
     method =
       method
       |> Atom.to_string()
       |> String.upcase()
+
+    params = Map.get(options, :params, [])
+
+    query_string =
+      if params == [],
+        do: "",
+        else: "?" <> (Enum.map(params, fn {k, v} -> "#{k}=#{v}" end) |> Enum.join("&"))
+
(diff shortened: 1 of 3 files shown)


lib/nostrum/api/base.ex
```diff
@@ -5,15 +5,27 @@ defmodule Nostrum.Api.Base do
...
       method
       |> Atom.to_string()
       |> String.upcase()
 
+    params = Map.get(options, :params, [])
+
+    query_string =
+      if params == [],
+        do: "",
+        else: "?" <> (Enum.map(params, fn {k, v} -> "#{k}=#{v}" end) |> Enum.join("&"))
```
jchristgit, line 25: I think we can use [`URI.encode_query/1`](https://hexdocs.pm/elixir/1.12/URI.html#encode_query/2) here instead, right? That way this would also work with arguments with spaces in it.

lib/nostrum/api/base.ex
```diff
@@ -5,15 +5,27 @@ defmodule Nostrum.Api.Base do
 
   import Nostrum.Constants, only: [base_route: 0]
 
+  def request(conn, method, route, body, raw_headers, options) when is_list(options) do
```
jchristgit, line 8: A question. Should we search-and-replace the current `options: []` with `options: %{}` to prevent building garbage here?

jchristgit (changes requested):
Thanks!

This is ROCK SOLID work, @BrandtHill. Thanks so much. Some days I think I may be going insane, but this PR really proves the opposite. This is nice Elixir code. I have two minor nitpicks, or rather, questions. Do not rush.

lib/nostrum/api/base.ex
BrandtHill: Ah, I knew there had to be a function for this.

lib/nostrum/api/base.ex
BrandtHill: I see that the `options` that gets passed to `HTTPoison` includes `params`, and that's the only key that Nostrum for options. So I'll change `options` to `params` wherever necessary and ignore the `:gun` specific `options`.

jchristgit (approved):
Rock solid. Thanks!

jchristgit:
Thanks!

Issue #290, opened 2021-08-06
Key Error on GUILD_UPDATE


Th3-M4jor:
This is another case of Discord being inconsistent in using strings or atoms as keys when using ETF payloads.

If this happens more than once in a quick enough succession, the cache process is not restarted and events stop being sent to the user's Consumer process.

This currently happens on the latest commit to `master`.

```
[error] GenServer Nostrum.Shard.Stage.Cache terminating
** (KeyError) key :id not found in: %{"id" => [guild_id], "verification_level" => 1, "preferred_locale" => "en-US", "afk_channel_id" => nil, "vanity_url_code" => nil, [ . . . ]}
```

Stacktrace:
```
(nostrum 0.4.7) lib/nostrum/cache/guild_cache.ex:204: Nostrum.Cache.GuildCache.update/1
(nostrum 0.4.7) lib/nostrum/shard/dispatch.ex:151: Nostrum.Shard.Dispatch.handle_event/3
(nostrum 0.4.7) lib/nostrum/shard/dispatch.ex:42: Nostrum.Shard.Dispatch.handle/1
(elixir 1.12.0) lib/enum.ex:1553: Enum."-map/2-lists^map/1-0-"/2
(nostrum 0.4.7) lib/nostrum/shard/stage/cache.ex:23: Nostrum.Shard.Stage.Cache.handle_events/3
(gen_stage 1.1.0) lib/gen_stage.ex:2462: GenStage.consumer_dispatch/6
(gen_stage 1.1.0) lib/gen_stage.ex:2641: GenStage.take_pc_events/3
(stdlib 3.15.2) gen_server.erl:695: :gen_server.try_dispatch/4
```

jchristgit:
99 little inconsistencies in the ETF gateway,
99 little inconsistencies in the ETF gateway,
Take one down, patch it around
117 little inconsistencies in the ETF gateway

PR #291, opened 2021-08-06
fix: `Dispatch` expected `Channel` struct but it's map with string keys

RealVidy:
In `Dispatch`'s `handle_event` functions, the payloads are maps with string keys, not atoms as the code expected.

The code doesn't seem to have changed in 4 years so I am not sure what is going on. Am I missing something or did we make a change recently that broke this?

Example payloads:

```elixir
:CHANNEL_DELETE
%{
  "bitrate" => 64000,
  "guild_hashes" => %{
    channels: %{hash: "FUnLGUp92zY"},
    metadata: %{hash: "e3mMLdqJvmU"},
    roles: %{hash: "RqryBXuGhAQ"},
    version: 1
  },
  "guild_id" => 808702654830084096,
  "id" => 873254279505399808,
  "name" => "Haha",
  "nsfw" => false,
  "parent_id" => 846862417451548682,
  "permission_overwrites" => [
    %{
      "allow" => 1024,
      "allow_new" => "1024",
      "deny" => 0,
      "deny_new" => "0",
      "id" => 668940363196792849,
      "type" => :member
    },
    %{
      "allow" => 0,
      "allow_new" => "0",
      "deny" => 1024,
      "deny_new" => "1024",
      "id" => 808702654830084096,
      "type" => :role
    }
  ],
  "position" => 9,
  "rtc_region" => nil,
  "type" => 2,
  "user_limit" => 0
}
```

### DIFF
--- a/lib/nostrum/cache/guild_cache.ex
+++ b/lib/nostrum/cache/guild_cache.ex
@@ -221,7 +221,7 @@ defmodule Nostrum.Cache.GuildCache do
   def channel_create(guild_id, channel) do
     [{_id, guild}] = :ets.lookup(@table_name, guild_id)
     new_channel = Util.cast(channel, {:struct, Channel})
-    new_channels = Map.put(guild.channels, channel.id, new_channel)
+    new_channels = Map.put(guild.channels, channel["id"], new_channel)
     new_guild = %{guild | channels: new_channels}
     true = :ets.update_element(@table_name, guild_id, {2, new_guild})
     new_channel
--- a/lib/nostrum/shard/dispatch.ex
+++ b/lib/nostrum/shard/dispatch.ex
@@ -59,13 +59,13 @@ defmodule Nostrum.Shard.Dispatch do
     end
   end
 
-  def handle_event(:CHANNEL_CREATE = event, %{type: 1} = p, state) do
+  def handle_event(:CHANNEL_CREATE = event, %{"type" => 1} = p, state) do
     {event, ChannelCache.create(p), state}
   end
 
-  def handle_event(:CHANNEL_CREATE = event, %{type: t} = p, state) when t in [0, 2] do
-    :ets.insert(:channel_guild_map, {p.id, p.guild_id})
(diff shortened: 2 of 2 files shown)


lib/nostrum/cache/guild_cache.ex
```diff
@@ -221,7 +221,7 @@ defmodule Nostrum.Cache.GuildCache do
   def channel_create(guild_id, channel) do
     [{_id, guild}] = :ets.lookup(@table_name, guild_id)
     new_channel = Util.cast(channel, {:struct, Channel})
-    new_channels = Map.put(guild.channels, channel.id, new_channel)
+    new_channels = Map.put(guild.channels, channel["id"], new_channel)
```
RealVidy, line 224: Copied from line 244

Th3-M4jor:
What's happening is that Discord provides no guarantee on keys being strings or atoms when receiving events as EETF payloads via the websocket.
And they will not standardize it because they consider it too much work and computation to do so.

See #290, #285, #283, and discord/discord-api-docs#296

RealVidy:
Thanks for clarifying, it makes more sense now :)

RealVidy:
Won't discord always send string keys on the websocket though since I am guessing their endpoint is language agnostic?
Or do they have a specific endpoint for erlang / elixir where they send binary erlang data?

Th3-M4jor:
It's that when creating the websocket connection you have a choice of receiving payloads as JSON or as EETF

jchristgit:
This looks good from a code perspective, but you also deleted the other function heads for the other format. In my testing, it seemed like we sometimes get this in one format, sometimes in the other. Was it different from you?

jchristgit:
@RealVidy Thank you so much for submitting this!

After some back and forth in the maintainer team, and given the sheer amount of issues that the key normalization change has made, we have decided to roll back the offending changes in #300. This should solve all the issues you were having, and fortunately results in us no longer having to work around all the string & atom key messes. If you still encounter issues, feel free to open an issue!

PR #294, opened 2021-08-08
docs: correct spelling of allowed_mentions atoms

Ovyerus:
Changes the spelling of `:user` and `:role` in the documentation to be pluralised correctly to match how it actually is in the code. Accidentally pinged everyone thanks to this, good thing it was in a testing guild 😅

Relevant lines
https://github.com/Kraigie/nostrum/blob/517894fd9d256210c3acdf5698b2dcc3be5b4990/lib/nostrum/api.ex#L3352-L3358

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -178,10 +178,10 @@ defmodule Nostrum.Api do
     * `:all` (default) - Ping everything as usual
     * `:none` - Nobody will be pinged
     * `:everyone` - Allows to ping @here and @everone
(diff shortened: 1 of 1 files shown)


jerbob (approved):
LGTM!

david-cooke (approved):
LGTM!

jchristgit (approved):
Thanks!

This is really nice. Makes sense. I think we should maybe have some way of validating these so it crashes at call time instead of pinging everyone.

jchristgit:
Thanks!

PR #299, opened 2021-08-11
Fix key error on guild update.

jchristgit:
Discord sends us string keys here, including the `"id"` key which we use
for looking up the guild in the cache. To properly upsert this, we use
the proper key and cast the new guild to a guild struct before merging
it with the old guild.

This also fixes us not returning the old guild on update, which the
consumer wants.

Fixes #290.

PR #300, opened 2021-08-14
Revert to normalizing gateway keys to atoms.

### LINKED ISSUE #290: Key Error on GUILD_UPDATE

jchristgit:
By the past bug reports and various issues related to variation of key
types coming from the gateway, this has become more of a hassle than
not.

This commit goes back to the behaviour of normalizing all keys to atoms,
with the small exception that snowflakes which look like integers are
parsed as integers. This was based off Matt Nowack (@ihumanable)s work
in our Discord channel, with a small documentation addition. Thank you!

Struct fields getters are updated to `.key` or `[:key]` depending on
whether or not they are documented as optional in the Discord API
documentation, with the exception of the `PartialApplication.flags`
field, which I blandly assume to be present.

Fixes #290.
Supersedes #291.
Supersedes #299.


Th3-M4jor:
Unless you already account for it elsewhere, and/or I missed it, It's possible that they key passed to `maybe_to_atom` will already be an integer. The API docs say that for ETF payloads all snowflakes will be 64-bit integers.
https://discord.com/developers/docs/topics/gateway#etfjson

jchristgit:
Hmm, I'm not sure if we had that case then. In that case, was #252 a bug on Discord's side? Or wrong documentation? Was this all wrong? 🤨

Th3-M4jor:
I'm not sure if that was a bug on discord's end or not, but I myself have checked the payload received on interactions and when keys were snowflakes they were sent as integers.

So I guess we should handle both for now? But make string snowflake keys be a warning since that should be a bug on discord's end according to the API docs.

jchristgit:
>  I'm not sure if that was a bug on discord's end or not, but I myself have checked the payload received on interactions and when keys were snowflakes they were sent as integers.

Huh - I observed the opposite, keys sent as binaries, which is why that issue was opened in the first place. Perhaps another case of inconsistent key types?

> So I guess we should handle both for now? But make string snowflake keys be a warning since that should be a bug on discord's end according to the API docs.

I'm 50/50 on adding a warning. On one hand, it's wrong according to the documentation, on the other hand, for someone who's building a bot a log message like that is noise and not really useful, isi t?

jb3 (approved):
Lovely, lovely.

jchristgit:
Thank you, thank you, Joe. Very kind of you.

Kraigie (approved):
Everything here looks good to me! I don't have strong opinions on adding a warning like was discussed.

Can't wait to revert this again in a week! :shipit: 🦖

jchristgit:
Thank you, thank you, Craig. Very kind of you.

jchristgit:
Thanks!

PR #303, opened 2021-08-16
Return old guild on guild update.

jchristgit:
Regression from the ETS guild cache migration.


jchristgit:
Merging since regression.

jchristgit:
100px-Erlang_logo
Thanks!

PR #306, opened 2021-08-18
Components

bdanklin:
This should provide all of the base types for the new message components.

My goal was to separate the components enough to split apart their distinct functionality, while not repeating all of the types and common structures that they need; because at the end of the day, they are all still "components".

I was unable to get the function below to work inside the macro. 

```elixir
  def new(opts \\ []) do
    opts
    |> Enum.into(@defaults)
    |> to_component()
  end

  def to_component(map), do: Component.to_struct(map)
  ```
  At the moment I have placed it inside each of the functions but hopefully, someone can help me fix it so that `use Component` would include that too and it doesn't need to be repeated...
  
Also had some issues with the `@spec` in a few places.

Wanted to get some feedback and make sure this is going in the right direction before going further.


bdanklin:
Wrong branch

PR #307, opened 2021-08-18
Components

### LINKED ISSUE #292: Does this library support the new buttons?
title, https://support.discord.com/hc/en-us/articles/1500012250861-Bots-Buttons

bdanklin:
EDIT: 
- Rebased to Master
- Adds all components ( buttons, action rows, selects ).
- Validates components upon create and update
- Component framework for any new components, very easy to add.

---
old:

This should provide all of the base types for the new message components.

My goal was to separate the components enough to split apart their distinct functionality, while not repeating all of the types and common structures that they need; because at the end of the day, they are all still "components".

I was unable to get the function below to work inside the macro.
```elixir
  def new(opts \\ []) do
    opts
    |> Enum.into(@defaults)
    |> to_component()
  end

  def to_component(map), do: Component.to_struct(map)
```
At the moment I have placed it inside each of the functions but hopefully, someone can help me fix it so that `use Component` would include that too and it doesn't need to be repeated...

Also had some issues with the `@spec` in a few places.

Wanted to get some feedback and make sure this is going in the right direction before going further.

should resolve #292

### DIFF
--- a/lib/nostrum/struct/component.ex
+++ b/lib/nostrum/struct/component.ex
@@ -75,8 +75,8 @@ defmodule Nostrum.Struct.Component do
 
   defmacro __using__(_opts) do
     quote do
-      alias Nostrum.Struct.{Component, Emoji}
       alias Nostrum.Struct.Component.{ActionRow, Button, Option, SelectMenu}
+      alias Nostrum.Struct.{Component, Emoji}
       alias Nostrum.Util
       @before_compile Component
     end
@@ -105,7 +105,7 @@ defmodule Nostrum.Struct.Component do
         |> Nostrum.Struct.Component.to_struct()
       end
 
-      defp allowed_keys(), do: Map.keys(@defaults)
+      defp allowed_keys, do: Map.keys(@defaults)
 
       ## Destroy all structs and ensure nested map
       def flatten(map), do: :maps.map(&do_flatten/2, map)
--- a/lib/nostrum/struct/component.ex
+++ b/lib/nostrum/struct/component.ex
@@ -1,0 +1,265 @@
+defmodule Nostrum.Struct.Component do
+  @moduledoc """
+  Components are a framework for adding interactive elements to the messages your app or bot sends. They're accessible, customizable, and easy to use. There are several different types of components; this documentation will outline the basics of this new framework and each example.
+
+  > Components have been broken out into individual modules for easy distinction between them and to separate helper functions and individual type checking between component types - especially as more components are added by Discord.
+
+  Each of the components are provided all of the valid types through this module to avoid repetition and allow new components to be added quicker and easier.
+
+  ## Action Row
+  An Action Row is a non-interactive container component for other types of components. It has a `type: 1` and a sub-array of `components` of other types.
+
+  - You can have up to 5 Action Rows per message
+  - An Action Row cannot contain another Action Row
+  - An Action Row containing buttons cannot also contain a select menu
+
+  ## Buttons
+  Buttons are interactive components that render on messages. They have a `type: 2`, They can be clicked by users. Buttons in Nostrum are further separated into two types, detailed below. Only the [Interaction Button](#module-interaction-buttons-non-link-buttons) will fire a `Nostrum.Struct.Interaction` when pressed.
+
+  ![Discord Buttons](https://discord.com/assets/7bb017ce52cfd6575e21c058feb3883b.png)
+
+  - Buttons must exist inside an Action Row
+  - An Action Row can contain up to 5 buttons
+  - An Action Row containing buttons cannot also contain a select menu
+
+  For more information check out the [Discord API Button Styles](https://discord.com/developers/docs/interactions/message-components#button-object-button-styles) for more information.
+
+  ## Link Buttons
+
+  - Link buttons **do not** send an `interaction` to your app when clicked
+  - Link buttons **must** have a `url`, and **cannot** have a `custom_id`
+  - Link buttons will **always** use `style: 5`
+
+  #### Link `style: 5`
+  ![Secondary](https://user-images.githubusercontent.com/34633373/129678527-d8d1c988-33c5-46d6-874e-2cfb2d9ba7f4.png)
+
+  ## Interaction Buttons ( Non-link Buttons )
+
+  > Discord calls these buttons "Non-link Buttons"  due to the fact that they do not contain a url. However it would be more accurate to call them an "Interaction Button" as they **do** fire an interaction when clicked which is far more useful for your applications interactivity. As such they are referred to as "Interaction Button" throughout the rest of this module.
+
+  - Interaction buttons **must** have a `custom_id`, and **cannot** have a `url`
+  - Can have one of the below `:style` applied.
+
+  #### Primary `style: 1`
+  ![Primary](https://user-images.githubusercontent.com/34633373/129678518-74790396-efcc-45f0-9a84-eba3f4c66950.png)
+
+  #### Secondary `style: 2`
+  ![Secondary](https://user-images.githubusercontent.com/34633373/129678527-d8d1c988-33c5-46d6-874e-2cfb2d9ba7f4.png)
+
+  #### Success `style: 3`
+  ![Success](https://user-images.githubusercontent.com/34633373/129678543-5d72ba09-f042-49b6-b56a-9ae072011ee1.png)
+
+  #### Danger `style: 4`
+  ![Danger (1)](https://user-images.githubusercontent.com/34633373/129678473-7e4b045f-b4a3-4993-96a2-4916cb88161f.png)
+
+  ## 🐼 ~~Emoji Buttons~~
+
+  > Note: The discord documentation and marketing material in relation to buttons indicates that there are three kinds of buttons: 🐼 **Emoji Buttons**, **Link Buttons** & **Non-Link Buttons**. When in fact all buttons can contain an emoji. Because of this reason 🐼 **Emoji Buttons** are not included as a seperate type. Emojis will be instead handled by the two included ( superior ) button types.
+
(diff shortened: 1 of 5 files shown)


bdanklin:
Updated this PR to become part of the API-v8 branch.

bdanklin:
The `new`, `to_component`, and `to_struct` functions all work correctly now.

Adding new components as they are released should be as easy as setting the `@defaults` and types and a cheeky `use Component`

lib/nostrum/struct/component.ex
```diff
@@ -75,13 +75,27 @@ defmodule Nostrum.Struct.Component do
...
+        |> Enum.into(@defaults)
+        |> to_component()
+      end
+
+      def to_component(map), do: Component.to_struct(map)
     end
   end
 
+  @doc """
+  Each component type implements this callback to be provided with a keyword list which will produce a probably-valid component of that type.
```
jchristgit, line 93: This "probably-valid" does not quite spark trust to users of the library. Does it produce valid components, or does it not? When could it not produce valid components? If we can hand out the reasons for this not generating valid components, that would already be a good thing to do.

jchristgit:
This `new` function is very nice 👍

lib/nostrum/struct/component.ex
bdanklin: Good point.

I was alluding to the fact that a user of the library _could_ attempt to create a component of `type: 2, label: "hello", placeholder: "🧀"` these are all valid component fields and they are typed correctly. But it is not a valid button.

That is what I was referring to here:

> These would greatly benefit from some helper functions that I showed when I initially asked for comment. 

The way I imagined it could be user-proofed was to include functions such as `create_link_button(label, url, emoji)` to ensure that components produced are always valid in regards to the combination of fields as well the contents of the fields themselves.

jchristgit: Currently, there's only the predefined set of components by Discord though, right?

If not, then we should probably document that Nostrum does not validate component types defined outside of the library. That way, users know which parts to use and which not to use.

bdanklin: Yes. Currently just the button, action row, and select.

bdanklin:
Changed this back to master, because even though we aren't receiving components on messages we can still send them.

This should be good to go

lib/nostrum/struct/component.ex
```diff
@@ -0,0 +1,267 @@
...
+  A partial emoji to display on the object.
+
+  Valid for [Buttons](#module-buttons)
+  """
+  @type emoji :: Emoji.partial_button_emoji() | nil
+
+  @typedoc """
+  A url for link buttons.
+
+  Valid for: Buttons, `String.t()` representing	an
```
jchristgit, line 213: Representing an?

lib/nostrum/struct/component/option.ex
```diff
@@ -0,0 +1,39 @@
...
+          value: value
+        }
+
+  def to_struct(map) do
+    %__MODULE__{
+      label: map["label"],
+      value: map["value"],
+      description: map["description"],
+      emoji: Util.cast(map["emoji"], {:struct, Emoji}),
+      default: map["default"]
```
jchristgit, line 36: This should have atom keys, right? We went back to normalizing them.

jchristgit (changes requested):
Looking great, thank you!

I have two minor nitpicks and then this looks good to me!

bdanklin:
Fixed both points.

Ty for review.

queer:
Would it be reasonable to use atoms for the component `type`s and convert them into integers at the point of serialisation? It's a lot less obvious to see `type: 1` or `type: 2` when looking at code as compared to `type: :action_row` / `type: :button`

Th3-M4jor:
I'd say allow for both somehow, so that if the user wants to use integers still they can.

bdanklin:
You don't need to use either when creating components, the helper functions I included handle the types and validations for you. ie.
```elixir
action_row = 
  ActionRow.action_row()
  |> ActionRow.append_lazy(nick_button)
  |> ActionRow.append_lazy(cage_button)
  |> ActionRow.append_lazy(con_button)
  |> ActionRow.append_lazy(air_button)
  |> ActionRow.append_lazy(Button.create_link_button("Best Movie Ever Made", "https://www.imdb.com/title/tt0118880/?ref_=fn_al_tt_1"))

Nostrum.Api.create_message(channel, content: "Test Massage", components: [action_row])
```

image

Th3-M4jor:
Don't have to, but what if I wanted to create a button by doing something like this?
```elixir
%Nostrum.Struct.Component{
    type: :action_row,
    components: [
     %Nostrum.Struct.Component{
      type: :button,
      style: :green_button,
      emoji: emoji_var,
      label: button_name,
      custom_id: lower_name
    }
   ]
}
````

Right now, I have to use the integers and remember which one is which.

**Comment by bdanklin** [comment]
I mean what if i wanted to create a button using a mad catz analog steering wheel for the playstation 1? Sometimes the functionality you want just aint there.

If you want to provide extra functionality after the fact by all means do. But my opinion is that the included helpers already go further toward 'beginner friendly' than most other modules in this library.

If i submit further work on components it would probably be:

- try to implement collectable and enumerable protocols for rows and select menus to ease bulk handling of inner types.
- helpers such as `|> success()` or `|> danger()` to swap a buttons style. Similar to the toggle/enable/disable functions.

**Comment by jb3** [comment]
I'm +1 with @Th3-M4jor — maybe we expose a module with a set of constants for this?

Either we provide a helper to get that integer value or convert it (I think I prefer the former in terms of implementation).

What are your thoughts on exposing something like the following?

```elixir
iex(1)> defmodule Nostrum.Struct.Component.ButtonStyles do
...(1)>   @styles [primary: 1, secondary: 2, success: 3, danger: 4, link: 5]
...(1)>
...(1)>   def style(selected_style) do
...(1)>     @styles[selected_style]
...(1)>   end
...(1)> end
{:module, Nostrum.Struct.Component.ButtonStyles, ..., {:style, 1}}
iex(2)> Nostrum.Struct.Component.ButtonStyles.style(:primary)
1
iex(3)> Nostrum.Struct.Component.ButtonStyles.style(:secondary)
2
```

End users can alias something like the proposed codeblock above, instead of directly passing an atom to the struct generation the helper can be used.

We could also expose functions in this module like `success/0` which just return the integer value, but I thought the module constant and getter-function was nicer.

bdanklin:
I feel like I am missing something because I find it superfluous and unnecessary trampolining through layers of abstraction.

Having a button object that you can pass through functions maintains a constant and obvious level of abstraction as you are updating messages or "control panels" for the user interface or whatever you are using buttons for.

```elixir
def disable(%{type: 2} = button) do
  button
  |> update({:disabled, true})
end

def sucess(%{type: 2} = button) do
  button
  |> update({:style, 3})
end 
```

from the users point of view button operations are simplified to:

```elixir
 def handle_something(:user_confirmed, button) do
  {:noreply, 
    button
    |> success()
    |> disable()}
end
```

If you want to handle everything at a lower level and manually update your structs ( or create a new one with a map every time ) then sure, being able to use `{:style, :success}` is easier than `{:style, 3}`. But I also think it is outside the scope of this PR. If you would like to add it separately then go ahead.

Also worth noting that there are far more egregious cases of "having to look up the type" throughout the library, for example, the [11 types of channels](https://discord.com/developers/docs/resources/channel#channel-object-channel-types) or the [21 types of messages](https://discord.com/developers/docs/resources/channel#message-object-message-types) that developers currently need to look up the value for. I don't think this is the straw that broke the googlers back.

Kraigie:
@bdanklin Hey! Sorry, I've been pretty hands off with this so didn't have time to jump in until now. First I wanted to say this looks very well done from an implementation standpoint. Nice job! Second, I wanted to ask - any specific reason you closed this?

If you're still interested in working on this, here are my thoughts: Everything looks killer! As for the conversation above regarding helpers for integer -> atom types, I don't think they're necessary at this time. I reckon if someone wants to build the whole component themselves, then looking up what integer a type is isn't going to be the deal breaker. If people feel strongly about this we can add something later but I don't think it's within the scope of this PR.

Besides that, the only other oddity I can find is the private function names such as `appnd`. Could these be rewritten to not be private functions, and instead pattern match on the function head, like the other `append` functions? If you want to continue to work on this and want a more thorough review, please let me know!

bdanklin:
Hi @Kraigie, I closed it because my other 2 PR's were closed without any reason, combined with some of the pushback I received for my chosen implementation it seemed that perhaps my contributions were either not fitting in with the style or purpose of the library or otherwise unwelcome. I am happy to reopen it.

I agree the inner workings of `append` and other helpers could be cleaned up. I am pattern matching on the function head for the entry to the function to allow the user to pass in various formats and we can process it accordingly, but ultimately the inner working is the same (`appnd`). eg. if the user passes in the new button already wrapped in a list, as might be the case when combining two sets of buttons, or if they pass in a button on its own, as could happen with a new button they create per my example above. They will both be resolved to a valid component. 

I'm also anticipating new components potentially needing to have their own rules around row limitations, eg a "text-input" component may count as 2 buttons wide, and this seemed like a fair compromise of where to leave the code for now.

FlXlT:
@bdanklin , thank you very much for your contribution. I would really appreciate it if we were to have this feature in Nostrum. 

As far as I know components are quite an important new feature for Discord and I think it would be a great deal for Nostrum to have this feature implemented.

For now I will I'll check out your branch and create some magic with the features you have created. I'd love to see the PR be fine-tuned and merged! :-)

Th3-M4jor:
I'm gonna take this one over, I'll have a replacement PR open soon.

bdanklin:
@FlXlT Per the discussion in discord I was unable to push updates from my branch since my fork became unlinked.

Thanks @Th3-M4jor.

FlXlT:
Thank you nonetheless, and @Th3-M4jor ofcourse!

Issue #297, opened 2021-08-11
Stream playback support


HATRED-arh:
Hello!
I might be wrong, but it seems that Nostrum currently doesn't support stream playback. 
Are there plans for this in the future? Would be nice to play some 24/7 radios or something. 

At the moment if I try to pass stream link into Voice.play() I get `ERROR: requested format not available`
If I change youtube-dl arguments in **audio.ex** file to `["-f", "bestaudio/best"], ["-o", "-"],  ["-q"], [url]`, it will process stream link and play it for several seconds. 
I know it's not a fix, since it still writes to file and eventually timeouts.  
Is it possible to implement function like Voice.play_stream() or something alike?

HATRED-arh:
> Can you provide an example URL you're trying to play?

Sure: 
https://youtu.be/oHY1RpbKDDA
https://youtu.be/LN4r-K8ZP5Q

BrandtHill:
I'm unable to get `youtube-dl` to download a live stream with `-f best`. It will download a 256KB segment then hang. Do you have a way to continually download ongoing streams with `youtube-dl`?

HATRED-arh:
I did some research
You can use `youtube-dl` only to get source link and just pass it to ffmpeg
Example:
`youtube-dl -f bestaudio/best -g https://youtu.be/LN4r-K8ZP5Q`
And pass the url you just got to ffmpeg:
[...]

PR #308, opened 2021-08-19
Support playing livestream audio with streamlink

BrandtHill:
#297


jchristgit (approved):
Thanks!

This looks rock solid. I do not have voice set up locally, but based on your past quality contributions I will assume that this works.

PR #310, opened 2021-08-19
Permissions update per: discord/discord-api-docs#3672

bdanklin:
https://github.com/discord/discord-api-docs/pull/3672/commits/2ed2414441169930349aae215f1d5c8946160299

Slightly early, but there were several missing permissions potentially causing issues so this one may want to go through early. I don't think it would do any harm to push this to main?

The thread permissions are due to go into effect on 9/7.

### DIFF
--- a/lib/nostrum/permission.ex
+++ b/lib/nostrum/permission.ex
@@ -31,6 +31,10 @@ defmodule Nostrum.Permission do
   """
   @type bitset :: non_neg_integer
 
+  @typedoc """
+  This grouping is from an older version of the permission API and no longer applies
+  """
+  @typedoc deprecated: "See t.0"
   @type general_permission ::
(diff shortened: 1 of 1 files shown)


jchristgit:
Thank you for the PR! I don't understand English but based on the PR https://github.com/discord/discord-api-docs/pull/3672 description "this is still in flux" they are still unsure about it? I think we should hold off with merging then until their PR is merged, in case they change something on their side?

bdanklin:
No problem. Just figured its better to get it on the board earlier rather than later incase someone else picked it up to work on.

bdanklin:
Updated with accurate typespec & categories according to the updated table here: https://discord.com/developers/docs/topics/permissions#permissions-bitwise-permission-flags

PR #311, opened 2021-08-20
302 Incorrect `@spec` for `Api.execute_webhook`

### LINKED ISSUE #302: Incorrect `@spec` for `Api.execute_webhook`
I noticed that the map defined in the `@spec` for [execute webook](https://github.com/kraigie/nostrum/blob/master/lib/nostrum/api.ex#L2822) is written such that all fields are required, however this is not the case.

Only one of `content`, `embeds`, or `file` are required fields.

bdanklin:
no dialyzer warnings for this file anymore but potentially the ugliest @spec ever?

potentially fixes #302 but might make it worse.


Th3-M4jor:
This is almost correct, just needs to use a different syntax for optional and required fields.
```elixir
@type m1 :: %{
    required(:content) => String.t(),
    optional(:components) => [components.t()]
    ...
}
```
Otherwise all fields are assumed to be required

jchristgit (changes requested):
This is very cool, thank you!

I have one question: Do we need the `| nil` here? The fields are marked as `optional` don't need to be passed when they are optional anyways, and when they are passed, they should match `nonempty_list(Embed.t())`. Same for `file` and `content`.

bdanklin:
> This is very cool, thank you!
> 
> Do we need the `| nil` here? .

Probably don't **need** it. 

I figured this was the most optimistic case to cover potential situations that may come up. For example. If I have `%{content: "hello"}`  and update it through webhook to `content: nil` and `embeds: [some_embeds]`, 
my intentions are explicit that I want to remove the `content` and add an `embed`. Whereas just putting the `embed` field in the update is ambiguous about what happens in the `content` field.

I may be mistaken and am happy to update it to remove them if desired!

Ty for review.

jchristgit (approved):
Rock solid. Thank you very much.

jchristgit:
Thanks!

jchristgit:
Thanks!

PR #312, opened 2021-08-22
Voice ready event

BrandtHill:
Add new voice ready event and add documentation on voice related events.


jchristgit (approved):
Thanks!

This is a really good idea, I especially like the inclusion of the `try_play/4` function since that's a useful helper 🙂

jchristgit:
Thanks!

PR #313, opened 2021-08-22
Fix all typespecs that mix docs complains about

BrandtHill:


jchristgit (approved):
Thanks!

Yup yup yup. Makes sense to me. I suppose [one day or another](https://www.youtube.com/watch?v=4kg9LasvLFE) we will want to define the user connection struct, but that said, is that even a thing with bot accounts..?

jchristgit:
Thanks!

PR #314, opened 2021-08-23
important fix

BrandtHill:


jchristgit (approved):
Thanks!

I can sleep better knowing that Nostrum won't run into this bug anymore.

jchristgit:
Rock solid.

PR #315, opened 2021-08-29
Fix ModuleUndefined error in InteractionOption

Th3-M4jor:
The `to_struct/1` fn in `Nostrum.Struct.ApplicationCommandInteractionDataOption` was not passing the correct arguments to `Nostrum.Util.cast/2`

Leading to an `UndefinedFunction` error when a slash command's sub-command is called.


jchristgit (approved):
Thanks!

Rock solid.

jchristgit:
Thanks!

PR #321, opened 2021-09-03
make interaction fields optional

### LINKED ISSUE #320: `:application_id` missing from `message.interaction`
If a message is created as an interaction response, the application_id field is missing from the interaction, instead it is present within the message object itself.

This causes a `KeyError` whenever a message that is created in response to an interaction is received by the bot.

Th3-M4jor:
This makes most fields in an interaction possibly `nil` for when it's part of a message.

Should this honestly instead be replaced with a `PartialInteraction` struct?

fixes #320

### DIFF
--- a/lib/nostrum/struct/interaction.ex
+++ b/lib/nostrum/struct/interaction.ex
@@ -93,15 +93,15 @@ defmodule Nostrum.Struct.Interaction do
   def to_struct(map) do
     %__MODULE__{
       id: map.id,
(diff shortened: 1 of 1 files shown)


Kraigie (approved):
LGTM! What do you think the `PartialInteraction` implementation solves? Could you elaborate?

jchristgit (approved):
Rock solid, thank you!

PR #324, opened 2021-09-07
Channel Docs

bdanklin:


bdanklin:
I recreated this PR to replace the closed one per @Kraigie request.

Not sure why the other pr could not be reopened

lib/nostrum/struct/channel.ex
```diff
@@ -37,132 +134,281 @@ defmodule Nostrum.Struct.Channel do
...
+  @type join_timestamp :: String.t()
+
+  @typedoc """
+  User thread settings, currently only used for notifications.
+  """
+  @typedoc since: "0.5"
+  @type flags :: integer()
+
+  @typedoc """
+  Default duration for newly created threads.
```
Kraigie, line 363: So these are in minutes? I think the example gets us there but we should just mention it IMO!

lib/nostrum/struct/channel.ex
```diff
@@ -1,27 +1,124 @@
 defmodule Nostrum.Struct.Channel do
-  @moduledoc ~S"""
+  @moduledoc """
+  Struct representing a channel.
 
-  A `Nostrum.Struct.Channel` represents a Guild or DM channel.
+  There are currently eleven channel types within discord, ranging from direct messages between two users to server announcement channels pushing a message to thousands. This module also contains several helper functions for handling channels within your applcation.
```
Kraigie, line 5: I think this could be cleaned up a bit. Maybe change the top line to "Struct and helper functions for working with channels."? We could then remove the second line completely. If you really want to include the bit about the different channel types then we could move it elsewhere. Up to you!

Kraigie: I also think we could consider this a `context` module as it'd be called in Phoenix. Just something to think about 🤷

Kraigie (approved):
Excellent work! Just a few grammatical changes/clarifications I think I'd like to see. Are you interested in doing more work like this for other modules? This is really good stuff!

PR #328, opened 2021-09-19
 Fix `GUILD_UPDATE` making some fields `nil` 

Th3-M4jor:
This fixes a fatal bug where if a Guild is updated, a member or channel update will then cause a `BadMapError`.


jchristgit (approved):
Smart fix. Thanks!

jchristgit:
Thanks!

PR #332, opened 2021-10-06
Fix docs workflow

jb3:
- Change docs workflow to build with Elixir v1.12
- Change Intro.md to specify that Nostrum targets Elixir v1.10 and up


jchristgit (approved):
Winner.

PR #336, opened 2021-10-09
Rtp duplicates

### LINKED ISSUE #335: Duplicate packets not handled ?
Not sure if this is on purpose but Nostrum.Voice.listen/2 doesnt seem to follow the RFC for the RTP payload format for Opus which states : 

> The receiving side MUST be prepared to receive duplicate RTP packets.
> The receiver MUST provide at most one of those payloads to the Opus
> decoder for decoding, and it MUST discard the others.

Perhaps a Enum.uniq() is enough to avoid this issue ?
source: https://datatracker.ietf.org/doc/html/rfc7587#section-4.1

BrandtHill:
Closes #335

### DIFF
--- a/lib/nostrum/voice.ex
+++ b/lib/nostrum/voice.ex
@@ -489,7 +489,7 @@ defmodule Nostrum.Voice do
     - `guild_id` - ID of guild that the bot is listening to.
     - `num_packets` - Number of packets to wait for.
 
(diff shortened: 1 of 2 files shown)


jchristgit:
Thanks!

jchristgit:
Thanks!

PR #339, opened 2021-10-10
Prevent SSL verification warning on `:gun.open`.

jchristgit:
This introduces a new dependency to `:certifi`, and adds a new helper
function which returns SSL options for verifying the target hostname.
`:certifi` is introduced in favour of pointing to a system certificate
bundle to stay platform-agnostic.

Thanks to snapcase for reporting this issue.


jb3:
Thanks!

PR #342, opened 2021-10-10
Implement pluggable guild caching.

jchristgit:
Similar to the pluggable user cache, this allows developers to choose
their own caching mechanism for structures received from Discord.
However, the approach taken here also differs slightly from the user
cache implementation to make improvements aimed at completely decoupling
the chosen cache implementation from nostrum itself. The following base
functionality is provided:

- Cache behaviour and dispatching module `Nostrum.Cache.GuildCache` with
  accompanying documentation.
- Default cache behaviour implementation `Nostrum.Cache.GuildCache.ETS`,
  which is the current ETS-based guild cache implementation.

Two architectural changes are made here in contrast to the user cache:

1. Guild cache implementations are supervisors. This allows any
implementations to completely manage what they do on their own, in a
manner that is managed and supported by OTP. Our ETS-based cache uses
this to set up ETS tables, an Mnesia-backed cache may perform and wait
for table creation, a Redis-backed cache may wait for the backing
connection to Redis to be ready, a "discard" cache may skip doing any
work, any child constellations fitting the OTP mechanisms are supported.

2. `defdelegate` is used instead of manually defining the functions to
dispatch to the backing cache. This reduces the code we need to write on
our end heavily, and prevents duplication of type specifications and
documentation. This applies to functions related to the cache itself and
functions for the `Supervisor`, allowing us to keep the
`CacheSupervisor` simple and not duplicate getting the configured cache.

Both of these changes will be integrated into the current user cache to
ensure that we have a uniform way of using pluggable caches that is
transparent to the user.

Just like with the user cache, a test module is introduced. As more
cache modules get introduced, this can be adapted to test new caching
modules by changing a single line.

### DIFF
--- a/lib/nostrum/application.ex
+++ b/lib/nostrum/application.ex
@@ -1,7 +1,6 @@
 defmodule Nostrum.Application do
   @moduledoc false
 
-  alias Nostrum.Cache.GuildCache
   use Application
 
   require Logger
@@ -34,7 +33,6 @@ defmodule Nostrum.Application do
     :ets.new(:presences, [:set, :public, :named_table])
     :ets.new(:guild_shard_map, [:set, :public, :named_table])
     :ets.new(:channel_guild_map, [:set, :public, :named_table])
-    :ets.new(GuildCache.tabname(), [:set, :public, :named_table])
   end
 
   defp check_token, do: check_token(Application.get_env(:nostrum, :token))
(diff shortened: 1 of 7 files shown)


jchristgit:
image

jchristgit:
Thanks!

PR #343, opened 2021-10-10
Consolidate channel docs

jchristgit:
Gobble in changes from https://github.com/Kraigie/nostrum/pull/324 plus upstream per https://github.com/Kraigie/nostrum/pull/324#issuecomment-936945611.


jchristgit:
Squashed on `git` CLI with author information preserved since Joe and me were a bit unsure about how that would work with a squash merge.

PR #344, opened 2021-10-10
Add assets directory to ex_doc config

jb3:
Adds an assets directory for any images we want to include in documentation, as of now just a channel types diagram.


jb3:
Thanks!

PR #346, opened 2021-10-12
     Implement pluggable presence caching.     

jchristgit:
Based off `pluggable-guild-cache` to prevent merge conflicts.
    
    This uses the same approach as the pluggable guild cache, in that we
    have a base behaviour and dispatcher for the presence cache, and a
    ETS-based implementation chosen by default which uses the current
    implementation. A small test is added to ensure correct cache function.


Th3-M4jor (approved):
This one looks good to me too

jchristgit:
@Th3-M4jor thank you for the reviews!

jchristgit:
Thanks!

PR #348, opened 2021-10-12
Steps towards 0.5-rc

Th3-M4jor:
- [x] Update docs to use application commands instead of slash commands
- [x] Replace `:poison` with `:jason` as the JSON library
- [x] Update `@spec` for some Api functions to be more descriptive than `map()` 
- [x] Inject `member.user` into `:user` on the `Interaction` struct
- [x] Document slash command autocomplete

Maybe?
- [x] Update all calls to `Jason.encode` to instead be `Jason.encode_to_iodata` and modify `@spec` accordingly
- [x] Rework `Nostrum.Api.create_multipart_body` etc to use iolists

Copied from the docs for Jason:
>This function should be preferred to `encode/2`, if the generated JSON will be handed over to one of the IO functions or sent over the socket. The Erlang runtime is able to leverage vectorised writes and avoid allocating a continuous buffer for the whole resulting string, lowering memory use and increasing performance.


jb3 (dismissed):
Minor note: `docs/static/Slash commands.md` should probably also be renamed to `docs/static/Application Commands.md`, and any hyperlinks should be updated accordingly.

Th3-M4jor:
Fixed it, updated a few more broken doc links while I was at it

jchristgit (approved):
A very nice change, thank you!

Code looks great, I also tested a few hotspots (e.g. multipart), and everything worked perfectly. Dialyzer is also happy.

jchristgit:
Thanks!

PR #349, opened 2021-10-30
change gateway to v8

Th3-M4jor:
Like it says on the tin.

This was the only part still using v6, everything should already be v8 ready from when we updated the api version to v8.


jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #350, opened 2021-11-03
Add guild_id to a guild's member_add event dispatch

RealVidy:
Most people will probably need to know in which guild a member was added so we should add the guild_id to the payload (that's what `member_remove` and `member_update` already do).

Can you please suggest a way to make this change backwards compatible?

### DIFF
--- a/lib/nostrum/shard/dispatch.ex
+++ b/lib/nostrum/shard/dispatch.ex
@@ -165,7 +165,7 @@ defmodule Nostrum.Shard.Dispatch do
 
   def handle_event(:GUILD_MEMBER_ADD = event, p, state) do
     UserCache.create(p.user)
-    {event, GuildCache.member_add(p.guild_id, p), state}
+    {event, {p.guild_id, GuildCache.member_add(p.guild_id, p)}, state}
   end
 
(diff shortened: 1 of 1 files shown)


jchristgit:
Ahhhh, good catch, I broke this with my pluggable guild caching, sorry for that!

The current (stable) behaviour was broken recently, so this PR fixes it properly - so there's no need to make this backwards compatible.

That said, to accommodate for other guild cache implementations and not have to redo documentation, it would be correct to add this in the `GUILD_MEMBER_ADD` event:

```elixir
  def handle_event(:GUILD_MEMBER_ADD = event, p, state) do
    UserCache.create(p.user)
    {event, GuildCache.member_add(p.guild_id, p), state}
  end
```

Instead of returning the result of `GuildCache.member_add` as the second field here, we should return `{p.guild_id, GuildCache.member_add...`. We can then skip the changes in the ETS-based cache. Could you add that change?

RealVidy:
I've made the change, is it what you had in mind?

Thanks for the quick review!

jchristgit:
Yup, this looks good! Thank you!

PR #352, opened 2021-11-03
Fix member cast in VoiceState.to_struct

RealVidy:
`member` is apparently an atom and not a string key in voice state payloads


jchristgit:
@BrandtHill you left out updating the access here in 67c3e5db955910a3d34f5acd94fe850b8bf8dc15, can you remember why?

jchristgit (approved):
Thanks!

PR #357, opened 2021-11-04
Enable component parsing.

### LINKED ISSUE #323: Define component type & re-enable component parsing in messages
See #322

jchristgit:
Define component type and re-enable parsing of message components.

Fixes #323.

### DIFF
--- a/lib/nostrum/struct/message/component.ex
+++ b/lib/nostrum/struct/message/component.ex
@@ -16,7 +16,7 @@ defmodule Nostrum.Struct.Message.Component do
   defstruct [
     :type,
     :custom_id,
-    :disabled?,
+    :disabled,
     :style,
     :label,
     :emoji,
@@ -58,7 +58,7 @@ defmodule Nostrum.Struct.Message.Component do
 
   Only present for buttons and select menus.
   """
-  @type disabled? :: boolean() | nil
+  @type disabled :: boolean() | nil
 
   @typedoc """
   An integer representing the style of the button.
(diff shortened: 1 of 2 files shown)


jchristgit:
Hi @jb3 or @Kraigie, does one of you have time to review this this weekend?

Th3-M4jor (changes requested):
Some minor things in the typespec that need fixing, as well as a `?` in a name that shouldn't be present.

jchristgit:
@Th3-M4jor the `?` was intentional to adhere to the Elixir "standard" of naming these fields. Should we keep it, or is this confusing?

Th3-M4jor:
I'm of the opinion we should try to mirror the official api docs as much as possible with our field names.

jchristgit:
Fair enough 👍

lib/nostrum/struct/message/component.ex
```diff
@@ -0,0 +1,196 @@
...
+
+  Only present for action rows.
+  """
+  @type components :: [t()]
+
+  @typedoc "Represents a message component."
+  @type t :: %__MODULE__{
+          type: type(),
+          custom_id: custom_id(),
+          disabled?: disabled?(),
```
jchristgit, line 166:
```suggestion
          disabled: disabled(),
```

lib/nostrum/struct/message/component.ex
```diff
@@ -0,0 +1,196 @@
...
+          components: components()
+        }
+
+  @doc false
+  @spec to_struct(map()) :: t()
+  def to_struct(map) do
+    %__MODULE__{
+      type: map.type,
+      custom_id: map[:custom_id],
+      disabled?: map[:disabled],
```
jchristgit, line 184:
```suggestion
      disabled: map[:disabled],
```

jb3 (approved):
LGTM! Some of the markdown links wrapping onto newlines is a bit peculiar but I get why you've done it.

jb3:
Thanks!

PR #360, opened 2021-11-12
Duplicate guild creates should be a cache replace

Th3-M4jor:
This fixes an issue where if a ready event is re-emitted for any reason a fatal error in the cache would occur due to disallowing duplicate Guild inserts.

Instead they are now treated as replace operations.


jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #362, opened 2021-11-15
fix: Parse role permissions if they're a string

Ovyerus:
Noticed when trying to get guild permissions for a user that permissions where being given as a string for some reason, resulting in an arithmetic error with the function using `:erlang.bor`, and sure enough listing roles and their permissions gives a list of strings:
```elixir
iex(3)> Nostrum.Cache.GuildCache.get!(441529557691006977).roles |> Enum.map(fn {_, role} -> role.permissions end)
["1071698533952", "6546644544", "8559918328"]
```

This adds an extra step in `Role.to_struct` that parses permissions if its a string, and corresponding test to make sure it works properly.


jchristgit (approved):
Clean and good Elixir code. Thank you!

jchristgit:
Thanks!

queer:
cool and good

PR #363, opened 2021-11-16
Decode Member from modify_guild_member endpoint response

PixeLInc:


Th3-M4jor:
Mostly a nitpick, but can you also modify the return value for the example in the doc comments to go with this fix?

jb3 (commented):
Looks good to me!

jb3 (approved):
Why did they make the "comment" button green and the approve button grey?

jb3:
Thanks!

Issue #355, opened 2021-11-04
Pluggable channel cache


jchristgit:
Implement pluggable channel cache, similar to the existing pluggable guild cache.

gbottari:
Hi! We (@panoramix360 and I) are trying to develop our Elixir skills and came across this project. Maybe this issue could be a good place to start, if that's ok.

What we need to change is `Nostrum.Cache.ChannelCache` to have similar extensibility as `Nostrum.Cache.GuildCache`, right? We welcome any tips and guidelines on this.

jchristgit:
Hi @gbottari and @panoramix360! Yes, we would welcome contributions here.

As far as guidelines go, the rough outline is to do the following:
- Find which functions a cache is required to implement (external functions)
- Create a behaviour with the functions selected above
- Move the current implementation into a "submodule", e.g. `Nostrum.Cache.ChannelCache.ETS`
- Add dispatch logic from the behaviour module to the configured cache module

.. and of course, make sure to include documentation for it 🙂. I'll assign you to the issue, let me know if you need help!

panoramix360:
Hi @jchristgit!

I'll discuss with @gbottari so we can split the tasks and try to work together on this.

We'll let you know if we need any help :D

Thank you!

jchristgit:
Thanks @panoramix360! Looking forward to hearing from you. 🙂

panoramix360:
Hi @jchristgit!

We (me and @gbottari) already finished the rough implementation of what you asked for this issue :tada:

We noticed that there isn't any tests for ChannelCache so we are creating these but we had some problems with an ETS table that is needed for ChannelCache.get/1 that is only created on the Application.setup_ets_tables/0.

We didn't find any examples on using Application on a test so just for now we did the setup like this.

```
setup do
  Application.setup_ets_tables()
  [pid: start_supervised!(@cache)]
end
```
The table is `:channel_guild_map`.

If you have any suggestions, feel free to let us know :)

We hope to open the PR until friday!

Thanks

jchristgit:
Closed via #365. Thank you kindly!

PR #365, opened 2021-11-25
Converting ChannelCache to be pluggable

panoramix360:
Work in progress PR resolving issue #355.

We made the adjustments to transform ChannelCache to be pluggable just as asked and created tests for this.

The PR is in Work in Progress because we need to update the Docs on the files created.

:)

### DIFF
--- a/lib/nostrum/cache/channel_cache.ex
+++ b/lib/nostrum/cache/channel_cache.ex
@@ -36,7 +36,8 @@ defmodule Nostrum.Cache.ChannelCache do
 
   ## Behaviour specification
 
-  @callback get(Channel.id() | Nostrum.Struct.Message.t()) :: {:error, reason} | {:ok, Channel.t()} | {:error, reason}
+  @callback get(Channel.id() | Nostrum.Struct.Message.t()) ::
+              {:error, reason} | {:ok, Channel.t()} | {:error, reason}
 
   @callback get!(Channel.id() | Nostrum.Struct.Message.t()) :: no_return | Channel.t()
 
--- a/lib/nostrum/cache/channel_cache.ex
+++ b/lib/nostrum/cache/channel_cache.ex
@@ -1,23 +1,50 @@
 defmodule Nostrum.Cache.ChannelCache do
   @default_cache_implementation Nostrum.Cache.ChannelCache.ETS
   @moduledoc """
-  Cache for channels.
+  Cache behaviour & dispatcher for channels.
 
-  The ETS table name associated with the Channel Cache is `:channels`. Besides the
-  methods provided below you can call any other ETS methods on the table.
+  You can call the functions provided by this module independent of which cache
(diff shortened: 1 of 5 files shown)


jchristgit:
Hi @panoramix360 and @gbottari, thanks for your contribution!

This looks great so far! I will give it a complete review once you consider it ready.

As for your question here:

> We noticed that there isn't any tests for ChannelCache so we are creating these but we had some problems with an ETS table that is needed for ChannelCache.get/1 that is only created on the Application.setup_ets_tables/0.

Due to the callbacks implementing the supervisor behaviour (just like in `Nostrum.Cache.GuildCache.ETS`), you should be able to skip the separate `Application.setup_ets_tables()` call, as the correct table should be created by `init`.

gbottari:
Hi @jchristgit, thanks for your kind words!

I guess the problem we ran into is that `ChannelCache.ETS.lookup()` actually relies on `GuildCache.select_by()` behind the scenes. `select_by()` in turn, uses `Mapping.ChannelGuild.get_guild()` which relies on table `:channel_guild_map`. The only place table `:channel_guild_map` is initialized is in `Application.setup_ets_tables()`.

As we see it right now the only way to test ChannelCache is to initialize `:channel_guild_map` table somehow or maybe work around this dependency somehow.

jchristgit:
Hi @gbottari. Of course, I completely missed that, sorry! Yeah, in that case, the current startup logic is perfectly fine. 👍

panoramix360:
Hi @jchristgit, we finished the PR :)

Waiting for your approval.

:+1:

lib/nostrum/cache/channel_cache.ex
```diff
@@ -1,24 +1,67 @@
...
+  by setting the `:channels` field to a different module implementing the
+  `Nostrum.Cache.ChannelCache` behaviour. Any module below
+  `Nostrum.Cache.ChannelCache` can be used as a cache.
+
+  ## Writing your own channel cache
+
+  As with the other caches, the channel cache API consists of two parts:
+
+  - The functions that the user calls, (none in this module as the time of
+    writing)
```
jchristgit, line 22:
```suggestion
  - The functions that the user calls, currently only `get/1` and `get!/1`
```

jchristgit (changes requested):
Thanks for the PR! I've gotten around to testing this. We need to do two things:

- Remove the now obsolete creation of the `:channels` ETS table in `setup_ets_tables` in `lib/nostrum/application.ex`
- Add the channel cache to the `CacheSupervisor` children in `lib/nostrum/cache/cache_supervisor.ex`

panoramix360:
@jchristgit thank you for your suggestions, I already changed and commited!

jchristgit (approved):
This all looks great now, thank you very much!

jchristgit:
Thanks!

PR #366, opened 2021-12-04
fix: parse string permissions for overwrites

Ovyerus:
Related to #362, should've checked if this was occurring elsewhere in that PR, but alas here I am. Looks like the same issue as in that PR occurs with permission overwrites for channels, noticed this when trying to use `Struct.Guild.Member.guild_channel_permissions` and got the same `:erlang.bor` error.

Thanks Discord!


Ovyerus:
Bump? 👀

Th3-M4jor:
This looks good to me. Gotta love the consistent inconsistency.

LGTM

jchristgit (approved):
Thanks!

If this trend continues, we should probably make this a helper function at some point.

jchristgit:
Thanks!

PR #369, opened 2021-12-20
Modify UserCache to use Supervisor

### LINKED ISSUE #354: Move user cache to supervisor behaviour
Just like with the guild cache, move the user cache to the `Supervisor` behaviour, such that setup logic and child processes can be started cleanly.

Th3-M4jor:
Modified the `UserCache` to use a Supervisor behaviour, and simplified the compile_env call in `GuildCache`.

Closes #354

### DIFF
--- a/lib/nostrum/cache/user_cache.ex
+++ b/lib/nostrum/cache/user_cache.ex
@@ -82,7 +82,6 @@ defmodule Nostrum.Cache.UserCache do
   """
   @callback delete(snowflake :: User.id()) :: :noop | User.t()
 
-
   ## Dispatching
   defdelegate get(id), to: @configured_cache
   defdelegate create(payload), to: @configured_cache
--- a/lib/nostrum/cache/user_cache.ex
+++ b/lib/nostrum/cache/user_cache.ex
@@ -2,22 +2,36 @@ defmodule Nostrum.Cache.UserCache do
   @default_cache_implementation Nostrum.Cache.UserCache.ETS
   @moduledoc """
   Cache behaviour & dispatcher for users.
+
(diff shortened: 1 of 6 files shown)


lib/nostrum/cache/user_cache.ex
```diff
@@ -2,22 +2,36 @@ defmodule Nostrum.Cache.UserCache do
   @default_cache_implementation Nostrum.Cache.UserCache.ETS
   @moduledoc """
   Cache behaviour & dispatcher for users.
+
+  You can call the functions provided by this module independent of which cache
+  is configured, and it will dispatch to the configured cache implementation.
+
+  By default, #{@default_cache_implementation} will be used for caching guilds.
```
jchristgit, line 9:
```suggestion
  By default, #{@default_cache_implementation} will be used for caching users.
```

lib/nostrum/cache/user_cache.ex
```diff
@@ -2,22 +2,36 @@ defmodule Nostrum.Cache.UserCache do
...
+  You can override this in the `:caches` option of the `:nostrum` application
+  by setting the `:users` field to a different module implementing the
+  `Nostrum.Cache.UserCache` behaviour.
+
+  See the documentation for the `Nostrum.Cache.GuildCache` module for more details.
```
jchristgit, line 14: Doesn't this belong into the `Nostrum.Cache.UserCache.ETS` module, with the typo fixed?

jchristgit (changes requested):
Very nice! Thank you!

jchristgit (approved):
Thanks!

PR #370, opened 2021-12-20
Fix bug with rogue watchdog timer.

### LINKED ISSUE #368: Audio timeout not being cancelled after calling `Nostrum.Voice.pause/1` and `Nostrum.Voice.resume/1`
In a barebones Nostrum(master branch) project, I'm able to consistently reproduce this issue doing the following in IEx:

```elixir
guild_id = <GUILD ID>
channel_id = <CHANNEL_ID>
url = "https://www.youtube.com/watch?v=o7fgFaXKVa0" # Serial experiments Lain opening song

alias Nostrum.Voice

Voice.join_channel(guild_id, channel_id)
Process.sleep(1000)
Voice.play(guild_id, url, :ytdl)
Process.sleep(10_000)
Voice.pause(guild_id)
Process.sleep(2000)
Voice.resume(guild_id)

```

The song will play for a while(10 seconds is enough for me to let it buffer and play for a couple of seconds), then pause, then resume and after 500ms if will stop playing.

The 500 seems to be the `:audio_timeout` option for Nostrum, according to this line:
https://github.com/Kraigie/nostrum/blob/e08d15dd18cfdeefcf1f12c63d676dbf6d5e1f0f/lib/nostrum/voice/audio.ex#L111

The troublesome code seems to be in this function:
https://github.com/Kraigie/nostrum/blob/e08d15dd18cfdeefcf1f12c63d676dbf6d5e1f0f/lib/nostrum/voice/audio.ex#L110-L125

For some reason, after a pause the timer created in line 112 is not being cancelled, so the bot stops playing. To verify this timer is the issue, I commented the call to `Proc.stop` here:
https://github.com/Kraigie/nostrum/blob/e08d15dd18cfdeefcf1f12c63d676dbf6d5e1f0f/lib/nostrum/voice/audio.ex#L310-L314
And the issue "goes away" (of course keeping the process alive in this case is surely a bug though I don't know enough to know how).

Changing the `:audio_timeout` option won't solve this. If I set it to, say, `15_000`, then the bot will resume playing and stop after 15 seconds.

BrandtHill:
Occasionally the watchdog timer responsible for detecting end of input is fired after `pause/1` is called, which causes the sound to abruptly end after `resume/1`. 

Fixes #368

### DIFF
--- a/lib/nostrum/voice/audio.ex
+++ b/lib/nostrum/voice/audio.ex
@@ -308,7 +308,7 @@ defmodule Nostrum.Voice.Audio do
   end
 
   def on_stall(%VoiceState{} = voice) do
-    unless is_nil(voice.ffmpeg_proc) do
+    if VoiceState.playing?(voice) and not is_nil(voice.ffmpeg_proc) do
       Proc.stop(voice.ffmpeg_proc)
     end
   end


jchristgit (approved):
Rock solid.

PR #375, opened 2022-01-05
add support for `guild_scheduled_events`

Th3-M4jor:
Adds support for `guild_scheduled_events`

- [x] Add API support
- [x] Add gateway support


jb3 (changes requested):
All looks good and works locally!

In accordance with #378 it'd be dope if we could serialise/deserialise the datetime types when going to and from the API. I'll open another issue to ensure we're doing this generally across the library.

jb3 (approved):
Looks good to me!


In future credo bumps should probably be a different PR but it's not a major problem.

jb3:
Thanks!

PR #377, opened 2022-01-06
Fix GuildCache.ETS not including guild_id in role_create and role_update methods.

### LINKED ISSUE #376: :GUILD_ROLE_CREATE and :GUILD_ROLE_UPDATE are not including the guild id in the events.
:GUILD_ROLE_CREATE and :GUILD_ROLE_UPDATE are supposed to include the guild id as part of the event, but they are not. My best **guess** is that something as part of the pluggable guild cache change messed this up. I think the return values of GuildCache.role_create and GuildCache.role_update are supposed to include the guild_id in them after updating the ETS table. 
 
 See [here](https://github.com/Kraigie/nostrum/blob/168367916e2d890a1dcc6e19389ac2dbc9666e37/lib/nostrum/cache/guild_cache/ets.ex#L268)
 
If this is the correct change, id be happy to make a quick PR for it.

wkpatrick:
I got no issues with the typespecs I changed, but this is the first time I have worked with them before so do double check I got those right.

Closes #376 

Thanks!

### DIFF
--- a/lib/nostrum/cache/guild_cache.ex
+++ b/lib/nostrum/cache/guild_cache.ex
@@ -263,7 +263,7 @@ defmodule Nostrum.Cache.GuildCache do
 
   Return the casted role.
   """
-  @callback role_create(Guild.id(), role :: map()) :: Role.t()
+  @callback role_create(Guild.id(), role :: map()) :: {Guild.id(), new_role :: Role.t()}
 
   @doc """
   Delete the given role on the given guild.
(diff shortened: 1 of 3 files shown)


jb3 (approved):
Thanks for the contribution! Was going to drop a comment about the failing tests but you got to it right before I clicked submit!

This all looks great, will get it merged!

jb3:
Thanks!

PR #382, opened 2022-01-10
Improve consistency of `to_struct` fns and remove `?` from field names

Th3-M4jor:
Now all the `to_struct/1` functions should follow the same pattern.

And from `VoiceState` I removed the `?` from field names to maintain consistency across the rest of the library and the 1:1 field mapping with the API.

While writing this PR, I noticed that I missed a couple places when replacing ISO8601 strings with the `DateTime` struct as well as a couple Unix Timestamp fields. I'll open a PR for those after this gets merged.


jb3 (approved):
I can't test every struct here, but the ones I have are still working absolutely fantastic, as expected! Good PR!

PR #383, opened 2022-01-11
Add the three time fields I missed earlier

Th3-M4jor:
Two PRs ago I missed the timestamp field in `TypingStart` and two fields in `Message`. This addresses that.

This also now addresses #384 because I forgot I had an open PR


jb3 (approved):

jb3:
Thanks!

PR #385, opened 2022-01-12
Allow for uploading files in interaction responses

Th3-M4jor:
Adds the option to include files when creating interaction responses. As an added bonus, the file's body can be an iolist and work just fine.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -3469,6 +3469,29 @@ defmodule Nostrum.Api do
   directly. See `create_interaction_response/2`.
   """
   @spec create_interaction_response(Interaction.id(), Interaction.token(), map()) :: {:ok} | error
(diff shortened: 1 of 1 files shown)


jb3 (approved):
All the changes look good. It might be nice to provide `file` as a backwards compatible option, thoughts @jchristgit?

Th3-M4jor:
Yeah I might as well add that, in addition to a check for when both `:file` and `:files` are included

Th3-M4jor:
Actually, we do provide `:file` as a backwards compatible option for both `execute_webhook` and `create_interaction_response`. It's just not documented because it should be treated as "soft deprecated" in my opinion.

jchristgit (approved):
Absolutely rock solid. Thanks!

jchristgit:
Thanks!

PR #392, opened 2022-02-18
Use native ports instead of porcelain

BrandtHill:
Removing Porcelain as a dependency, which has not had a new release in many years, and using native Erlang ports instead with fewer processes than Porcelain uses.
Also takes care of closing secondary processes piping into ffmpeg when ffmpeg is closed.
Various readability improvements.
Yes it works.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -7,6 +7,4 @@ config :nostrum,
 
 config :logger, :console, metadata: [:shard, :guild, :channel]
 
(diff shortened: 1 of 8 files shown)


jchristgit (approved):
This is super great, finally we're getting rid of the compilation warnings! Thank you for doing this!

jchristgit:
Thanks!

PR #394, opened 2022-02-19
Kraigie master

danhawkins:


bdanklin:
@danhawkins I see you merged this change in your own fork. It will crash on a number of API functions using the official gun rc.

danhawkins:
Which change specifically would cause that?

danhawkins:
OK, I closed my fork anyway and switch to 0.5.0 stable, I only forked to get around some dependency stuff, which I should've just used override for. Appreciate the comment though

PR #395, opened 2022-02-21
Fix bug where voice player stalls on initial timeout

BrandtHill:
`player_pid` isn't current on the first iteration of `player_loop` so the `on_stall` check fails and causes the player to stall if the audio source times out or fails before producing any audio.


jchristgit (approved):

jchristgit:
Thanks!

PR #396, opened 2022-02-22
Streamline Streamlink

BrandtHill:
Streamlink version 3.x.x works with more urls out of the box. Removing the extra step of using `youtube-dl` to grab the raw stream url dramatically improves startup time for `play/4` and allows the user to pass any url directly to streamlink, while docs show how to fetch the "raw stream url" using `youtube-dl` if needed. 

Additionally fixing a bug with stateful raw audio source which I caught coincidentally because I didn't have a proper way to test it before and I don't think anyone has used it yet.

Up-to-date version of streamlink can be installed with `pip install --upgrade streamlink` (3.1.1 is current newest and recommended).

Testing can be done with these helper funcs
```elixir
  def join_voice, do: Voice.join_channel(@guild_id, @channel_id)
  def leave_voice, do: Voice.leave_channel(@guild_id)

  def play_stream_yt(opts \\ []),
    do: Voice.play(@guild_id, "https://www.youtube.com/watch?v=5qap5aO4i9A", :stream, opts)

  def play_stream_tt(opts \\ []),
    do: Voice.play(@guild_id, "https://twitch.tv/lvndmark", :stream, opts)

  def play_raw_stateful(opts \\ []) do
    url = "https://soundcloud.com/theshadowchild/shadow-child-sinden-get-at"
    p = Audio.spawn_ffmpeg(url, :ytdl, opts)
    s = Voice.Ports.get_stream(p)
    Voice.play(@guild_id, s, :raw_s)
  end
```

### DIFF
--- a/docs/static/Voice.md
+++ b/docs/static/Voice.md
@@ -31,13 +31,23 @@ When `Nostrum.Voice.play/4` is called with `:ytdl` for the `type` parameter, `yo
 run with options `-f bestaudio -q -o -`, which will attempt to download the audio at the given url and pipe it to `ffmpeg`.
 
 ## streamlink
-Nostrum also has support for `streamlink`, yet another powerful command line utility for downloading live streams from online video streaming services.
+Nostrum also has support for `streamlink`, yet another powerful command line utility 
+for downloading live streams from online video streaming services.
 By default Nostrum will look for the executable `streamlink` in the system path. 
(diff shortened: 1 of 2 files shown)


jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #400, opened 2022-02-28
fix huge exploit

BrandtHill:


jchristgit (approved):
Thank you for your contribution.

While we appreciate the security fix being pushed into this branch, we are asking you to send future security reports to security@craig.cat instead, where we can handle and disclose the vulnerability as soon as time allows.

PR #401, opened 2022-02-28
Threads support

Th3-M4jor:
Add support for Threads

- [x] Bump Gateway version to v9
- [x] Add new Thread events
- [x] Add new API routes
- [x] Add thread specific fields to existing structs

Why bump to API v9 in this PR and not skip directly to v10? Because these should be able to be done without it being a breaking change, allowing for one last hex release before v10 and `0.6.0`

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -3900,6 +3900,7 @@ defmodule Nostrum.Api do
   - `rate_limit_per_user`: Rate limit per user in seconds, can be set to any value in `0..21600`.
 
   """
+  @doc since: "0.5.1"
   @spec start_thread_with_message(
           Channel.id(),
           Message.id(),
@@ -3940,6 +3941,7 @@ defmodule Nostrum.Api do
   - `invitable`: whether non-moderators can add other non-moderators to a thread; only available when creating a private thread defaults to `false`.
   - `rate_limit_per_user`: Rate limit per user in seconds, can be set to any value in `0..21600`.
   """
+  @doc since: "0.5.1"
   @spec start_thread(Channel.id(), thread_without_message_params, AuditLogEntry.reason()) ::
           {:ok, Channel.t()} | error
   def start_thread(channel_id, options, reason \\ nil) do
@@ -3956,6 +3958,7 @@ defmodule Nostrum.Api do
   @doc """
   Returns a thread member object for the specified user if they are a member of the thread
   """
+  @doc since: "0.5.1"
   @spec get_thread_member(Channel.id(), User.id()) :: {:ok, ThreadMember.t()} | error
   def get_thread_member(thread_id, user_id) do
     request(:get, Constants.thread_member(thread_id, user_id))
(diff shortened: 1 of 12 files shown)


Manu590:
LGTM!

Manu590:
cool idea too!

Manu590:
make sure to check bbts's

BrandtHill:
API and events look good. Are there functions or endpoints for editing/archiving threads?

Th3-M4jor:
Yes, you edit/delete/archive them the same way you edit/delete a channel. See here https://discord.com/developers/docs/topics/threads#editing-deleting-threads

lib/nostrum/api.ex
```diff
@@ -3853,6 +3879,255 @@ defmodule Nostrum.Api do
...
+  Threads are sorted by the `archive_timestamp` field, in descending order.
+
+  ## Response body
+  Response body is a map with the following keys:
+  - `threads`: A list of channel objects.
+  - `members`: A list of `ThreadMemer` objects, one for each returned thread the current user has joined.
+  - `has_more`: A boolean indicating whether there are more archived threads that can be fetched.
+
+  ## Options
+  - `before`: Returns threads before this timestamp, can be either a `DateTime` or (ISO8601 timestamp)[`DateTime.to_iso8601/3`].
```
jchristgit, line 4016: I think this link is the wrong way around, isn't it?

```suggestion
  - `before`: Returns threads before this timestamp, can be either a `DateTime` or [ISO8601 timestamp](`DateTime.to_iso8601/3`).
```

lib/nostrum/struct/event/thread_members_update.ex
```diff
@@ -0,0 +1,74 @@
+defmodule Nostrum.Struct.Event.ThreadMembersUpdate do
+  @moduledoc """
+  Struct representing a thread members update event.
+
+  This event is sent whenever a user is added or removed from a thread.
+
+  If the current user does not have the `GUILD_MEMBERS` intent,
+  this event will only be send when the current user is added to or removed from a thread.
```
jchristgit, line 8: Small typo:
```suggestion
  this event will only be sent when the current user is added to or removed from a thread.
```

lib/nostrum/struct/thread_member.ex
```diff
@@ -0,0 +1,68 @@
...
+
+  @typedoc """
+  The timestamp of when the user last joined the thread
+  """
+  @typedoc since: "0.5.1"
+  @type join_timestamp :: DateTime.t()
+
+  @typedoc """
+  Any user-thread settings
+  """
```
jchristgit, line 37: This also needs the `@typedoc since: "0.5.1"`.

Alternatively, would it be more readable in the docs to simply have a `@moduledoc since: "0.5.1"` and keep `@typedoc since:` for any new struct fields that were added after the module was added?

jchristgit (changes requested):
Thank you! I have three minor nitpicks, afterwards this is good to merge.

jchristgit:
Also, a request, if you have the time: could you add `@doc since: "0.5.1"` to all the new functions in `Nostrum.Api` as well? Thank you!

jchristgit (approved):

jchristgit:
Thanks!

PR #402, opened 2022-03-04
Fix audit log entries not being saved

frm:
`Nostrum.Struct.Guild.AuditLog` has an `:entries` key, while the API
returns an `:audit_log_entries`. However, when casting the response, the
`:audit_log_entries` field was casted. Since there was no `:entries`
field in the resulting map, when calling `struct/2`, `:entries` was
always `nil`.


jchristgit (approved):
Thanks!

PR #407, opened 2022-03-12
Ogg encoding and rtp ext fixes

BrandtHill:
Three main things

1. **Fix bug with generating ogg pages with incorrect page_segments value.** - _page_segments is the byte length of the segment_table, and when segments exceed 254 bytes they take up multiple bytes in the segment table. This was leading to some encoding errors when some opus packets were 255 bytes in length or greater._
2. **Properly skip over rtp header extension.** - _the ext_length is actually the number of 32 bit words of the extension, not the number of extension elements to iterate through. Thanks to @imayhaveborkedit for pointing me to the first paragraph of the rfc 🤡_ 
3. **Place binary pattern matches in function headers versus body.** - _good practice_

### DIFF
--- a/lib/nostrum/voice/opus.ex
+++ b/lib/nostrum/voice/opus.ex
@@ -19,7 +19,7 @@ defmodule Nostrum.Voice.Opus do
   discord.js appears to be the only other library that officially implements receiving
   audio as a part of their voice modules, and in their stripping of the RTP header
   extension they account for an extra "Discord byte" but it looks like they are actually
-  misinterpreting the 16-bit extension lenth as the number of extension elements, then
+  misinterpreting the 16-bit extension length as the number of extension elements, then
   iterating through each one-byte header and variable-length element but incorrectly
   incrementing the offset by an extra 1, which causes them to land on an extra 1 byte,
   instead a full extension (1 byte header with 1 byte extension).
--- a/lib/nostrum/voice/opus.ex
+++ b/lib/nostrum/voice/opus.ex
@@ -132,7 +132,7 @@ defmodule Nostrum.Voice.Opus do
       when is_list(body) do
     segment_table_raw =
       body
-      |> Enum.map(&byte_size(&1))
+      |> Enum.map(&byte_size/1)
       |> encode_ogg_segment_table()
 
     %{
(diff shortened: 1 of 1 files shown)


lib/nostrum/voice/opus.ex
```diff
@@ -4,58 +4,41 @@ defmodule Nostrum.Voice.Opus do
...
-  the offset by an extra 1, which causes them to land on an extra 1 byte,
+  RFC 8285 describes the format of RTP header extensions. Discord uses RTP
+  extensions with the one-byte header format that begins with a 0xBEDE pattern
+  followed by a 16 bit extension length. This extension length represents the length
+  of the following extension in 32-bit words.
+
+  discord.js appears to be the only other library that officially implements receiving
+  audio as a part of their voice modules, and in their stripping of the RTP header
+  extension they account for an extra "Discord byte" but it looks like they are actually
+  misinterpreting the 16-bit extension lenth as the number of extension elements, then
```
jchristgit, line 22:
```suggestion
  misinterpreting the 16-bit extension length as the number of extension elements, then
```

lib/nostrum/voice/opus.ex
```diff
@@ -4,58 +4,41 @@ defmodule Nostrum.Voice.Opus do
...
+  iterating through each one-byte header and variable-length element but incorrectly
+  incrementing the offset by an extra 1, which causes them to land on an extra 1 byte,
   instead a full extension (1 byte header with 1 byte extension).
```
jchristgit, line 25: thinking about this - should we maybe report it to their devs?

lib/nostrum/voice/opus.ex
```diff
@@ -147,35 +130,37 @@ defmodule Nostrum.Voice.Opus do
         } = _state
       )
       when is_list(body) do
+    segment_table_raw =
+      body
+      |> Enum.map(&byte_size(&1))
```
jchristgit, line 135:
```suggestion
      |> Enum.map(&byte_size/1)
```

My BEAM-fu tells me this avoids an extra anonymous function, so there is space for an insanely important optimization here, saving us lightyears of CPU time.

lib/nostrum/voice/opus.ex
```diff
@@ -147,35 +130,37 @@ defmodule Nostrum.Voice.Opus do
...
       granule_position: granule_position,
       bitstream_serial: bitstream_serial,
       page_sequence: page_sequence,
       crc_checksum: 0,
-      page_segments: body |> Enum.count(),
-      segment_table_raw:
-        body
-        |> Enum.map(&byte_size(&1))
-        |> encode_ogg_segment_table(),
+      page_segments: segment_table_raw |> byte_size(),
```
jchristgit, line 145: Personal preference, feel free to ignore:
```suggestion
      page_segments: byte_size(segment_table_raw),
```

lib/nostrum/voice/opus.ex
BrandtHill: Yeah probably

jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #408, opened 2022-03-14
Make breaking change to voice listening

BrandtHill:
Make listening more ergonomic. Breaking change. Auto parse RTP header and opus when listening. Optionally return raw RTP packet. Change extract function to expect full RTP packet vs just payload.
For 0.6.0 release. 
did this on me fone🤳


jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #409, opened 2022-03-14
add missing permissions

Th3-M4jor:
Add the missing permission fields to `permissions.ex` and replaces the hex values in the map with bit-shifted values for easier readability


jchristgit:
Thanks!

PR #412, opened 2022-03-15
Declutter multipart functionality in api.

BrandtHill:
Remove buggy and now unused code that was causing some issues when moving from .5 rc to latest
Make request_multipart useful, removing duplicate code to make multipart and file handling easier

Additionally:

- fix spec for request_multipart
- fix bug where execute_webhook and edit_webhook message wasn’t casting to `Message` struct
- make internal function private
- silently handle combining `:embed` and `:embeds` similarly to `:file` and `:files`
- make file form name `files[n]` where `n` begins at 0 according to most recent docs
- support files for all applicable message editing functions

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -252,26 +252,15 @@ defmodule Nostrum.Api do
   def create_message(%Message{} = message, options),
     do: create_message(message.channel_id, options)
 
+  def create_message(channel_id, content) when is_binary(content),
+    do: create_message(channel_id, %{content: content})
+
   def create_message(channel_id, options) when is_list(options),
     do: create_message(channel_id, Map.new(options))
 
   def create_message(channel_id, %{} = options) when is_snowflake(channel_id) do
     options = prepare_allowed_mentions(options) |> combine_embeds()
 
-    if has_files(options),
-      do: create_message_with_multipart(channel_id, options),
-      else: create_message_with_json(channel_id, options)
-  end
-
-  def create_message(channel_id, content) when is_snowflake(channel_id) and is_binary(content),
-    do: create_message_with_json(channel_id, %{content: content})
-
-  defp create_message_with_multipart(channel_id, options) do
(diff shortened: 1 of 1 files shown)


lib/nostrum/api.ex
```diff
@@ -4208,12 +4075,19 @@ defmodule Nostrum.Api do
   @spec request_multipart(atom(), String.t(), any, keyword() | map()) :: {:ok} | error
   def request_multipart(method, route, body, params \\ []) do
     boundary = generate_boundary()
+    body = combine_files(body)
+    files = body.files
+
+    json =
+      body
+      |> Map.delete(:files)
+      |> Jason.encode_to_iodata!()
```
Th3-M4jor, line 4084: Could probably simplify this to
```elixir
{files, body} = combine_files(body) |> Map.pop!(:files)

json = body |> Jason.encode_to_iodata!()
```

Other than that, this looks good to me.

Th3-M4jor (changes requested):

BrandtHill:
Btw please test these. I’ve been writing this on my phone.

BrandtHill:
Fixed the create_interaction_response, could you test again?

Th3-M4jor:
Tested and it works, but now `Api.create_followup_message!/3` is throwing a dialyzer error on my machine. Saying the function call will not succeed.

BrandtHill:
> Tested and it works, but now `Api.create_followup_message!/3` is throwing a dialyzer error on my machine. Saying the function call will not succeed.

I think it was the request_multipart spec which I just fixed

Th3-M4jor:
I found the issue, `Api.execute_webhook` isn't decoding the response into a Message struct, it needs a `|> handle_request_with_decode({:struct, Message})`

jchristgit (approved):
Deleted code ✅ 

Thanks!

jchristgit:
Thanks!

PR #414, opened 2022-03-17
fix improper call to `Integration` cast bug

Th3-M4jor:
Improperly wrote the calls to `Util.cast/2`, this fixes that.


jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #415, opened 2022-03-24
Implement asynchronous voice packet event listening. 

BrandtHill:
Add voice incoming packet events. Add ssrc map. Add a bunch of types for the Voice module.


lib/nostrum/voice/opus.ex
```diff
@@ -333,6 +333,20 @@ defmodule Nostrum.Voice.Opus do
...
+      |> div(960)
+      |> Kernel.-(1)
+      |> max(0)
+      |> (&List.duplicate({nil, <<0xF8, 0xFF, 0xFE>>}, &1)).()
+      |> Enum.reduce(acc, &[&1 | &2])
+      |> (&[x | &1]).()
+    end)
+    |> Stream.map(fn {_, opus} -> opus end)
+    |> Enum.reverse()
+  end
```
jchristgit, line 348: 🪄🪄🪄🪄🪄🪄

jchristgit: I have no idea what this is doing and how it works, but I don't know what i'm doing either and I also work!

jchristgit (approved):
Completely bonkers. Super, super, super cool.

jchristgit:
Thanks!

lib/nostrum/voice/opus.ex
BrandtHill: This function is totally fried so I will explain what this pipe does for posterity:

The RTP packets come in as follows
{{seq, time, ssrc}, opus} so three consecutive packets may look like this:
```elixir
{{8_001, 419_000, 77_777}, <<…>>}
{{8_002, 419_960, 77_777}, <<…>>}
{{8_003, 420_920, 77_777}, <<…>>}
```
Initial sequence and time are random. Each packet the time increases by 960 where 960 samples is 20ms at 48kHz sample rate. If the speaker pauses for 1 second (or 50 packets worth of data) the next packet will have its time of previous + 960*50. 

So what this does is makes the first packet the accumulator for comparison, then iterate through each packet, matching against the current time and the accumulator head’s time (previous packet time)
- get difference between times
- integer divide by 960 to convert from samples to number of 20ms time chunks
- subtract by one to get the number of chunks missing (from a pause in audio)
- ensure not a negative value
- generate an opus silence packet for each missing chunk
- splice any generated silence packets between previous and current audio packets
- finally add our current audio packet to the accumulator

Once that’s done, map the opus packet from the tuple, then reverse it. 

Doing this in a functional language where you have to do things in reverse order and comparing adjacent list elements isn’t trivial definitely doesn’t help make things easier to read.

jchristgit: Thank you very much for the detailed explanation, this makes a lot of sense!

Do you think it would make sense to include your comment in the code, and maybe move the `960` to a module constant e.g. `@samples_per_packet` as well as the silence packet to e.g. `@opus_silence_packet`?

BrandtHill: Yes I’m working on a few changes to makes things a little less arcane

PR #418, opened 2022-04-30
Fix `check_token/1` for variable length tokens

daskycodes:
Generating a Discord Bot Token will sometimes return a longer token format than what the `check_token/1` function is checking for.

This will currently throw the following error:

```elixir
** (Mix) Could not start application nostrum: exited in: Nostrum.Application.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (RuntimeError) Invalid token format, copy it again from the `Bot` tab of your Application
            (nostrum 0.5.1) lib/nostrum/application.ex:40: Nostrum.Application.check_token/1
            (nostrum 0.5.1) lib/nostrum/application.ex:10: Nostrum.Application.start/2
            (kernel 8.2) application_master.erl:293: :application_master.start_it_old/4
```

I wasn't able to reliably find a triggering setting but this is an example token (Already resetted, so it's unusable)

`OTY4NTU2MzQ4MzkwMzkxODU5.G49NjP.pD8PLpKp-Xx8sr-8m1DCxSPTJZdcpcJZOExc1c`

This PR replaces the `Nostrum.Application.check_token/1` function with the `Nostrum.Token.check_token!` function.

It pattern matches on a variable hash for the last part of the binary token and additionally tries to decode the `user_id` which is the first part of the binary token.

### DIFF
--- a/lib/nostrum/token.ex
+++ b/lib/nostrum/token.ex
@@ -48,14 +48,13 @@ defmodule Nostrum.Token do
   end
 
   defp decode_user_id!(user_id) do
-    try do
+    _user_id =
       user_id
       |> :base64.decode_to_string()
       |> :erlang.list_to_integer()
 
-      :ok
-    rescue
-      _ -> raise(@invalid_token_error_message)
-    end
+    :ok
+  rescue
+    _ -> raise(@invalid_token_error_message)
   end
 end
--- a/lib/nostrum/token.ex
+++ b/lib/nostrum/token.ex
@@ -55,6 +55,11 @@ defmodule Nostrum.Token do
 
     :ok
   rescue
-    _ -> raise(@invalid_token_error_message)
+    exception ->
+      reraise(
+        RuntimeError,
+        [message: @invalid_token_error_message, exception: exception],
+        __STACKTRACE__
+      )
   end
 end
--- a/lib/nostrum/token.ex
+++ b/lib/nostrum/token.ex
@@ -1,0 +1,61 @@
+defmodule Nostrum.Token do
+  @moduledoc """
+  A helper module for verifying the Discord bot token.
+
+  The token can be generated in the "Bot" tab of your Application the [Discord Developer Portal](https://discord.com/developers/applications) and
+  can be configured in your config file.
+
+  ```elixir
+  config :nostrum,
+    token: "666" # The token of your bot as a string
+  ```
+  """
+
+  @invalid_token_error_message ~S[Invalid token format. Copy it again from the "Bot" tab of your Application in the Discord Developer Portal.]
+  @no_token_error_message "A bot token needs to be supplied in your config file"
+
+  @doc """
+  Checks if the Discord bot token has the correct format.
+
+  We check if the token is a binary followed by splitting it into 3 parts separated by a dot `"."`.
+  The first part is the Base64 encoded user_id which we decode and parse into as integer.
+  The second part is an encoded timestamp, and the last part an arbitrary cryptographic signature.
+
+  Raises on failure.
+
+  ## Examples
+
+      iex> token = "OTY4NTU2MzQ4MzkwMzkxODU5.G49NjP.pD8PLpKp-Xx8sr-8m1DCxSPTJZdcpcJZOExc1c"
+      iex> Nostrum.Token.check_token!(token)
+      :ok
+
+      iex> token = "ODY4MDcxODUzMDMyMzU3OTc4.YPqU6Q.jNJcq1daGG3otexX3c1LcxCpgpQ"
+      iex> Nostrum.Token.check_token!(token)
+      :ok
+  """
+  def check_token!, do: check_token!(Application.get_env(:nostrum, :token))
+  def check_token!(nil), do: raise(@no_token_error_message)
+
+  def check_token!(<<user_id::binary-size(24), 46, _ts::binary-size(6), 46, _hmac_auth::binary>>) do
+    decode_user_id!(user_id)
+  end
+
+  def check_token!(token) when is_binary(token) do
+    case String.split(token, ".") do
+      [user_id, _timestamp, _hmac_auth] -> decode_user_id!(user_id)
+      _ -> raise(@invalid_token_error_message)
+    end
+  end
+
+  defp decode_user_id!(user_id) do
+    try do
+      user_id
+      |> :base64.decode_to_string()
+      |> :erlang.list_to_integer()
+
+      :ok
+    rescue
+      _ -> raise(@invalid_token_error_message)
+    end
+  end
+end
--- a/lib/nostrum/application.ex
+++ b/lib/nostrum/application.ex
@@ -34,8 +34,7 @@ defmodule Nostrum.Application do
 
   defp check_token, do: check_token(Application.get_env(:nostrum, :token))
   defp check_token(nil), do: raise("Please supply a token")
-  defp check_token(<<_::192, 46, _::48, 46, _::216>>), do: :ok
-  defp check_token(<<_::192, 46, _::48, 46, _::304>>), do: :ok
+  defp check_token(<<_::192, 46, _::48, 46, _rest::binary>>), do: :ok
 
   defp check_token(_invalid_format),
     do: raise("Invalid token format, copy it again from the `Bot` tab of your Application")
--- a/lib/nostrum/application.ex
+++ b/lib/nostrum/application.ex
@@ -3,11 +3,13 @@ defmodule Nostrum.Application do
 
   use Application
 
+  alias Nostrum.Token
+
   require Logger
 
   @doc false
   def start(_type, _args) do
-    Nostrum.Token.check_token!()
+    Token.check_token!()
     check_executables()
     setup_ets_tables()
 
--- a/lib/nostrum/application.ex
+++ b/lib/nostrum/application.ex
@@ -35,6 +35,7 @@ defmodule Nostrum.Application do
   defp check_token, do: check_token(Application.get_env(:nostrum, :token))
   defp check_token(nil), do: raise("Please supply a token")
   defp check_token(<<_::192, 46, _::48, 46, _::216>>), do: :ok
+  defp check_token(<<_::192, 46, _::48, 46, _::304>>), do: :ok
 
   defp check_token(_invalid_format),
     do: raise("Invalid token format, copy it again from the `Bot` tab of your Application")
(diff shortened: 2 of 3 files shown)


Andris0:
Even though the token pattern has been consistent over the years (up to this point I guess), we should probably not rely on things like length of token components, especially the last part, which is an arbitrary cryptographic signature. So I'd say we should replace the token checking function with something like this instead, which should be good enough.
```elixir
defp check_token do
  with token when is_binary(token) <- Application.get_env(:nostrum, :token),
        [user_id, _timestamp, _hmac_auth] <- String.split(token, "."),
        {:ok, user_id} <- Base.decode64(user_id),
        {_user_id, ""} <- Integer.parse(user_id) do
    :ok
  else
    nil -> raise("Please supply a token")
    _ -> raise("Invalid token format, copy it again from the `Bot` tab of your Application")
  end
end
```

bdanklin:
binary pattern matching is fast, I'd propose the below to avoid all those function calls that do essentially nothing.  This should still not accept a user token which is the main point of it.

```elixir
 defp check_token(<<_::192, 46, _::48, 46, _rest::binary>>), do: :ok
 ```

**Comment by daskycodes** [comment]
@bdanklin updated it to pattern match on a variable binary length.

@Andris0 I have gone with the variable names you suggest, which will make the structure of the token we are pattern matching on much clearer.

```elixir
  defp check_token(<<_user_id::binary-size(24), 46, _ts::binary-size(6), 46, _hmac_auth::binary>>) do
    :ok
  end
```

Andris0:
@bdanklin Binary pattern matching is fast and cool, and all, but we're not exactly number crunching a simulation of the entire universe over here or something. It's just a small, one time check at startup. I do not really see the point of favoring solutions just based on speed in this context. The snippet that I suggested is not going to be perceivably any slower.

What I'm suggesting, is just slightly dipping into the substance of the token that got passed in to check for its validity, which seems like a more reasonable and flexible approach, than just relying on encoded lengths of token segments. I looked at 2 other Discord libraries on how they handle this particular scenario (Serenity and Discord.Net), and the approach of just checking for 3 parts and then attempting to decode and parse the `user_id` segment just seems like a better way of doing this. 

And no, rejecting user tokens was not the main point of this check. This was put in place because, on occasion, people were copying other things than the actual bot token from their developer portal page and placing it as the token, which at the time, would return a rather vague error message with a response from the gateway, and so, this was put in place just right after one of these occurrences.

daskycodes:
We can also go with a compromise, keeping the pattern matching and adding another check in the function body.

This way we would end up with the following, which is ~ 4x Faster than the snippet you posted @Andris0 :

```elixir
  defp check_token(nil), do: raise("Please supply a token")

  defp check_token(<<user_id::binary-size(24), 46, _ts::binary-size(6), 46, _hmac_auth::binary>>) do
    try do
      user_id
      |> :base64.decode_to_string()
      |> :erlang.list_to_integer()

      :ok
    rescue
      _ -> raise("Invalid token format, copy it again from the `Bot` tab of your Application")
    end
  end

  defp check_token(_invalid_token), do: raise("Invalid token format, copy it again from the `Bot` tab of your Application")
```

But yes it's `0.37 μs` vs `1.40 μs` on startup...

```elixir
Mix.install([
  {:benchee, "~> 1.0", only: :dev}
])

defmodule Discord do
  def check_token_with_pattern_matching_and_erlang_decoding(nil), do: raise("Please supply a token")

  def check_token_with_pattern_matching_and_erlang_decoding(<<user_id::binary-size(24), 46, _ts::binary-size(6), 46, _hmac_auth::binary>>) do
    try do
      user_id
      |> :base64.decode_to_string()
      |> :erlang.list_to_integer()

      :ok
    rescue
      _ -> raise("Invalid token format, copy it again from the `Bot` tab of your Application")
    end
  end

  def check_token_with_pattern_matching_and_erlang_decoding(_invalid_token), do: raise("Invalid token format, copy it again from the `Bot` tab of your Application")

  def check_token_with_pattern_matching_and_elixir_decoding(nil), do: raise("Please supply a token")

  def check_token_with_pattern_matching_and_elixir_decoding(<<user_id::binary-size(24), 46, _ts::binary-size(6), 46, _hmac_auth::binary>>) do
    with {:ok, user_id} <- Base.decode64(user_id),
         {_user_id, ""} <- Integer.parse(user_id) do
      :ok
    else
      _ -> raise("Invalid token format, copy it again from the `Bot` tab of your Application")
    end
  end

  def check_token_with_pattern_matching_and_elixir_decoding(_invalid_token), do: raise("Invalid token format, copy it again from the `Bot` tab of your Application")

  def check_token_with_string_split(token) do
    with token when is_binary(token) <- token,
         [user_id, _timestamp, _hmac_auth] <- String.split(token, "."),
         {:ok, user_id} <- Base.decode64(user_id),
         {_user_id, ""} <- Integer.parse(user_id) do
      :ok
    else
      nil -> raise("Please supply a token")
      _ -> raise("Invalid token format, copy it again from the `Bot` tab of your Application")
    end
  end
end

token = "OTY4NTU2MzQ4MzkwMzkxODU5.G49NjP.pD8PLpKp-Xx8sr-8m1DCxSPTJZdcpcJZOExc1c"

Benchee.run(
  %{
    "Erlang Pattern Matching" => fn token -> Discord.check_token_with_pattern_matching_and_erlang_decoding(token) end,
    "Elixir Pattern Matching" => fn token -> Discord.check_token_with_pattern_matching_and_elixir_decoding(token) end,
    "Elixir With String Split" => fn token -> Discord.check_token_with_string_split(token) end
  },
  time: 10,
  inputs: %{"token" => token},
  print: [fast_warning: false]
)
```

Andris0:
The whole point of what I was suggesting, was to get rid of length dependent pattern matching at any capacity, for any of the token's individual pieces, so that there would not be these kind of issues in the first place, of what this PR intends to patch. I do not see the point for the compromise solution of using both suggestions, as they are not complimentary in dealing with the issue.

I appreciate you writing up benchmarks for the performance implications that were voiced, but weighing given suggestions based on **_fractions of a microsecond_** for a one time startup check is a mind boggling perversion of the matter.

I just want the check to have 3 things, if the token is present, if it has 3 parts, if the first part actually contains something that resembles an ID, and not deal with hardcoded assumptions of what things like the length of a token needs to be.

Kraigie:
Thanks for the PR @daskycodes! It's not often we get actual benchmarks and the work you've done in that regard seems impressive! I wanted to jump in and give my two cents on the topic -

I think realistically that either approach, be it pattern matching on the binary, or splitting inside the function, will work and be okay. That said, I do think the benefits of future-proofing our implementation through splitting the string do outweigh the performance gains of the pattern matching.

However, if we think that the splitting implementation doesn't future proof our application, or that the performance gains ARE worth it, then I'm open to reconsider. I do want us to keep in mind that as part of the start-up we end up making an API request, so unless this function ends up taking half a second, the impact on the user will likely be minimal as they're already going to be waiting.

I do quite like attempting to decode the first bit of the token as an extra step, just to ensure there weren't errors with copy/pasting or something weird happening.  If we're going to go with the slower method then I think I'd like to add/keep this functionality.

I think another suggestion would include the headers attempting to match on the binary, as well as a fallback function that attempts the string splitting method if all the pattern matches fail. I'm not sure if this maintains the performance gains, but if it does then I think it'd be a happy middle ground. If we end up doing something heavy-handed like this I think we should move this functionality into its own module though! Thoughts on this?

Andris0:
I'm okay with having the string splitting method as a fallback and have something like this
```elixir
defp check_token, do: check_token(Application.get_env(:nostrum, :token))

defp check_token(nil), do: raise("Please supply a token")

defp check_token(<<_user_id::binary-size(24), 46, _ts::binary-size(6), 46, _hmac_auth::binary>>) do
  :ok
end

defp check_token(token) do
  with token when is_binary(token) <- token,
        [user_id, _timestamp, _hmac_auth] <- String.split(token, "."),
        {:ok, user_id} <- Base.decode64(user_id),
        {_user_id, ""} <- Integer.parse(user_id) do
    :ok
  else
    _ -> raise("Invalid token format, copy it again from the `Bot` tab of your Application")
  end
end
```

daskycodes:
@Kraigie thank you for the suggestions.

I do like your way of thinking here and think we can neglect some performance increases in favor of giving the user somewhat earlier feedback with splitting the response and decoding the first bit.

I will try to go with your suggested middleground approach in a separate module.

I have added the `Nostrum.Token.check_token!/1` function with proper moduledocs and tests. Also adding the bang `!` to the function so it's clear that it raises an exception.

@Andris0 my solution is now very close to the last one you have suggested here. I have just moved the base64 decoding and integer parsing into a separate function that uses the erlang functions `:base64.decode_to_string` and `:erlang.list_to_integer`, which is more performant.

I agree with you the performance increase here is neglectable -> I am not arguing against your points!


```elixir
  defp decode_user_id!(user_id) do
    try do
      user_id
      |> :base64.decode_to_string()
      |> :erlang.list_to_integer()

      :ok
    rescue
      _ -> raise(@invalid_token_error_message)
    end
  end
```

Andris0:
Well this looks good! Thanks @daskycodes for going above and beyond for what started out as a minor fix.
I do have a small nitpick though, if a `try` block spans across the whole function body, it can be simplified to this
```elixir
defp decode_user_id!(user_id) do
  user_id
  |> :base64.decode_to_string()
  |> :erlang.list_to_integer()

  :ok
rescue
  _ -> raise(@invalid_token_error_message)
end
```

daskycodes:
> Well this looks good! Thanks @daskycodes for going above and beyond for what started out as a minor fix. I do have a small nitpick though, if a `try` block spans across the whole function body, it can be simplified to this
> 
> ```elixir
> defp decode_user_id!(user_id) do
>   user_id
>   |> :base64.decode_to_string()
>   |> :erlang.list_to_integer()
> 
>   :ok
> rescue
>   _ -> raise(@invalid_token_error_message)
> end
> ```

That's really nice syntax. Thanks!

jchristgit (approved):
Thank you for this, this looks great 🙂

lib/nostrum/token.ex
```diff
@@ -0,0 +1,60 @@
...
+
+  defp decode_user_id!(user_id) do
+    _user_id =
+      user_id
+      |> :base64.decode_to_string()
+      |> :erlang.list_to_integer()
+
+    :ok
+  rescue
+    _ -> raise(@invalid_token_error_message)
```
daskycodes, line 58:
```suggestion
    exception -> reraise(RuntimeError, [message: @invalid_token_error_message, exception: exception], __STACKTRACE__)
```

@jchristgit I suppose this would make credo happy. Should we update the line to keep the old stacktrace in place?

jchristgit (approved):
Thanks for the credo fix 🙂

jchristgit:
Thanks!

PR #419, opened 2022-05-02
Update audio example to use application commands

BrandtHill:
Same thing needs to be done to other examples to be compatible with v10 gateway

### DIFF
--- a/examples/audio_player_example.ex
+++ b/examples/audio_player_example.ex
@@ -29,68 +29,70 @@ defmodule AudioPlayerConsumer do
   # Audio file will be fed directly to ffmpeg
   @nut_file_url "https://brandthill.com/files/nut.wav"
 
+  # Compile-time helper for defining Discord Application Command options
+  opt = fn type, name, desc, opts ->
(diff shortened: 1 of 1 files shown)


examples/audio_player_example.ex
```diff
@@ -29,68 +29,70 @@ defmodule AudioPlayerConsumer do
...
     |> Map.get(:channel_id)
   end
 
-  def do_not_ready_msg(msg) do
-    Api.create_message(msg.channel_id, "I need to be in a voice channel for that.")
+  # If you are running this example in an iex session where you manually call
+  # AudioPlayerSupervisor.start_link, you will have to call this function
+  # with your guild_id as the argument
+  def create_guild_commands(guild_id) do
+    Enum.each(@commands, fn {name, description, options} ->
```
jchristgit, line 72: Would it make sense to run `Enum.map` here, so in case the `create_guild_application_command` fails for whatever reason, it's visible in the return value?

jchristgit (approved):
Rock solid, as usual. Thanks!

PR #421, opened 2022-05-18
Fix a bug related to `Message.to_url`

Th3-M4jor:
Previously this function didn't work if any of the snowflakes were integers, now it will work as intended.

Yes I know `IO.iodata_to_binary()` is probably overkill here.


jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #425, opened 2022-06-11
fix dialyzer issues with webhooks

Th3-M4jor:
A few fields in the possible arguments here should be marked as optional, and file content for upload can successfully be `iodata`, not just a `binary`.


jchristgit (approved):
Yes. Yes. Great. Awesome. Thank you very much.

jchristgit:
Thanks!

PR #426, opened 2022-06-20
Resolve deprecation of `$` prefixing properties in identify payload, and add support for auto-mod

Th3-M4jor:
Just like it says on the tin. Some further testing would be appreciated as I'm unable to test this feature fully at present, hasn't rolled out to my test guild yet.

### DIFF
--- a/lib/nostrum/api.ex
+++ b/lib/nostrum/api.ex
@@ -50,6 +50,7 @@ defmodule Nostrum.Api do
 
   alias Nostrum.Struct.{
     ApplicationCommand,
+    AutoModerationRule,
     Channel,
     Embed,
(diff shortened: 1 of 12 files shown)


jchristgit:
I get a 404 trying to `get_guild_auto_moderation_rules` on a guild that has auto moderation rules enabled. Bot is in the guild. The URL in the constants also looks correct. Do we need to bump the API version or something for this to work?

jchristgit (approved):

jchristgit:
Well, the thought of just enabling an auto moderation rule did not quite arrive back then 🤨

jchristgit:
Thanks!

PR #429, opened 2022-07-23
support non padded base64 userid in the discord bot tokens

cedeno:
i was running into problems starting the bot and noticed my userid in the discord token was not a multiple of 4 and also not padded. using elixir modules for base64 seems to solve this by passing the padding: false option. not sure if there's a reason to stick with erlang modules though.


jb3:
yep, heard about this problem, once CI runs we're good to merge this, thanks!

jchristgit:
Thanks!

PR #430, opened 2022-07-26
Add information to Api Rest only section

inaniyants:
While using Nostrum with Rest Only approach, an error occurred in runtime of build (released with mix release):
```
[info] Application example exited: exited in: Example.Application.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (ArgumentError) The module Nostrum.Api.Ratelimiter was given as a child to a supervisor but it does not exist.
            (elixir 1.12.2) lib/supervisor.ex:631: Supervisor.init_child/1
```

The cause of it related to all `runtime: false` deps are not included into build, while building with `mix release`. 
I decided to add this info to docs as a time saver for devs, who will meet this non-obvious problem only in production.

### DIFF
--- a/docs/static/API.md
+++ b/docs/static/API.md
@@ -59,5 +59,8 @@ The ratelimiter at a high level works something like this:
 ## Rest Only
 If you only want to use the REST portion of the provided API, the only process
 needed is the ratelimiter. This can be manually started by calling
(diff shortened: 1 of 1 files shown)


jchristgit (approved):
Thank you! 🐱🐱🐱🐱🐱

PR #431, opened 2022-08-06
Update README.md

mtvillwock:
Emphasize message content intent warning

### Why
This change adds highlighting and italics to a section which, if missed, can create confusion on why a bot cannot reply to user input despite receiving the message.

E.g., the `msg` struct will have `''` in the `content` field, which fails the `case` statement.
Screen Shot 2022-08-06 at 11 51 34 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ config :nostrum,
   token: "666" # The token of your bot as a string
 ```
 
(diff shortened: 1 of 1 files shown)


jchristgit (approved):

jchristgit:
Thank you!

PR #438, opened 2022-09-07
Accept RTP packet without header

ushitora-anqou:
Discord sometimes sends us RTP packets without headers, which make nostrum crash with the following error:
```
20:52:08.330 [error] GenServer #PID<0.765.0> terminating
** (FunctionClauseError) no function clause matching in Nostrum.Voice.Opus.strip_rtp_ext/1
    (nostrum 0.6.1) lib/nostrum/voice/opus.ex:34: Nostrum.Voice.Opus.strip_rtp_ext(<<252, 255, 254>>)
    (nostrum 0.6.1) lib/nostrum/voice/session.ex:154: Nostrum.Voice.Session.handle_info/2
    (stdlib 4.0.1) gen_server.erl:1120: :gen_server.try_dispatch/4
    (stdlib 4.0.1) gen_server.erl:1197: :gen_server.handle_msg/6
    (stdlib 4.0.1) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
```
This patch fixes this problem. I believe discord.js processes RTP packets in the same way as this patch ([here](https://github.com/discordjs/discord.js/blob/d8e774138d5ab0333a767599e5470b4ea3f35ba5/packages/voice/src/receive/VoiceReceiver.ts#L120-L126)).

### DIFF
--- a/lib/nostrum/voice/opus.ex
+++ b/lib/nostrum/voice/opus.ex
@@ -42,6 +42,8 @@ defmodule Nostrum.Voice.Opus do
       ),
       do: rest
 
+  def strip_rtp_ext(packet), do: packet
+
   def parse_ogg(<<>>), do: []
 
   def parse_ogg(binary) do


BrandtHill:
Do you know if the data is valid opus when this occurs?

ushitora-anqou:
Sorry for the late reply. I tried to figure out what is happening here. Now I think that RTP packets without headers from Discord are valid.

I wrote [some code](https://github.com/ushitora-anqou/nostrum-rtp-packets-poc) to investigate this problem. This is forked from an voice example of Nostrum; I added a new slash command `/listen` to trigger `Voice.start_listen_async`. By following the README, you can launch two voice bots from the same code; one plays music, and the other listens to it and stores all received RTP packets in `/tmp/foobar`. I confirmed that all incoming packets could be decoded to wav by FFmpeg, and the result seemed correct.

Actually, I'm not so familiar with Discord's architecture, and not sure that I'm doing the right thing. Please feel free to point out if I've done something wrong.

BrandtHill:
Do you know if some or all of those packets you’ve recorded that were valid opus are ones that lacked the extension header? My thought is these packets that don’t have one-byte header extensions are either

- packets with two-byte header extensions
- packets with no header extensions
- control type packet we need to skip akin to rtcp packets

I haven’t encountered any of these packets. I suspect it’s not a control packet if the RTP header elements like rtp sequence, time, ssrc look normal. Maybe this happens strictly when receiving packets from another bot. 

If you know that you’re receiving multiple extensions-less packets that are good audio then I would say this is a good solution, and maybe later we can support two-byte header extensions in case discord starts sending those in the future.

ushitora-anqou:
> Do you know if some or all of those packets you’ve recorded that were valid opus are ones that lacked the extension header?

As far as I tested by using the above code, any received packet didn't have extension headers. You can confirm this by modifying [here](https://github.com/Kraigie/nostrum/pull/438/files#diff-3cfbd2cbdcbcc327943d03d3eacf672c97a47716556a64b4a3b0434e57f961cdR43) to `do: raise("debug")` and checking that no exception is raised.

> My thought is these packets that don’t have one-byte header extensions are either
> 
> - packets with two-byte header extensions
> - packets with no header extensions
> - control type packet we need to skip akin to rtcp packets

An incoming packet is something like this (on `:VOICE_INCOMING_PACKET` event):

```
<<252, 203, 162, 207, 150, 7, 57, 246, 136, 30, 30, 159, 172, 137, 194, 244, 210, 16, 78, 165, 133, 158, 143, 203, 152, 73, 17, 249, 48, 14, 2, 31, 242, 50, 214, 16, 134, 106, 53, 67, 96, 85, 28, 49, 78, 71, 74, 83, 21, 146, ...>>
```

This packet seems to be neither a two-byte header nor an RTCP packet. So, I think it's a packet with no extensions.

> Maybe this happens strictly when receiving packets from another bot.

I agree. In my environment, I've not yet encountered this problem when no voice bots are speaking.

jchristgit (approved):

jchristgit:
Thank you!

PR #441, opened 2022-11-18
Support role icons

jswny:
Tested this with a local copy and it seems to properly populate the `:icon` field into the struct from the API response.

### DIFF
--- a/lib/nostrum/struct/guild/role.ex
+++ b/lib/nostrum/struct/guild/role.ex
@@ -61,7 +61,7 @@ defmodule Nostrum.Struct.Guild.Role do
   @type mentionable :: boolean
 
   @typedoc "The hash of the role icon"
-  @type icon :: String.t()
+  @type icon :: String.t() | nil
 
   @type t :: %__MODULE__{
           id: id,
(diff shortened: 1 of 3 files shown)


lib/nostrum/struct/guild/role.ex
```diff
@@ -59,6 +60,9 @@ defmodule Nostrum.Struct.Guild.Role do
   @typedoc "Whether the role is mentionable"
   @type mentionable :: boolean
 
+  @typedoc "The hash of the role icon"
+  @type icon :: String.t()
```
Th3-M4jor, line 64:
```suggestion
  @type icon :: String.t() | nil
```
One small thing since this field is nullable

jswny: Nice catch!

jchristgit (approved):
Thanks!

bdanklin:
Absolutely fantastic work. This is the type of cutting edge contribution I needed to see to reignite my interest in the project. Lets keep this ball rolling. @jchristgit hit the gas on the workflows!!

jchristgit:
Thanks @bdanklin for your motivational words. I am excited to announce that as part of the Craig Dazey Nostrum Contributor Fund 2023, we are able to position three open source contributors to the nostrum library in Low Moon Orbit for experimenting with nostrum's performance in a low-gravity environment. As maintainers are excluded from this package, we are now seeking contributions to the project, and a sponsor providing us with a rocket and station assembly.

PR #442, opened 2022-11-20
Support emoji role icons

jswny:
Apparently the image icon (from #441) is separate from the emoji icon, so we need this too.

### DIFF
--- a/lib/nostrum/struct/guild/role.ex
+++ b/lib/nostrum/struct/guild/role.ex
@@ -29,7 +29,8 @@ defmodule Nostrum.Struct.Guild.Role do
     :permissions,
     :managed,
     :mentionable,
-    :icon
+    :icon,
+    :unicode_emoji
   ]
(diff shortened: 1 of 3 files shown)


lib/nostrum/struct/guild/role.ex
```diff
@@ -64,6 +65,10 @@ defmodule Nostrum.Struct.Guild.Role do
   @typedoc since: "0.7.0"
   @type icon :: String.t() | nil
 
+  @typedoc "The standard unicode character emoji icon for the role"
+  @typedoc since: "0.7.0"
```
jswny, line 69: Not sure if I should be adding this in PR's so let me know (copied from above)

jchristgit: This is good, thank you

jchristgit (approved):
Thanks!

PR #447, opened 2023-01-06
Minor voice refactors and improve get reactions api call

BrandtHill:
I had noticed my bot hell looping trying to reconnect after getting a voice ws error 4006 so I'm leaving less to chance when sequencing events for reconnecting. Using cast in some places to reduce blocking processes. The get reactions api changes are unrelated because I forgot I had those unstaged changes in my branch.

### DIFF
--- a/lib/nostrum/shard/dispatch.ex
+++ b/lib/nostrum/shard/dispatch.ex
@@ -357,7 +357,7 @@ defmodule Nostrum.Shard.Dispatch do
 
           # Already in this channel:
           true ->
-            Voice.update_voice(p.guild_id)
+            :noop
         end
       else
         # Leaving Channel:
--- a/lib/nostrum/shard/supervisor.ex
+++ b/lib/nostrum/shard/supervisor.ex
@@ -56,16 +56,10 @@ defmodule Nostrum.Shard.Supervisor do
 
   def update_voice_state(guild_id, channel_id, self_mute, self_deaf) do
     case GuildShard.get_shard(guild_id) do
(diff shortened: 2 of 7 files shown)


lib/nostrum/api.ex
```diff
@@ -557,25 +557,33 @@ defmodule Nostrum.Api do
...
+      Constants.channel_reactions_get(channel_id, message_id, emoji_api_name),
+      "",
+      params
+    )
     |> handle_request_with_decode({:list, {:struct, User}})
   end
 
   @doc ~S"""
-  Same as `get_reactions/3`, but raises `Nostrum.Error.ApiError` in case of failure.
+  Same as `get_reactions/4`, but raises `Nostrum.Error.ApiError` in case of failure.
```
jchristgit, line 582: This should be
```suggestion
  Same as `get_reactions/3-4`, but raises `Nostrum.Error.ApiError` in case of failure.
```
right?

jchristgit (commented):
Sorry for the delay on this. This looks good. Will look into fixing CI.

lib/nostrum/api.ex
skwerlman: no, that breaks linking to the other function

BrandtHill: image
Legally the functions are arity 4

jchristgit:
LEGEND :fire::fire::fire::fire::fire: Keep it up man!!!!!!!! Thanks for the pull request!!!!!!!!

PR #448, opened 2023-01-15
Retrieve original message

Awlexus:
Adds the functions `get_webhook_message/2` and `get_original_interaction_response/1` to the API module. 

### Remarks
* `get_webhook_message/2` has a parameter `thread_id`, but I don't use webhooks, so I wouldn't be able to test them properly. 
* The [docs for `get_original_interaction_response/1`](https://discord.com/developers/docs/interactions/receiving-and-responding#get-original-interaction-response) mention that it functions the same as `get_webhook_message/2`, but I don't see how `thread_id` would translate to interactions, given that the only documented value is "@original"


Awlexus:
Hi, the changes have been made and the CI has a nice green color now

PR #449, opened 2023-01-15
Fix button status

Awlexus:
Replaces the functions `enable/1` and `disable/1` from `Nostrum.Struct.Component.Button` with the function `enable/2` and fixes a crash in `toggle/1`

### DIFF
--- a/lib/nostrum/struct/component/button.ex
+++ b/lib/nostrum/struct/component/button.ex
@@ -105,31 +105,18 @@ defmodule Nostrum.Struct.Component.Button do
   """
   def toggle(%{type: 2, disabled: button_state} = button) do
     button
-    |> struct({:disabled, !button_state})
+    |> struct([{:disabled, !button_state}])
   end
 
   def toggle(_), do: buttons_only()
 
   @doc """
-  Disables the button.
+  Enables the button when `enabled` is true. Disables it otherwise.
   """
-
-  def disable(%{type: 2} = button) do
+  def enable(%{type: 2} = button, enabled) do
     button
-    |> update({:disabled, true})
+    |> update([{:disabled, !enabled}])
   end
-
-  def disable(_), do: buttons_only()
-
-  @doc """
-  Enables the button.
-  """
-  def enable(%{type: 2} = button) do
-    button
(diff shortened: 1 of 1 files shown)


jchristgit:
I'm 50/50 on this.

If we remove the `disable` function, then I think we should update our functions to match the actual name of the attribute we're setting: `disabled(button, false)` or `disabled(button, true)`. Alternatively, we should leave `disable` in, it can just call the new `enable` function with `false` passed, but then I do think it gets kind of hairy again. So I do think that having two functions for enabling and disabling was a cleaner way. What do you think?

Awlexus:
I think I should have explained my reasoning too. I had almost forgotten myself. When I was working with a button I often found myself forced to do something like this, which could be avoided by passing the boolean directly.  
```elixir
button = 
  if cond do
    Button.disable(button)
  else
    Button.enable(button)
  end
```

But now that you mention it, calling it `disable/2` instead makes more sense. I thought if nobody noticed that both functions were broken for several months, we might as well replace them

jchristgit:
I think the `set_disabled` version we had on Discord with `enable`, `disable` and `toggle` being aliases of it was the best solution to go with here. What do you thinkß

bdanklin:
> 

yes hello, bdanklin here the resident native english speaker (non usa) 

'set disabled' sounds like you are making the button disabled. not that you are controlling the 'disabled' field. This is obviously discord devs intelligence shining through but we shouldn't propagate it further.

bdanklin:
If the current situation warrants a change. Then you should accept the same request on any field on any component. (or they should also be included in this pr)

This suggestion by my esteemed colleague is clearly the ideal solution.

https://discord.com/channels/81384788765712384/381889573426429952/1088582496242892883

jchristgit:
To be honest, instead of yak shaving for another while here, I think the way to go is to merge this as-is. Just one request, can you rebase from or merge master to get the CI fixes (and ensure CI passes)?

Awlexus:
Fair enough. I'll also follow at least one of the first suggestions and rename enable, because it matches the internal name of the field

jchristgit (approved):
Thanks!

PR #454, opened 2023-01-31
Managed application commands

portasynthinca3:
This pull request is intended to reduce boilerplate in projects that use application commands. Instead of requiring library users to implement a `Consumer` and manually handle `INTERACTION_CREATE` events, this pull requests enables them to define a separate module per command containing the command specification and its handling function.

As a bonus, it removes magic numbers in command specs, instead opting in for atoms, and merges some spec fields into one field (for example, `name` and `name_localizations` are represented as a tuple and `value_min` and `value_max` as a range)

### DIFF
--- a/docs/static/Application Commands.md
+++ b/docs/static/Application Commands.md
@@ -18,62 +18,78 @@ documentation](https://discord.com/developers/docs/interactions/application-comm
 ## Getting started
 
 Discord differentiates between **global** and **guild-specific** slash
-commands. Global commands will be distributed across all guilds that your bot
-is in within an hour. Guild-specific commands slash commands will be available
-instantly, which is why we will use guild-specific commands for testing.
+commands. Global commands used to take up to an hour to update across all
+guilds, but that is not the case anymore.
 
 We will create a command that will allow the user to assign or remove a role of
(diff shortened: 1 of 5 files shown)


jchristgit (changes requested):
Thank you for the PR, and sorry for the late review!

This looks mostly good. It suffers from one problem: when an event is received, it's passed to a random consumer, due to the way we're using GenStage at the moment.

I think a better solution than to create an internal consumer would be to simply instruct users to call `handle_event` or maybe just a function `handle_interaction` that is called directly with the received interaction.

portasynthinca3:
Hi! Sorry for not replying earlier. I have a couple of design decisions that i need but cannot make:

  1. Does this belong in the main library? When I originally presented my PR in the unofficial Discord API server, people there suggested that I instead make it into a separate library, so I started working on that. It even included a powerful but not-fully-complete localization engine. I stopped working on it because I kind of stopped programming for a while, but I will for certain return to it later.
  2. While I do agree that users passing `:INTERACTION_CREATE` events to this component of the main library or my separate library, whatever shape my idea ends up taking, is more "Elixir-ian" and more true to FP dogmas, I also think that automatic handling of the events will result in less boilerplate on the user side

So, what do you think?

PR #455, opened 2023-02-19
update the version-type to strict and specify matrix versions more nicely

bdanklin:
Might fix the CI if we are lucky, cant test it on a fork though.

This could use a lot more work but hopefully this should band-aid it for now until I can steal the one from work without anyone noticing.

🦆

### DIFF
--- a/.github/workflows/test_and_lint.yml
+++ b/.github/workflows/test_and_lint.yml
@@ -7,24 +7,27 @@ on:
   pull_request:
 
 jobs:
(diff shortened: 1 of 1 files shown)


bdanklin:
Nevermind, guess its beyond fucked.

jchristgit:
Sorry I saw this a bit late. CI works now, adding the new versions would be nice, I'm working on it :)

The problem was that the `ubuntu-latest` image does not support the oldest versions anymore

PR #461, opened 2023-03-31
Fix button API

Awlexus:
Because the discussions on discord have staled, I had time to rethink the Button API and I've come to the conclusion that not providing a Button API would leave everyone equally unhappy and is therefore the only possible way to move forward


jchristgit (approved):
Thanks!

Awlexus:
Putting an end to this April fools' joke while nobody is looking 👀

PR #481, opened 2023-05-07
Namespacing

jchristgit:


lib/nostrum/voice/supervisor.ex
```diff
@@ -9,12 +9,12 @@ defmodule Nostrum.Voice.Supervisor do
...
 
   def start_link(_opts) do
-    Supervisor.start_link(__MODULE__, [], name: VoiceSupervisor)
+    Supervisor.start_link(__MODULE__, [], name: __MODULE__)
   end
 
   def init(_opts) do
     children = [
-      {DynamicSupervisor, strategy: :one_for_one, name: VoiceSessionSupervisor},
+      {DynamicSupervisor, strategy: :one_for_one, name: Nostrum.Voice.SessionSupervisor},
```
BrandtHill, line 17: Need to change in `create_session/1` too

jchristgit: Thanks! Fixed in 468b9385

jchristgit:
Thanks!

PR #484, opened 2023-05-08
Fix crash when trying to connect when session pid isn't alive

BrandtHill:
Fix bug where when session_pid was dead, joining a voice channel would make a gen server call to the dead session and would end up crashing a cache.
This situation shouldn't occur since the session processes are supervised and they should be alive for the duration that the bot is connected to a voice channel, but I've seen it occur a couple times and it ended up crashing the bot. I suspect it may be due to gun2.0 or due to other recent async voice reconnect changes, but this change covers that case now.
Simplify `on_channel_join_change` as voice events always come in in guaranteed order so calling the session gen server isn't necessary so we can further reduce number of gen server calls.

### DIFF
--- a/lib/nostrum/shard/dispatch.ex
+++ b/lib/nostrum/shard/dispatch.ex
@@ -358,6 +358,18 @@ defmodule Nostrum.Shard.Dispatch do
           voice.channel_id != p.channel_id and is_pid(voice.session_pid) ->
             Voice.on_channel_join_change(p, voice)
 
+          # Already in this channel but connection died:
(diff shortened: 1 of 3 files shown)


BrandtHill:
Nothing special in the logs that I've found. 
When you attempt to switch channels, the call `Session.get_ws_state(voice.session_pid)` would fail and cause one of the caches to Cache GenStage stage to fail catastrophically. This fixes that major error.
Now I'd like to rebase against master and possibly improve some of the edge case handling.

PR #487, opened 2023-05-15
Pluggable state everywhere

jchristgit:
Make all of nostrum's state pluggable. This includes internal state, such as ratelimiter buckets, guild shard mappings, and more.


jchristgit (commented):
Great PR!

PR #496, opened 2023-05-23
Requeue up to 50 times

jchristgit:
Previously we used to requeue infinitely long. Let's use 50 instead, which should prevent any normal usage from running into the maximum retry exceeded error.


jchristgit:
Thanks!

PR #498, opened 2023-05-27
Supply appup for hot code upgrades

jchristgit:


jchristgit:
Current Elixir version on Debian stable is 1.10. So Nostrum already does not work there unfortunately. On time for Debian Bookworm we can thus lift the version requirement by one Elixir minor release.

jchristgit:
Thanks!

PR #500, opened 2023-05-29
add new channel field to interactions

Th3-M4jor:
Like it says on the tin. New `:channel` field to replace `channel_id` which they say will eventually be deprecated and removed.


jchristgit (approved):
Thank you!

PR #501, opened 2023-05-29
Convert ratelimiter into an async state machine

jchristgit:
This commit reimplements the existing ratelimiter GenServer using a state machine via `:gen_statem`. This brings us the following benefits:

- Requests are pipelined as soon as we know that the ratelimit allows us to. For any endpoint that returns a body, this happens as soon as we have retrieved the request headers, even before we have started parsing the body.

- No more garbage collection of ratelimiter buckets is necessary. They are automatically removed as soon as any pending retry-after headers have expired.

- Clients no longer need to sleep out ratelimits. The server does not sleep anything out either. In fact, the server is a fully-asynchronous hellbeast that just pipelines in the next request as soon as it sees fit.

- To be more specific, queueing requests does not require anybody to look up anything aside from the bucket, because the bucket's queue will effectively handle all queueing as soon as possible.

- The special handling of the global ratelimit has mostly been removed, individual requests don't need to look that up at all. The global ratelimit is implemented as a special state in the state machine that postpones anything you ask it to do until the global ratelimit has expired again.

- The impure ratelimit bucket store using four different callbacks to user-defined data stores has been removed and replaced with what's effectively `:erlang.send_timer`, 30 years of sane language design, and a small bit of magic powder.

- The previous "ask gun to reconnect forever and don't do anything on down or up events" has been removed, so we can properly tell clients when their request is done.

- We can use `:sys.trace(Nostrum.Api.Ratelimiter, true)` or other functions from `:sys` to get really valuable insights into what the ratelimiter is doing.

This is not just any ratelimiter state machine, this is _The_ State Machine. This is The Way going forward.

As for the drawbacks: well, distributing the ratelimiter is a bit more difficult now. And if something very bad happens to the ratelimiter while a request is in flight, that request is gonna have a real bad time. Previously we had the request sleeps on client-side so that was mostly fine. I also have not implemented any back-off logging (yet?).

As for alternatives - I have not really found an alternative which doesn't involve responding to a specific process that can deal with ratelimit responses _or_ using some external data storage for it. And depending on how you do it, that brings you synchronisation problems.

> of course you throw away the first few implementations, you didn't
> understand the problem when you wrote those!


lib/nostrum/api/ratelimiter.ex
```diff
@@ -9,188 +9,374 @@ defmodule Nostrum.Api.Ratelimiter do
...
+        {:keep_state, %{data | outstanding: :maps.put(bucket, queue, outstanding)},
+         [{{:timeout, bucket}, 0, :next}]}
 
-      # We've had enough. Bail.
-      {:error, {:retry_after, _time}} when attempts_remaining == 0 ->
-        {:error, :max_retry_attempts_exceeded}
+      queue ->
+        {:keep_state, %{data | outstanding: %{outstanding | bucket => :queue.in(entry, queue)}}}
+    end
+  end
```
Th3-M4jor, line 239: Why not do something like this:
```elixir
case outstanding do
  %{^bucket => queue} ->
    # found bucket
  _ ->
    # no bucket found
  end
  ```

File: lib/nostrum/api/ratelimiter.ex [hunk 2]
```diff
@@ -9,188 +9,374 @@ defmodule Nostrum.Api.Ratelimiter do
...
+          Base.request(
+            conn,
+            request.method,
+            request.route,
+            request.body,
+            request.headers,
+            request.params
+          )
+
+        running_with_this = :maps.put(stream, {bucket, request, from}, running)
```
Th3-M4jor, line 259: Could you convert the calls to `:maps` into calls to `Map`? Just to avoid confusion since they usually have different argument ordering?

lib/nostrum/api/ratelimiter.ex
jchristgit: This is a good idea. I updated this.

lib/nostrum/api/ratelimiter.ex
jchristgit: This is also a good idea. I have also updated this.

jchristgit:
Made it work. Next will make it pretty:

- [ ] add proper types and document them
- [ ] add functions for building common structures
- [ ] add a bit more documentation
- [ ] look into some variable renaming

jchristgit:
Thanks!

PR #502, opened 2023-05-30
Add `@behaviour`, `@impl` and default `handle_event` to `Nostrum.Consumer`

bcksl:
A couple of additions to `Nostrum.Consumer`:

- Use `@impl` declarations where appropriate.
- Add a `@behaviour Nostrum.Consumer` declaration to `__using__`.
- Add a `__before_compile__` hook to inject a default `:noop` implementation of `handle_event`.

### DIFF
--- a/lib/nostrum/consumer.ex
+++ b/lib/nostrum/consumer.ex
@@ -374,6 +374,8 @@ defmodule Nostrum.Consumer do
     quote location: :keep do
       use GenServer
 
+      @behaviour Nostrum.Consumer
+
       def child_spec(opts) do
         %{
           id: __MODULE__,
--- a/lib/nostrum/consumer.ex
+++ b/lib/nostrum/consumer.ex
@@ -21,7 +21,7 @@ defmodule Nostrum.Consumer do
 
       {:noreply, state}
     end
-  ``` 
+  ```
 
   ## Example consumer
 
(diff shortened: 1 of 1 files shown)


lib/nostrum/consumer.ex
```diff
@@ -406,7 +413,14 @@ defmodule Nostrum.Consumer do
...
 
-      defoverridable child_spec: 1, start_link: 1, init: 1, handle_continue: 2, handle_info: 2
+      defoverridable GenServer
+    end
+  end
+
+  defmacro __before_compile__(_env) do
+    quote location: :keep do
+      @impl Nostrum.Consumer
+      def handle_event(_), do: :noop
```
jchristgit, line 423: Is there any guarantee on where this will be inserted into the module?

If it would be inserted early then no events would be received in the consumer.

jchristgit (commented):
Thank you very much for the pull request!

I have a minor question, apart from that it looks great!

lib/nostrum/consumer.ex
bdanklin: > Is there any guarantee on where this will be inserted into the module?

Yes.

@before_compile is inserted at the end of the module.

jchristgit: Thank you for the explanation 👍

PR #503, opened 2023-06-01
Convert shard sessions into state machines

jchristgit:
This fixes the previous implementation's problem of retrying connections indefinitely without any proper handling of errors. Ratelimiting will be added separately.

### DIFF
--- a/lib/nostrum/shard/event.ex
+++ b/lib/nostrum/shard/event.ex
@@ -20,7 +20,9 @@ defmodule Nostrum.Shard.Event do
 
     if payload.t == :READY do
       Logger.info("READY")
-      {%{state | session: payload.d.session_id}, []}
+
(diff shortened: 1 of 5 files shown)


lib/nostrum/shard/event.ex
```diff
@@ -20,42 +20,41 @@ defmodule Nostrum.Shard.Event do
 
     if payload.t == :READY do
       Logger.info("READY")
-      %{state | session: payload.d.session_id}
+      {%{state | session: payload.d.session_id}, []}
```
Th3-M4jor, line 23: Since ready events now include the URL Discord wants us to use for [future resumes](https://discord.com/developers/docs/topics/gateway-events#ready-ready-event-fields), should we add that to the shard data here?

jchristgit:
This is a great idea. I have implemented it.

jchristgit:
Thanks!

PR #504, opened 2023-06-01
Add state machine report translator

jchristgit:
Elixir will, by default, not output `:gen_statem` errors, which is ... suboptimal. We add a new logger translator to deal with these errors and report them.


jchristgit:
Thanks!

PR #507, opened 2023-06-09
Fix Util.get_all_shard_latencies

Th3-M4jor:
`Session.get_ws_state` now returns the tuple `{:connected, %WSState{}}`, it was causing KeyNotFound errors to be raised.

### DIFF
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -287,7 +287,7 @@ defmodule Nostrum.Util do
     |> Enum.map(fn {_id, pid, _type, _modules} -> Supervisor.which_children(pid) end)
     |> List.flatten()
     |> Enum.map(fn {_id, pid, _type, _modules} -> Session.get_ws_state(pid) end)
-    |> Enum.reduce(%{}, fn s, m -> Map.put(m, s.shard_num, get_shard_latency(s)) end)
+    |> Enum.reduce(%{}, fn {_, s}, m -> Map.put(m, s.shard_num, get_shard_latency(s)) end)
   end
 
   @doc """


lib/nostrum/util.ex
```diff
@@ -287,7 +287,16 @@ defmodule Nostrum.Util do
...
+    |> Enum.reduce(%{}, fn s, m ->
+      # only add the latency if the shard is connected
+      case s do
+        {:connected, %Nostrum.Struct.WSState{shard_num: shard_num} = state} ->
+          Map.put(m, shard_num, get_shard_latency(state))
+
+        _ ->
+          m
+      end
+    end)
```
BrandtHill, line 299: I might prefer to keep the behavior the same where all the shard keys are present in the map but `nil` if it's not connected. Something like
```elixir
|> Enum.reduce(%{}, fn
  {:connected, %WSState{shard_num: shard_num} = state}, map ->
    Map.put(map, shard_num, get_shard_latency(state))

  {_non_connected, %WSState{shard_num: shard_num}}, map ->
    Map.put(map, shard_num, nil)
end)
```
Or more simply 
```elixir
|> Enum.reduce(%{}, fn {_, s}, m -> Map.put(m, s.shard_num, get_shard_latency(s)) end)
```
since `get_shard_latency/1` will return `nil` if either of the heartbeat fields are `nil`

Th3-M4jor: Updated

BrandtHill (approved):

jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #508, opened 2023-06-11
Enable -Wunmatched_return in dialyzer

jchristgit:


jchristgit:
@BrandtHill There's a lot of voice-related warnings from Dialyzer here, where we are ignoring potential ok / error results.

Do you think we should ignore the results as we do presently or match against them?

```elixir
lib/nostrum/shard/dispatch.ex:346:unmatched_return
The expression produces a value of type:

nil | :ok | {:error, <<_::248, _::size(32)>>}

but this value is unmatched.

________________________________________________________________________________
lib/nostrum/voice.ex:556:unmatched_return
The expression produces a value of type:

{_, boolean()}

but this value is unmatched.

________________________________________________________________________________
lib/nostrum/voice.ex:576:unmatched_return
The expression produces a value of type:

:ignore | {:error, _} | {:ok, pid()} | {:ok, pid(), _}

but this value is unmatched.

________________________________________________________________________________
lib/nostrum/voice.ex:755:unmatched_return
The expression produces a value of type:

:ignore | nil | {:error, _} | {:ok, pid()} | {:ok, pid(), _}

but this value is unmatched.

________________________________________________________________________________
lib/nostrum/voice/audio.ex:127:unmatched_return
The expression produces a value of type:

{:error, _} | {:ok, :cancel}

but this value is unmatched.

________________________________________________________________________________
lib/nostrum/voice/audio.ex:139:unmatched_return
The expression produces a value of type:

:ok | {:error, atom()}

but this value is unmatched.

________________________________________________________________________________
lib/nostrum/voice/ports.ex:43:unmatched_return
The expression produces a value of type:

nil | {:ok, pid()}

but this value is unmatched.

________________________________________________________________________________
lib/nostrum/voice/session.ex:124:unmatched_return
The expression produces a value of type:

{:error, _} | {:ok, :cancel}

but this value is unmatched.

________________________________________________________________________________
done (warnings were emitted)
```

BrandtHill:
I've added matching to most of these, but the others should be ignored

jchristgit:
Thank you!

jb3 (approved):
Looks solid, and works locally, so no complaints here.

jb3:
Thanks!

PR #510, opened 2023-06-12
Fix reconnect requests

Th3-M4jor:
fix reconnect requests from discord to make it so we resume properly and without an infinite loop.

Was unable to test fully, do not merge until someone has had a chance to.

### DIFF
--- a/lib/nostrum/shard/session.ex
+++ b/lib/nostrum/shard/session.ex
@@ -151,23 +151,14 @@ defmodule Nostrum.Shard.Session do
 
   # If we've been here before, we want to use the resume gateway URL to connect
   # instead of the regular gateway URL.
-  def connecting_http(:enter, from, %{resume_gateway: resume_gateway} = data)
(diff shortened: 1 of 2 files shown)


lib/nostrum/shard/session.ex
```diff
@@ -145,16 +145,29 @@ defmodule Nostrum.Shard.Session do
...
+    # strip off the wss:// prefix if it's there
+    # since :gun doesn't know how to handle it
+    resume_url =
+      case resume_gateway do
+        <<"wss://", gateway_url::binary>> ->
+          gateway_url
+
+        gateway_url ->
+          gateway_url
+      end
```
jchristgit, line 167: Can't we just do

`%{resume_gateway: "wss://" <> resume_gateway}`

in the function head?

jchristgit (approved):
Thanks!

jchristgit:
Thanks!

PR #516, opened 2023-06-19
Unsafe & Async Helper Proof of Concept

bdanklin:
This shows a proof of concept for auto generating bang functions. A similar method can be applied to trivially generate async functions using Task.

```elixir
@async
@unsafe
def create_reaction(arg1, arg2, arg3) do
```
results in the following function definitions
```elixir
def create_reaction(arg1, arg2, arg3), do: ...
def create_reaction!(arg1, arg2, arg3), do: ...
def create_reaction_async(arg1, arg2, arg3), do: ...
```

image


jchristgit:
With async I think we need to go the reverse route.

The default needs to be `_async` and the sync functions then call that and await on the response.

PR #520, opened 2023-07-23
Handle nil member count in guild caches

Th3-M4jor:
Makes it so that member_count being `nil` in the guild cache won't cause a crash on a member leaving/joining.

Ideally this shouldn't happen but 🤷‍♂️

Also, ignore the commit before this, I was debugging a separate issue and removed all changes from it.

### DIFF
--- a/lib/nostrum/api/ratelimiter.ex
+++ b/lib/nostrum/api/ratelimiter.ex
@@ -971,7 +971,7 @@ defmodule Nostrum.Api.Ratelimiter do
   # defp parse_headers({:error, _reason} = result), do: result
 
   # credo:disable-for-next-line
-  defp parse_headers({:ok, {status, headers, body}}) do
+  defp parse_headers({:ok, {_status, headers, _body}}) do
     limit_scope = header_value(headers, "x-ratelimit-scope")
     remaining = header_value(headers, "x-ratelimit-remaining")
     remaining = unless is_nil(remaining), do: String.to_integer(remaining)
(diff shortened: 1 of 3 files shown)


Th3-M4jor:
@Leastrio has came across it
<details>
<summary>here's the relevant log data</summary>

```
Reason for termination = exit:{aborted,
                                 {badarith,
                                   [{erlang,'-',
                                     [nil,1],
                                     [{error_info,
                                      #{module => erl_erts_errors}}]},
                                   {'Elixir.Nostrum.Cache.GuildCache.Mnesia',
                                     '-member_count_down/1-fun-0-',1,
                                     [{file,
                                    "lib/nostrum/cache/guild_cache/mnesia.ex"},
                                     {line,238}]},
                                    {'Elixir.Nostrum.Cache.GuildCache.Mnesia',
                                     '-update_guild/2-fun-0-',2,
                                    [{file,
                                     "lib/nostrum/cache/guild_cache/mnesia.ex"},
                                      {line,250}]},
                                   {mnesia_tm,apply_fun,3,
                                     [{file,"mnesia_tm.erl"},{line,889}]},
                                    {mnesia_tm,execute_transaction,5,
                                   [{file,"mnesia_tm.erl"},{line,865}]},
                                   {mnesia,wrap_trans,6,
                                     [{file,"mnesia.erl"},{line,502}]},
                                   {'Elixir.Nostrum.Cache.GuildCache.Mnesia',
                                     member_count_down,1,
                                     [{file,
                                      "lib/nostrum/cache/guild_cache/mnesia.ex"},
                                     {line,237}]},
                                    {'Elixir.Nostrum.Shard.Dispatch',
                                    handle_event,3,
                                     [{file,"lib/nostrum/shard/dispatch.ex"},
                                    {line,213}]}]}}
```

</details>
The root of the problem is probably actually the GuildUpdate event now that I think about it, since we're casting it to a Guild struct, and that doesn't include the `member_count` field.
I'll change that to be a merge instead of a replace maybe.

Th3-M4jor:
If you want to cherry pick the last commit that's fine by me.

Th3-M4jor:
Was the only commit actually needed I think.

Issue #524, opened 2023-10-18
Connection Websocket upgrade failed


tidusIO:
Since switching to the latest master I get following repeating error when starting my application. This happens in Google Cloud and my local macOS system as well.

```[debug] Establishing initial connection
[info] WAITING 128 BEFORE NEXT SHARD CONNECT
[debug] Upgrading connection to websocket
[debug] Shard connection up, waiting for consumers to boot
[error] ** State machine <0.1213.0> terminating
** Last event = {info,{gun_upgrade,<0.1214.0>,
                                   #Ref<0.2030176606.2015625217.39321>,
                                   [<<"websocket">>],
                                   [{<<"date">>,
                                     <<"Wed, 18 Oct 2023 17:26:32 GMT">>},
                                    {<<"connection">>,<<"upgrade">>},
                                    {<<"sec-websocket-accept">>,
                                     <<"XTX4ZupV6bXSO3Txkhavuwd22D4=">>},
                                    {<<"upgrade">>,<<"websocket">>},
                                    {<<"cf-cache-status">>,<<"DYNAMIC">>},
                                    {<<"report-to">>,
                                     <<"{\"endpoints\":[{\"url\":\"https:\\/\\/a.nel.cloudflare.com\\/report\\/v3?s=RQzrHe90NXM3fNbUxzSRVnVs9hWpdOGipWPR%2FBbWQL%2BA0v267FBgfcViDC85WxoAzwtevd6xku7G2UhJJYhbqBLKsfvpW%2FI53iGlAy8Ujy1LUnaEV0YPiRmiJmv9nk4IpK8n4w%3D%3D\"}],\"group\":\"cf-nel\",\"max_age\":604800}">>},
                                    {<<"nel">>,
                                     <<"{\"success_fraction\":0,\"report_to\":\"cf-nel\",\"max_age\":604800}">>},
                                    {<<"strict-transport-security">>,
                                     <<"max-age=31536000; includeSubDomains; preload">>},
                                    {<<"x-content-type-options">>,
                                     <<"nosniff">>},
                                    {<<"server">>,<<"cloudflare">>},
                                    {<<"cf-ray">>,
                                     <<"81827ca38dca9325-ICN">>}]}}
** When server state  = {connecting_ws,
                            #{stream => #Ref<0.2030176606.2015625217.39321>,
                              seq => nil,
                              '__struct__' => 'Elixir.Nostrum.Struct.WSState',
                              session => nil,conn => <0.1214.0>,
[...]

PR #525, opened 2023-10-19
Add error message for lacking consumer startup

jchristgit:
Instead of relying on the user to interpret the existing debug logs and badmatch error, send an explicit error message when no consumer could be found that we can use in the shard connection. See #524.


jchristgit (commented):
Great PR!

jchristgit:
Thanks!

PR #526, opened 2023-10-24
bump minimum elixir version and test matrix

Th3-M4jor:
Since we use `:pg.monitor/2` within Nostrum and it was introduced in OTP 25.1, this bumps the Elixir version mentioned in our `mix.exs` to the earliest version of Elixir which officially supports OTP 25. Also updates CI to include the latest versions of OTP and Elixir while dropping unsupported versions.


jchristgit (approved):
Thanks!

PR #528, opened 2023-11-01
Check OTP version on startup and update deprecated Logger.warn/2 calls

BrandtHill:


jchristgit (approved):
Would be nice if we could set a required OTP version in `mix.exs`. But this is nice as well, because you made it. Happy Wednesday my guy. Hope you're having a good one.

PR #531, opened 2023-11-16
[Experimental] support multiple gun connections in the rate limiter

Th3-M4jor:
Experimental PR to see if we can support multiple gun connections in the ratelimiter at once without too much of a rewrite while we wait for Discord to enable http2 again. Needs extensive testing before it should be considered ready to merge, I've only done a few smoke tests so far.


jb3:
Closing since HTTP2 is re-enabled now and this shouldn't be necessary anymore.

If we need to we can re-open/redo this in future, but let's hope for no more crit vulns in HTTP2.

jb3:
Thanks!

PR #541, opened 2024-04-19
Document StateMachineTranslator as internal

jchristgit:


jb3 (approved):
Needs a format but otherwise let's get this bread.

jb3:
Thanks!

PR #554, opened 2024-04-19
Move ChannelCache to DMChannelCache

### LINKED ISSUE #537: ChannelCache documentation enhancements
Whilst it is mentioned in the documentation that the `ChannelCache` only stores `DMChannels`, I think we should make this more clear to the user, my proposals are:

- We rename this structure to `DMChannelCache`, as it is not a generic channel cache
- We add a warning to the docs instead of just having this feature in the blurb

This is based off the fact that several users in the nostrum channel have been confused by this behavior, including myself at first. We did change this behavior and as such I think we should make it as clear as possible that this is how it works. At the very least, if users get errors about the ChannelCache module rename it should prompt them to check they are actually using it as intended.

We could also include a code sample like the following on how to get a `GuildChannel`:

```elixir
iex(20)> channel_id = 1229955694258684005
1229955694258684005
iex(21)> guild_id = Nostrum.Cache.ChannelGuildMapping.get(channel_id)
1226944827137069107
iex(22)> {:ok, guild} = Nostrum.Cache.GuildCache.get(guild_id)
{:ok,
 %Nostrum.Struct.Guild{
   id: 1226944827137069107,
   name: "nostrum test",
   ...
 }}
iex(24)> guild.channels[channel_id].name
"test"
```

jb3:
Since we modified the behaviour of the DMChannelCache to only cache DM Channels it is a good idea to amend the naming of the cache itself to avoid confusion, as well as fleshing out the documentation to further make it clear how caches are expected to operate.

Closes #537

### DIFF
--- a/examples/cache.ex
+++ b/examples/cache.ex
@@ -32,7 +32,7 @@ defmodule ExampleCommands do
   import Nostrum.Snowflake, only: [is_snowflake: 1]
 
   alias Nostrum.Api
-  alias Nostrum.Cache.{ChannelCache, GuildCache, UserCache}
+  alias Nostrum.Cache.{DMChannelCache, GuildCache, UserCache}
   alias Nostrum.Struct.User
 
   # Fetch the defined prefix from our config, however,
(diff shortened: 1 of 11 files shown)


jb3:
Closing this since we have realized this PR can never populate. `CHANNEL_CREATE` is only sent for guild channels and `private_channels` is no longer sent for Bot users.

I will re-open a PR removing this cache.

PR #555, opened 2024-04-19
Nuke the ChannelCache from existence

### LINKED ISSUE #537: ChannelCache documentation enhancements

jb3:
After we changed the behaviour of the ChannelCache to only store DM channels, we were then hit with behaviour that the cache never filled up at all.

This is because Discord does not send `CHANNEL_CREATE` events for private channels, nor does it send the `private_channels` field in the `READY` payload for bots.

As such, we can no longer identify DM channels the bot has open and put them into a cache (nor is it particularly useful to do so).

This PR indirectly closes #537 as we no longer have a channel cache.

Closes #537


jb3:
Thanks!

lib/nostrum/struct/channel.ex
```diff
@@ -29,44 +29,28 @@ defmodule Nostrum.Struct.Channel do
   "<#766435015768539156>"
   ```
 
-  ## Channel Cache
+  ## Channel Caching
 
-  The [`ChannelCache`](`Nostrum.Cache.ChannelCache`) module provides functionality for you to retrieve information about any channel that your application can see. It provides two functions: [`get/1`](`Nostrum.Cache.ChannelCache.get/1`) and [`get!/1`](`Nostrum.Cache.ChannelCache.get!/1`).
+  Channels are cached within the guild object they are a member of. To fetch a cached channel you should use the `Nostrum.Cache.ChannelGuildMapping` cache to map a cache to a guild ID.
```
jchristgit, line 34: We should document here that DM channels are never cached due to the way Discord gateway works here.

PR #556, opened 2024-04-19
Store all channels in channel cache

jb3:
Previously, we would only store created text and voice channels in the channels field of guilds in the `GuildCache`.

This was weird, because on application start, we would cache all channels regardless of type.

This removes that restriction to allow all channels to be cached and queried.


jb3:
Thanks!

PR #560, opened 2024-04-22
Re-support uploading files from memory

jb3:
Run `combine_files()` for all message endpoints

This ensures that we are actually handling the cases that we claim to
support, passing both paths and uploading files from memory (by using a
map).

We previously had changed this behaviour which meant that path uploading
was working but users could not create files from memory, which is a
heavily used feature and was not intentionally removed.


jb3:
Thanks!

PR #562, opened 2024-04-25
Add OTP 27.0-rc3 to our test matrix

jb3:
We can catch any potential snags with Nostrum on OTP 27 here.

I have disabled OTP 27 being used with Elixir v1.14 as that is not going to be a combination we see in production.

If we do have failures on this suite under the OTP release candidate we should investigate it but not make it a required criterion for merges.


jb3:
Thanks!

PR #566, opened 2024-04-29
Only attempt resume if the gateway says we can

jchristgit:
Per [1] and [2], if the `d` field is set to `true` we may attempt to
resume again with the same session, otherwise we should not. nostrum
would previously loop on trying to resume on the same session after an
invalid session was sent, and the behaviour of simply sending an
identify payload did not work as expected. When the session that we try
to resume with is invalid, perform a reconnect and do the whole
handshake, greeting and meeting the parents thing all over again.

[1]: https://discord.com/developers/docs/topics/gateway-events#invalid-session
[2]: https://discord.com/developers/docs/topics/gateway#resuming

### DIFF
--- a/lib/nostrum/shard/event.ex
+++ b/lib/nostrum/shard/event.ex
@@ -56,9 +56,14 @@ defmodule Nostrum.Shard.Event do
     end
   end
 
-  def handle(:invalid_session, _payload, state) do
-    Logger.info("INVALID_SESSION")
-    {{state, Payload.identity_payload(state)}, []}
+  def handle(:invalid_session, %{d: _can_resume? = true}, state) do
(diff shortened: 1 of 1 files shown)


jb3 (changes requested):
This PR introduces some warnings of unused variables (`can_resume?`), can we confirm if that is intentional and amend if not? (I am happy to fix this if you can't)

broman (approved):

jchristgit:
> This PR introduces some warnings of unused variables (can_resume?), can we confirm if that is intentional and amend if not? (I am happy to fix this if you can't right now)

it's intentional although I haven't seen the warning until now. I will update this to underscore it

jb3 (approved):

jb3:
Thanks!

PR #567, opened 2024-05-01
Implement memory optimizations in user struct

jchristgit:
- Do not automatically parse user public flags
- Remove OAuth2-only user attributes


jb3 (approved):

jb3:
Thanks!

PR #569, opened 2024-05-02
Remove Nostrum.Api.get_token

jchristgit:
This will be useless once / if we support multiple bots or library
    usage, and at the moment it simply returns a value that the user sets
    and can read on his or her own.


jb3 (approved):

jb3:
Thanks!

PR #572, opened 2024-05-03
Don't attempt to convert integers to atoms in map casting

jb3:
This fixes a bug reported in the DAPI channel which caused the Shard state machine to crash out with an error.

Discord has started sending some maps with integer keys for interactions payloads. We previously handled snowflake payloads (although these payloads used string keys anyway).

This PR:
- Checks whether `Nostrum.Util.maybe_to_atom/1` is receiving an integer, if so, just return that integer.
- Extend the existing snowflake casting check to include 0, that way we return maps that are indexed with `0` instead of `:"0"` (cursed!)

### DIFF
--- a/lib/nostrum/util.ex
+++ b/lib/nostrum/util.ex
@@ -178,12 +178,19 @@ defmodule Nostrum.Util do
   Binary `token`s that consist of digits are assumed to be snowflakes, and will
   be parsed as such.
 
+  Some maps sent from Discord are integer-indexed, for these we just return the integer
+  provided.
+
   If atom does not currently exist, will warn that we're doing an unsafe conversion.
   """
-  @spec maybe_to_atom(atom | String.t()) :: atom | integer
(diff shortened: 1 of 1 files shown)


jb3:
Thanks!

jchristgit:
Joe,

please be aware that as a nostrum maintainer, you are entitled to at
least 8 hours of sleep, so long as you have taken part in the daily
check-in at supervisor Dazey. I understand your recent increase in
medications may interfere with your sleep schedule, but it is still
important to maintain a healthy sleep schedule should you have to debug
the ratelimiter state machine early in the morning. If you need advice
for this, please contact nostrum Ltd. HR at ***@***.***`.

Best regards

broman:
Epic commit bro! Wow!

jb3:
Team, thanks!

broman:
i love nostrum

PR #573, opened 2024-05-03
Add missing fields from member API object

jb3:
Add some fields that we have missed over recent API changes to the member object.

- `flags`: Flags related to membership screening
- `avatar` and `avatar_url/3` bring support for custom guild avatars
- `pending` is also related to membership screening


jb3:
Thanks!

PR #574, opened 2024-05-03
Update state docs with new QLC examples

### LINKED ISSUE #552: State documentation uses outdated example for user discriminator query
As brought up by @jb3, the state documentation currently mentions an example of
querying users within nosedrum by matching via username and discriminator,
whilst discriminator for active user accounts are no longer in use. This might
have to be fixed in nosedrum itself, but in the meantime we could update the
documentation to use the username attribute only as it is supposed to be unique
for user accounts these days. A question for @jb3 here, could the query return
multiple users then, if you query for, for instance, an inactive user?

jb3:
Updates state docs with the previously agreed upon QLC examples in issue

Additionally, provides an example of how to evaluate a QLC query and
return the result to use within Elixir.

Closes #552

### DIFF
--- a/guides/functionality/state.md
+++ b/guides/functionality/state.md
@@ -23,18 +23,60 @@ nostrum's built-in functions to query the cache should be sufficient to cover
 common use cases. If you need more involved queries, it is recommended to use
 nostrum's [qlc](https://www.erlang.org/doc/man/qlc.html) support.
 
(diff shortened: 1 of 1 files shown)


jchristgit (approved):
Looks good to me. Maybe we should mention how to profile QLC queries in case they are unexpectedly slow, e.g. using `qlc:info/1` to see the query plan. But not sure we need that now.

PR #578, opened 2024-05-04
Update GitHub Action versions

jb3:
Updates GitHub Action versions to latest versions of available actions.

Prevents a spam of 15+ warnings in CI about deprecated node.js versions.


jb3:
Thanks!

PR #581, opened 2024-05-04
Log at INFO level when shard connection is up

jchristgit:
Make logs such as this:

    [info] Asked to reconnect with session SESSION
    [info] Re-established websocket connection
    [info] Resuming on shard session SESSION
    [info] Asked to reconnect with session SESSION
    [info] Re-established websocket connection
    [info] Resuming on shard session SESSION

Less confusing by ensuring the user knows we successfully resumed.


jb3 (approved):

jb3:
Thanks!

PR #582, opened 2024-05-04
Fix the error caused by trying to cast a datetime to datetime

jb3:
Users have reported a state machine error caused by what looks to be the
guild upsert process calling the `Nostrum.Util.maybe_to_datetime`
function on a value that is already a datetime.

We can resolve this by just returning datetimes as is from that function
if we are passed one as an argument.


jb3:
Thanks!

PR #583, opened 2024-05-04
Remove blob of zlib data from state machine errors

jb3:
Due to how zlib and the state machine work, when the state machine
crashed with an error, it would log the last blob that was received.

This blob could not be decoded from error traces because it depends on
the context that it was received by, additionally we cannot decode it as
an error logging step as once it has been decoded by a context it cannot
be inflated again.

Removing this blob is the best solution here as it shrinks the size of
state machine errors and makes them easier to read by users and Nostrum
team.

In state machine traces, the payload will now be shown as "PAYLOAD
REMOVED", if the last event was not a payload, the event will remain
unmodified.

### DIFF
--- a/lib/nostrum/shard/session.ex
+++ b/lib/nostrum/shard/session.ex
@@ -337,6 +337,27 @@ defmodule Nostrum.Shard.Session do
     {:next_state, :disconnected, %{data | conn: nil, stream: nil}, connect}
   end
 
+  # :gen_statem callback helper that removes the huge zlib data blob from any
+  # errors experienced by the shard session state machine. This zlib blob is
+  # only decodable by the zlib context that the error ocurred in anyway so it's
+  # inclusion in bug reports only hinders. It has also already been decoded by
+  # the time it reaches this call so we cannot attempt to decode it here.
+  def format_status(%{queue: queue} = state) do
+    queue =
+      Enum.map(queue, fn queue_item ->
+        case queue_item do
+          # match gun websocket messages and remove the data blob
+          {:info, {:gun_ws, conn, stream, {:binary, _payload}}} ->
+            {:info, {:gun_ws, conn, stream, {:binary, "PAYLOAD REMOVED"}}}
+
+          _ ->
+            queue_item
+        end
+      end)
+
+    %{state | queue: queue}
+  end
+
   # Internal helper. Wait for consumers to start up.
   defp wait_for_consumer_boot(reference, timeout) do
     receive do


jb3:
An example state machine trace now looks like the following (it's still long, but now all content is mostly relevant):

```

21:51:19.695 shard=0 [error] ** State machine <0.471.0> terminating
** Last event = {info,{gun_ws,<0.473.0>,#Ref<0.1439088919.2704539650.19368>,
                              {binary,<<"PAYLOAD REMOVED">>}}}
** When server state  = {connected,#{stream =>
                                         #Ref<0.1439088919.2704539650.19368>,
                                     seq => nil,
                                     '__struct__' =>
                                         'Elixir.Nostrum.Struct.WSState',
                                     session => nil,
                                     gateway => <<"gateway.discord.gg">>,
                                     shard_num => 0,conn_pid => <0.471.0>,
                                     total_shards => 1,resume_gateway => nil,
                                     conn => <0.473.0>,
                                     zlib_ctx =>
                                         #Ref<0.1439088919.2704670722.19387>,
                                     last_heartbeat_ack =>
                                         #{microsecond => {447937,6},
                                           second => 19,
                                           calendar => 'Elixir.Calendar.ISO',
                                           month => 5,
                                           '__struct__' => 'Elixir.DateTime',
                                           day => 4,year => 2024,minute => 51,
                                           hour => 20,
                                           time_zone => <<"Etc/UTC">>,
                                           zone_abbr => <<"UTC">>,
                                           utc_offset => 0,std_offset => 0},
                                     heartbeat_ack => true,
                                     heartbeat_interval => 41250,
                                     last_heartbeat_send => nil}}
** Reason for termination = error:badarg
** Callback modules = ['Elixir.Nostrum.Shard.Session']
** Callback mode = [state_functions,state_enter]
** Stacktrace =
**  [{erlang,binary_to_atom,[0],[{error_info,#{module => erl_erts_errors}}]},
     {'Elixir.Nostrum.Shard.Session',connected,3,
                                     [{file,"lib/nostrum/shard/session.ex"},
                                      {line,258}]},
     {gen_statem,loop_state_callback,11,[{file,"gen_statem.erl"},{line,1395}]},
     {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,241}]}]
** Time-outs: {1,[{state_timeout,send_heartbeat}]}
```

jchristgit (approved):
Man, I LOVE this one!

PR #584, opened 2024-05-05
Add full support for guild stickers

jb3:
This extends the support we already had in Nostrum for guild stickers by adding
all the missing API routes, structs and so on.

- Add all missing fields to sticker struct and move the enums to atoms
- Add a new `cdn_url/1` method to the sticker struct for fetching the CDN URL
- Add all API routes required to fetch and manage guild stickers
- Add API route for fetching public sticker packs
- Update all caches to be able to handle sticker updates for dispatches to users
- Add tests for all of the above
- Change the permissions to reflect the new naming for
  `manage_guild_expressions` and the addition of the new
  `create_guild_expressions` permission


lib/nostrum/struct/sticker.ex
```diff
@@ -0,0 +1,164 @@
...
+  the sticker.
+
+  This is technically a free-text field so consistency in formatting is not guaranteed.
+  """
+  @type tags :: String.t()
+
+  @typedoc """
+  Type of the sticker as an integer enum.
+  """
+  @type type :: :standard | :guild
```
jchristgit, line 58: This says it's an integer enum but then we spec it as atoms. Shouldn't we just use "Type of the sticker" and explain that a standard sticker is a global thing and guild a guild-specific one?

lib/nostrum/struct/sticker.ex
```diff
@@ -0,0 +1,164 @@
...
+
+  @typedoc """
+  Format of the sticker. This field is used to determine the return URL in `cdn_url/1`.
+  """
```
jchristgit, line 62:
```suggestion
  @typedoc """
  Format of the sticker.
  
  This field is used to determine the return URL in `cdn_url/1`.
  """
```

lib/nostrum/struct/sticker.ex
```diff
@@ -0,0 +1,164 @@
...
+  @type format_type :: :png | :apng | :lottie | :gif
+
+  @typedoc """
+  Whether this guild sticker can be used, may be false due to loss of Server Boosts
+  """
+  @type available :: boolean
```
jchristgit, line 68:
```suggestion
  @typedoc """
  Whether this guild sticker can be used on a guild.
  
  May be false due to loss of Server Boosts.
  """
  @type available :: boolean
```

lib/nostrum/struct/sticker.ex
```diff
@@ -0,0 +1,164 @@
...
+
+  @typedoc """
+  Id of the guild that owns this sticker
```
jchristgit, line 71:
```suggestion
  ID of the guild that owns this sticker.
  
  `nil` if the sticker is a global sticker.
```

lib/nostrum/struct/sticker.ex
```diff
@@ -0,0 +1,164 @@
...
+  """
+  @type guild_id :: Guild.id() | nil
+
+  @typedoc """
+  User that uploaded the guild sticker
+  """
```
jchristgit, line 77:
```suggestion
  @typedoc """
  User that uploaded the guild sticker.
  
  `nil` if the sticker is a built-in sticker.
  """
```

lib/nostrum/struct/sticker.ex
```diff
@@ -0,0 +1,164 @@
...
+  @type user :: User.t() | nil
+
+  @typedoc """
+  The sticker's sort order within its pack
+  """
+  @type sort_value :: integer() | nil
```
jchristgit, line 83: When is this nil? How can users depend on it? Which direction should it be rendered in?

lib/nostrum/api.ex
```diff
@@ -1416,6 +1419,149 @@ defmodule Nostrum.Api do
...
+
+  @doc ~S"""
+  Create a sticker in a guild.
+
+  Every guild has five free sticker slots by default, and each Boost level will
+  grant access to more slots.
+
+  Uploaded stickers are constrained to 5 seconds in length for animated stickers, and 320 x 320 pixels.
+
+  Lottie stickers can only be uploaded on guilds that have either the `VERIFIED`
```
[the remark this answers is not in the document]
jchristgit: whoops, sorry, this is the name of a German animator, and a file format. Can we mention that here? I'm not sure this format is too widely known compared to the other formats mentioned here.

```suggestion
  Stickers in the Lottie file format can only be uploaded on guilds that have either the `VERIFIED`
```

lib/nostrum/struct/sticker.ex
```diff
@@ -0,0 +1,164 @@
+defmodule Nostrum.Struct.Sticker do
+  @moduledoc """
+  A `Nostrum.Struct.Sticker` represents a sticker that can be sent inside a
+  `Nostrum.Struct.Message` and created/modified with the relevant API methods
```
jchristgit, line 4: Let's remove the prose:
```suggestion
  `Nostrum.Struct.Message`.
```

lib/nostrum/struct/sticker/pack.ex
```diff
@@ -0,0 +1,93 @@
+defmodule Nostrum.Struct.Sticker.Pack do
+  @moduledoc """
+  Represents an in-built sticker pack on Discord
```
jchristgit, line 3: Is `in-built` correct? You're the native English speaker, and it reads fine, but I feel `Represents a built-in sticker pack` would sound more natural, or am I going crazy?

lib/nostrum/struct/sticker/pack.ex
```diff
@@ -0,0 +1,93 @@
...
+
+  @typedoc """
+  SKU ID of the sticker pack.
+  """
+  @type sku_id :: Snowflake.t()
+
+  @typedoc """
+  ID of a sticker contained within the pack that should be the cover.
+  """
+  @type cover_sticker_id :: Sticker.t()
```
jchristgit, line 42: Isn't this a `Sticker.id()`?

lib/nostrum/struct/sticker/pack.ex
```diff
@@ -0,0 +1,93 @@
...
+      |> Map.update(:id, nil, &Util.cast(&1, Snowflake))
+      |> Map.update(:sku_id, nil, &Util.cast(&1, Snowflake))
+      |> Map.update(:cover_sticker_id, nil, &Util.cast(&1, Snowflake))
+      |> Map.update(:banner_asset_id, nil, &Util.cast(&1, Snowflake))
+
+    struct(__MODULE__, new)
+  end
+
+  @doc ~S"""
+  Return the banner pack URL for a given sticker pack.
```
jchristgit, line 79: What is the banner pack URL for? For stickers it's clear: it returns the sticker as a picture. Or as some strange JSON thing, in case of the lottie format, apparently. But what does a "banner pack URL" contain, especially if we document it as returning a PNG?

jchristgit (changes requested):
Very high quality. A few minor nitpicks.

lib/nostrum/api.ex
```diff
@@ -1416,6 +1419,149 @@ defmodule Nostrum.Api do
...
+  end
+
+  @doc ~S"""
+  Modify a guild sticker with the specified ID.
+
+  Pass in a map of properties to update, with any of the following keys:
+
+  - `name`: Name of the sticker (2-30 characters)
+  - `description`: Description of the sticker (2-100 characters)
+  - `tags`: Autocomplete/suggestion tags for the sticker (max 200 characters)
```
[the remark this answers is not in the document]
jb3: Probably not, this is intentionally free-text as there is no formatting mandated, for example stickers uploaded in the client put an emote in this field. I think leaving it blank as auto-complete suggestions is probably sufficient?

lib/nostrum/struct/sticker.ex
jb3: Only filled for pack information from Discord. I think the use case for needing this field is niche enough that users should know the information. No custom stickers can have this field set.

I believe it's integer ascending but there is no authoritative source on this.

lib/nostrum/struct/sticker/pack.ex
jb3: I'm actually going to sub this for "Represents a platform-curated sticker pack on Discord"

lib/nostrum/struct/sticker/pack.ex
jb3: Yes.

lib/nostrum/struct/sticker/pack.ex
jb3: It's a marketing banner for the platform curated sticker packs, for example: https://cdn.discordapp.com/app-assets/710982414301790216/store/796139192806801428.png?size=1024

lib/nostrum/api.ex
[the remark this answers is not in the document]
jchristgit: Sure, sounds good.

jchristgit (approved):
Great!!!

jb3:
Thanks!

PR #585, opened 2024-05-05
Add new guild attributes

jb3:
Adds a heap of guild attributes upstream from Discord that we had not yet added to the guild struct.

Details of new attributes are contained within the added typedocs in the changelog, for brevity I will not list them again here.

### DIFF
--- a/lib/nostrum/struct/guild.ex
+++ b/lib/nostrum/struct/guild.ex
@@ -41,6 +41,7 @@ defmodule Nostrum.Struct.Guild do
     :system_channel_id,
     :rules_channel_id,
     :public_updates_channel_id,
+    :safety_alerts_channel_id,
     :joined_at,
     :large,
     :unavailable,
@@ -50,7 +51,21 @@ defmodule Nostrum.Struct.Guild do
     :guild_scheduled_events,
     :vanity_url_code,
     :threads,
-    :stickers
+    :stickers,
+    :discovery_splash,
+    :system_channel_flags,
+    :max_presences,
+    :max_members,
+    :description,
+    :banner,
+    :premium_tier,
+    :premium_subscription_count,
+    :preferred_locale,
+    :max_video_channel_users,
+    :max_stage_video_channel_users,
+    :welcome_screen,
+    :nsfw_level,
+    :premium_progress_bar_enabled
   ]
 
   @typedoc "The guild's id"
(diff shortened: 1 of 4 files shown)


lib/nostrum/struct/guild.ex
```diff
@@ -161,6 +176,55 @@ defmodule Nostrum.Struct.Guild do
   @typedoc since: "0.5.1"
   @type threads :: %{required(Channel.id()) => Channel.t()} | nil
 
+  @typedoc "Hash of the Discovery splash screen"
+  @type discovery_splash :: String.t() | nil
+
+  @typedoc """
+  Bitset representing the system channel flags
+
+  Use `Nostrum.Struct.Guild.SystemChannelFlags` to parse.
```
jchristgit, line 185: should this reference a function?

lib/nostrum/struct/guild.ex
```diff
@@ -161,6 +176,55 @@ defmodule Nostrum.Struct.Guild do
...
+  """
+  @type system_channel_flags :: pos_integer()
+
+  @typedoc "Maximum number of presences for the guild"
+  @type max_presences :: pos_integer() | nil
+
+  @typedoc "Maximum members for the guild"
+  @type max_members :: pos_integer() | nil
```
jchristgit, line 193: When are these `nil`? When are they not?

lib/nostrum/struct/guild.ex
```diff
@@ -161,6 +176,55 @@ defmodule Nostrum.Struct.Guild do
...
+  @type description :: String.t() | nil
+
+  @typedoc "Banner hash for the guild, if prefixed with `a_` an animated GIF is available."
+  @type banner :: String.t() | nil
+
+  @typedoc "Premium tier of the guild (0-3)"
+  @type premium_tier :: 0..3
+
+  @typedoc "Number of boosts received by the guild"
+  @type premium_subscription_count :: pos_integer() | nil
```
jchristgit, line 205: This is also nonintuitive, why would this be `nil`? Wouldn't the "default" be `0`? Should we post-process it?

lib/nostrum/struct/guild.ex
```diff
@@ -161,6 +176,55 @@ defmodule Nostrum.Struct.Guild do
...
+
+  @typedoc "Preferred locale for the guild, set by the user"
+  @type preferred_locale :: String.t() | nil
+
+  @typedoc "The maximum amount of users in a video channel"
+  @type max_video_channel_users :: pos_integer() | nil
+
+  @typedoc "The maximum amount of users in a stage video channel"
+  @type max_stage_video_channel_users :: pos_integer() | nil
```
jchristgit, line 214: When are these `nil`? Can we document it? Or does `nil` simply mean "no limit"? Should we document that?

lib/nostrum/struct/guild.ex
```diff
@@ -364,6 +488,52 @@ defmodule Nostrum.Struct.Guild do
...
+  """
+  @spec discovery_splash_url(t, String.t()) :: String.t() | nil
+  def discovery_splash_url(guild, image_format \\ "webp")
+  def discovery_splash_url(%__MODULE__{discovery_splash: nil}, _), do: nil
+
+  def discovery_splash_url(%__MODULE__{discovery_splash: dsp, id: id}, image_format),
+    do: Constants.cdn_url() <> Constants.cdn_discovery_splash(id, dsp, image_format)
+
+  @doc ~S"""
+  Returns the URL of the guild's banner, or `nil` if no splash.
```
jchristgit, line 515:
```suggestion
  Returns the URL of the guild's banner, or `nil` if no banner has been set.
```

lib/nostrum/struct/guild/system_channel_flags.ex
```diff
@@ -0,0 +1,110 @@
+defmodule Nostrum.Struct.Guild.SystemChannelFlags do
+  @moduledoc """
+  Struct representing the flags on a guilds system channel
```
jchristgit, line 3:
```suggestion
  Struct representing the flags on a guild's system channel
```

jchristgit (changes requested):
Looks very good. Some minor nitpicks.

lib/nostrum/struct/guild.ex
jb3: I've updated the wording on this a bit.

jchristgit (approved):
Thanks!

PR #591, opened 2024-05-08
Check for youtube-dl version on application startup

jb3:
Recent updates to YouTube break old versions of youtube-dl, including
the last released version from 2021.

We add a startup check to attempt to fetch the version using the
youtube-dl command line if the configured youtube-dl path contains
`youtube-dl` (to avoid running this check for `yt-dlp`).

The warning advises the user to switch to a maintained fork and outputs
the located youtube-dl binary and the path it is at.

To give the user flexibility, this warning can be disabled using another
configuration key which will suppress the check from running.

### DIFF
--- a/lib/nostrum/application.ex
+++ b/lib/nostrum/application.ex
@@ -7,6 +7,8 @@ defmodule Nostrum.Application do
 
   require Logger
 
+  @outdated_youtubedl_version 2021
+
   # Used for starting nostrum when running as an included application.
   def child_spec(_opts) do
     %{
@@ -22,6 +24,12 @@ defmodule Nostrum.Application do
   def start(_type, _args) do
     Token.check_token!()
     check_executables()
+
+    unless Application.get_env(:nostrum, :suppress_youtubedl_version) do
+      if String.contains?(Application.get_env(:nostrum, :youtubedl, "youtube-dl"), "youtube-dl"),
+        do: check_youtubedl()
+    end
+
     check_otp_version()
     Logger.add_translator({Nostrum.StateMachineTranslator, :translate})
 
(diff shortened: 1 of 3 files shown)


lib/nostrum/application.ex
```diff
@@ -69,6 +77,41 @@ defmodule Nostrum.Application do
...
+  # DOES work with some other services)
+  #
+  # We opt here to give users the choice on what library to use, and allow for
+  # suppresssion of this warning, but we advise that using the default
+  # youtube-dl install may not work with YouTube.
+  defp check_youtubedl do
+    bin = System.find_executable(Application.get_env(:nostrum, :youtubedl, "youtube-dl"))
+
+    if bin do
+      with {version, 0} <- System.cmd(bin, ["--version"], stderr_to_stdout: true),
```
BrandtHill, line 91:
```elixir
with bin when is_binary(bin) <- System.find_executable(...)
```

jchristgit:
This works, but do we want it? There are other platforms that do not run non-free JavaScript and allow even older youtube-dl versions to access it.

jb3:
> This works, but do we want it? There are other platforms that do not run non-free JavaScript and allow even older youtube-dl versions to access it.

I agree with you, but given that I'd guess at least 90% of the requests *will* be for YouTube, and I hit this roadblock just the other day personally, I do not think there is harm in adding a warning, especially as it can be easily dismissed.

lib/nostrum/application.ex
```diff
@@ -22,6 +24,12 @@ defmodule Nostrum.Application do
   def start(_type, _args) do
     Token.check_token!()
     check_executables()
+
+    unless Application.get_env(:nostrum, :suppress_youtubedl_version) do
+      if String.contains?(Application.get_env(:nostrum, :youtubedl, "youtube-dl"), "youtube-dl"),
+        do: check_youtubedl()
+    end
```
jchristgit, line 31: am I understanding the code right that this will _always_ check for `youtube-dl` on the system, even if someone is not using voice nor not using youtube-dl, as long as they have it installed on their system? I am not sure we should automatically discover it like that. Can't we do it when nostrum is explicitly used to play audio via youtube-dl?

jb3:
> [!NOTE]  
> I've pushed my work here for reference, but Brandt is going to look over the executable checks anyway and refactor them, including this check in there. I'll close this PR with the goal of it being superseded by that future PR.

jchristgit:
> [!NOTE]  
> Your comment has been received by the nostrum committee and will be forwarded to Betty for detailed inspection. Please remain seated in your nostrum programming chair until the review has completed.

PR #593, opened 2024-05-10
Remove Application.get_env call from Nostrum.Api.Base

### LINKED ISSUE #546: Remove usage of token from environment in Nostrum.Api.Base
Find a way to pass in the bot token from other sources than the application
environment without requiring users to update all their API calls.

This will be more complicated than #545. I think that the only clean way to
implement this is to add the API token as a parameter to each call in
`Nostrum.Api`. I can't imagine a clean other way to do it.

jchristgit:
Closes #546


jb3 (approved):

jb3:
Thanks!

PR #594, opened 2024-05-11
Refactor executable checks and add version check for non-forked youtube-dl

BrandtHill:
Don't check ytdl/streamlink executables if ffmpeg is not configured
Add one-time version check for `youtube-dl` if not suppressed at compile time
Remove useless stuff from nostrum config
Stop starting 2 Dummy consumers in dev since events are duplicated with `:pg`


lib/nostrum/voice.ex
```diff
@@ -834,4 +846,34 @@ defmodule Nostrum.Voice do
...
+        """)
+      end
+    end
+
+    defp maybe_warn(:ytdl) do
+      unless :persistent_term.get(:has_checked_youtubedl_version, false) do
+        :persistent_term.put(:has_checked_youtubedl_version, true)
+        {:ok, _pid} = Task.start(fn -> check_youtubedl_version() end)
+      end
+    end
```
jchristgit, line 875: We should namespace this:
```suggestion
    defp maybe_warn(:ytdl) do
      unless :persistent_term.get(:nostrum_has_checked_youtubedl_version, false) do
        :persistent_term.put(:nostrum_has_checked_youtubedl_version, true)
        {:ok, _pid} = Task.start(fn -> check_youtubedl_version() end)
      end
    end
```

jchristgit (commented):
Looks good to me except for one minor suggestion.

lib/nostrum/voice.ex
BrandtHill: 👌

PR #601, opened 2024-05-21
Bump CI to OTP 27

jb3:
Move from the OTP 27 release candidate to the OTP 27 stable release.


jb3:
It will need some changes to the matrix I think, will give it a go in a second.

jb3:
Well, this looks like a bug with Credo on 1.17-otp-27 
image

jb3:
Bumping Credo version works. I am going to rewrite some commits since we now also add 1.17 into CI.

PR #603, opened 2024-05-24
Docs update: manual sharding

jb3:
- Update documentation of manual sharding methods in `Nostrum.Shard.Supervisor`
- Create new advanced documentation page showing an example of a reconnection


tignear:
I think that since the session is invalidated, the events after the `seq` will not be delivered twice...

PR #606, opened 2024-06-19
fix key error in ratelimiter

Th3-M4jor:
Came across this while trying to chase down an unrelated issue for a user where requests would just hang, and never complete.


Th3-M4jor:
This probably never came up before because it can only happen when the `:gun` connection closes while we had a request in flight

PR #610, opened 2024-06-22
Ensure user struct is casted properly before updating ets cache

Th3-M4jor:
Discovered that the user cache did not always have a properly casted user struct after an update, this fixes that as well as makes it so that updated maps are merged in a way where missing values don't replace existing ones with `nil`


jb3 (approved):
Looks good, do we need to do this style of adding `to_struct/2` definition to all cached objects? (Do we already do this?).

PR #615, opened 2024-07-19
replace `{:ok}` with the more standard `:ok`

Th3-M4jor:
DO NOT MERGE THIS before we consider `0.10` to be in a stable state, as this is intended to be part of a future `1.0.0` release.

Rationale behind replacement: In all Elixir and Erlang libraries that I can think of, you don't see the one element tuple of `{:ok}` as a possible return value. Instead you usually see `:ok | {:error, reason}` when the happy path has no meaningful return value.

This has the added benefit of a micro-optimization that avoids allocating a one element tuple on the heap.


Th3-M4jor:
Agreed. I'll try to get it into a mergable state this evening

Th3-M4jor:
Closing in favor of #667

PR #618, opened 2024-07-26
Support for new select menu components

MiniAppleTheApple:
There are 4 new select menu components which haven't added into Nostrum. Many people, including me might need them.

### DIFF
--- a/lib/nostrum/struct/component/default_value.ex
+++ b/lib/nostrum/struct/component/default_value.ex
@@ -1,0 +1,14 @@
+defmodule Nostrum.Struct.Component.DefaultValue do
+  defstruct [
+    :id,
+    :type,
+  ]
+  
+  @type t :: %__MODULE__{
+    id: id(),
+    type: type(),
+  }
+
+  @type id :: Nostrum.Snowflake.t()
+  @type type :: :user | :role | :channel 
+end
--- a/lib/nostrum/struct/component.ex
+++ b/lib/nostrum/struct/component.ex
@@ -84,6 +84,8 @@ defmodule Nostrum.Struct.Component do
   """
   @moduledoc since: "0.5.0"
 
+  alias Nostrum.Struct.Channel
+
   defmacro __using__(_opts) do
     quote do
       alias Nostrum.Struct.Component.{ActionRow, Button, Option, SelectMenu, TextInput}
@@ -167,7 +169,9 @@ defmodule Nostrum.Struct.Component do
     :max_length,
     :required,
     :value,
-    :components
+    :components,
+    :channel_types,
+    :default_values,
   ]
 
   @typedoc """
@@ -307,6 +311,10 @@ defmodule Nostrum.Struct.Component do
   Valid for [Action Row](#module-action-row).
   """
   @type components :: [SelectMenu.t() | Button.t() | nil]
+
+  @type channel_types :: [Channel.type()]
+
+  @type default_values :: [DefaultValue.t()]
 
   @spec to_struct(map()) :: struct
   def to_struct(%{} = map) do


lib/nostrum/struct/component/default_value.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Nostrum.Struct.Component.DefaultValue do
+  defstruct [
+    :id,
+    :type,
+  ]
```
Th3-M4jor, line 5:
```suggestion
  @derive Jason.Encoder
  defstruct [
    :id,
    :type,
  ]
```
Doesn't this need to be added here?

lib/nostrum/struct/component/mentionable_select.ex
```diff
@@ -0,0 +1,37 @@
...
+  @type opts :: [opt]
+
+  def mentionable_select(custom_id, opts \\ []) when is_binary(custom_id) do
+    [
+      {:custom_id, custom_id},
+      {:disabled, opts[:disabled]},
+      {:placeholder, opts[:placeholder]},
+      {:min_values, opts[:min_values]},
+      {:max_values, opts[:max_values]}
+    ]
```
Th3-M4jor, line 34:
```suggestion
    [
      custom_id: custom_id,
      disabled: opts[:disabled],
      placeholder: opts[:placeholder],
      min_values: opts[:min_values],
      max_values: opts[:max_values]
    ]
```

lib/nostrum/struct/component/role_select.ex
```diff
@@ -0,0 +1,37 @@
...
+  @type opts :: [opt]
+
+  def mentionable_select(custom_id, opts \\ []) when is_binary(custom_id) do
+    [
+      {:custom_id, custom_id},
+      {:disabled, opts[:disabled]},
+      {:placeholder, opts[:placeholder]},
+      {:min_values, opts[:min_values]},
+      {:max_values, opts[:max_values]}
+    ]
```
Th3-M4jor, line 34:
```suggestion
    [
      custom_id: custom_id,
      disabled: opts[:disabled],
      placeholder: opts[:placeholder],
      min_values: opts[:min_values],
      max_values: opts[:max_values]
    ]
```

lib/nostrum/struct/component/user_select.ex
```diff
@@ -0,0 +1,37 @@
...
+  @type opts :: [opt]
+
+  def mentionable_select(custom_id, opts \\ []) when is_binary(custom_id) do
+    [
+      {:custom_id, custom_id},
+      {:disabled, opts[:disabled]},
+      {:placeholder, opts[:placeholder]},
+      {:min_values, opts[:min_values]},
+      {:max_values, opts[:max_values]}
+    ]
```
Th3-M4jor, line 34:
```suggestion
    [
      custom_id: custom_id,
      disabled: opts[:disabled],
      placeholder: opts[:placeholder],
      min_values: opts[:min_values],
      max_values: opts[:max_values]
    ]
```

Th3-M4jor (commented):
A few minor things, and I still need to test it but otherwise I think it looks good.

lib/nostrum/struct/component/user_select.ex
MiniAppleTheApple: Thanks for the advice

lib/nostrum/struct/component/channel_select.ex
```diff
@@ -0,0 +1,43 @@
...
+  @types opts :: [opt]
+
+  def channel_select(custom_id, opts \\ []) when is_binary(custom_id) do
+    [
+      {:custom_id, custom_id},
+      {:disabled, opts[:disabled]},
+      {:min_values, opts[:min_values]},
+      {:max_values, opts[:max_values]}
+      {:channel_types, opts[:channel_types]}
+    ]
```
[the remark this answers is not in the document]
MiniAppleTheApple: I wrote like this to look identical like others `Nostrum.Struct.Component` submodules

MiniAppleTheApple:
> A few minor things, and I still need to test it but otherwise I think it looks good.

Thanks for the advice

MiniAppleTheApple:
> Hi, thanks for the PR!
> 
> I think we're good to use `[key: value]` in new code, I'm not exactly sure why we have it in different forms - but feel free to use whichever you prefer :-)

Sure, I will use `[key: value]` then.

MiniAppleTheApple:
> This looks good so far.
> 
> I don't think we need a lot of documentation for the new structs: a brief mention of what they are used for and maybe a link to the upstream documentation would be sufficient, we don't necessarily need examples etc. for a start.

Thanks for the advice, I think the PR is basically done. I can do better, but at least we can use these new components now.

MiniAppleTheApple:
> This looks good so far.
> 
> I don't think we need a lot of documentation for the new structs: a brief mention of what they are used for and maybe a link to the upstream documentation would be sufficient, we don't necessarily need examples etc. for a start.

Thanks for the advice, I think the PR is basically done. I can do better, but at least we can use these new components now/

MiniAppleTheApple:
> I wonder if there is a way to have the struct modules generated for us, given they look mostly alike. I think just having it work is a great feature to have in the first place though, we can always improve it later :-)

Sure!

Issue #620, opened 2024-08-09
Cannot use Nostrum.Api functions in a multi-host configuration due to the request Ratelimiter not running


jonklein:
SCENARIO: 
Nostrum is running in a multi-node environment, with the Nostrum.Application & consumer processes running on just one node and serving bot requests, as described in the documentation.  We'd like the other nodes in the cluster to be able to use Nostrum.Api functions needed to perform web requests, background jobs, etc.  Currently, these calls fail because the Ratelimiter instance is not running on the other nodes.

I do have a workaround, but it's probably not the right long-term solution.  I'm happy to submit a fix, but want to create this issue to figure out the correct approach.

SOLUTIONS TRIED:

- Start just the Ratelimiter on the non-main nodes: this fails when a failover occurs – the Nostrum application fails to restart on the new host due to the Ratelimiter already running.

- Run the Nostrum application on multiple nodes, but run the main consumer on only a single node: this almost works, but results in duplicate events, because _every_ node's connection dispatches to the consumer, which is globally registered with :pg in ConsumerGroup.

CURRENT WORKAROUND:

I'm running the Nostrum application on multiple nodes as described above, but have forked the library and updated the ConsumerGroup to use `:pg.get_local_members` instead of `:pg.get_members`.   Even though running the connection on multiple hosts is not the right approach in general, I do feel like `get_local_members` is more correct here - based on the current multi-node support, I'm not sure what the justification would be for dispatching to consumers on other nodes.  

OTHER POSSIBLE SOLUTIONS:

- Make it possible to run Nostrom.Api.Ratelimiter independent of Nostrum.Application: if Nostrum.Application allowed an option to not start the Ratelimiter, applications could optionally start it themselves separately on every node. 

- Register the Ratelimiter as a global process (some tradeoffs are discussed in the Ratelimiter docs)

Any thoughts on the preferred approach, especially taking into consideration possible future distributed multi-node consumer support?

Th3-M4jor:
For the `1.0.0` release, which should be the next non-patch release, we do have plans to make it so that users have to start Nostrum as part of their Supervision tree instead of leveraging the Application path. As part of that we could make it possible to configure how the ratelimiter module is named.

Also have you tried what's suggested in our [multi-node guide](https://github.com/Kraigie/nostrum/blob/master/guides/advanced/multi_node.md)?

eliasdarruda:
[...]

PR #621, opened 2024-08-10
Distribute ratelimiter requests across the cluster by route

jchristgit:
As described in #620, currently the ratelimiter is not usable across
multiple nodes. Lay the groundwork for doing this by registering
ratelimiters in a process group and selecting the matching ratelimiter
based on the request route.

Note that some minor further adjustments need to be made to make the
ratelimiter fully functional over multiple nodes, primarily related to
process registration, but some further documentation is also missing and
will be amended once manual consumer startup is prepared.

### DIFF
--- a/lib/nostrum/api/ratelimiter_group.ex
+++ b/lib/nostrum/api/ratelimiter_group.ex
@@ -1,0 +1,68 @@
+defmodule Nostrum.Api.RatelimiterGroup do
+  @moduledoc """
+  Tracks ratelimiters and determines correct ratelimiters to use per request.
+
+  ## Purpose
+
+  In a multi-node setup, users want to be able to make API requests from any
+  node in the cluster without having to worry about hitting ratelimits. This
+  module serves as the mediator between API clients on any nodes and their
+  target ratelimiter.
+
+  > ### Internal module {: .info}
+  >
+  > This module is intended for exclusive usage inside of nostrum, and is
+  > documented for completeness and people curious to look behind the covers.
+
+  ## Approach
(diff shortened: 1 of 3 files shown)


lib/nostrum/api/ratelimiter_group.ex
```diff
@@ -0,0 +1,68 @@
...
+
+  Instead, the approach is that we have a locally running ratelimiter on each
+  node, all of which are registered via the `:pg` process group managed by this
+  module. When an API request comes in, we determine its ratelimit bucket (see
+  `Nostrum.Api.Ratelimiter.get_endpoint/2`) and based on that, determine the
+  target ratelimiter by selecting it from the list of known ratelimiters via
+  `:erlang.phash2/2`.
+  """
+
+  @scope_name __MODULE__
```
Th3-M4jor, line 32: Should we make this value configurable? Issues could theoretically arise if you were running multiple bots and had the nodes connected and have 403s that were difficult to debug.

lib/nostrum/api/ratelimiter_group.ex
```diff
@@ -0,0 +1,68 @@
...
+  @group_name :ratelimiters
+
+  @doc """
+  Return a a ratelimiter PID to use for requests to the give nratelimiter `bucket`.
```
jb3, line 36:
```suggestion
  Return a ratelimiter PID to use for requests to the given ratelimiter `bucket`.
```

jb3 (changes requested):

PR #622, opened 2024-08-10
Optimize fetching individual records on each cache type

Th3-M4jor:
It was reported that for bots in a large number of servers it takes sometimes half a second to fetch a user from the cache under the current `:qlc` based method as it will do a full table scan.

As a workaround, functions which only return a single record by its primary key will be optimized to avoid `:qlc`

### DIFF
--- a/lib/nostrum/cache/user_cache.ex
+++ b/lib/nostrum/cache/user_cache.ex
@@ -23,37 +23,9 @@ defmodule Nostrum.Cache.UserCache do
   ## Behaviour specification
 
   @doc ~s"""
-  Retrieves a user from the cache by id.
-
-  This function can be called with the cache to use as an optional argument. By
-  default, the cache configured at compile time is used.
-
-  ## Example
-
-  ```elixir
-  case Nostrum.Cache.UserCache.get(1111222233334444) do
-    {:ok, user} ->
-      "We found " <> user.username
-    {:error, _reason} ->
-      "No es bueno"
-  end
-  ```
+  Retrieve a user from the cache by id.
   """
-  @spec get(User.id()) :: {:ok, User.t()} | {:error, atom}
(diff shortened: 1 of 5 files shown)


Bentheburrito:
Hello, I'm curious: couldn't you give QLC a more efficient match spec via the the `{select, MatchSpec}` option to [`:mnesia.table/2`](https://www.erlang.org/doc/apps/mnesia/mnesia.html#table/2)? 
(disclaimer: I'm not familiar with the codebase here, just learning mnesia/QLC and wondering why that option wouldn't work if y'all already evaluated it)

Th3-M4jor:
Okay this is even more confusing.

I checked against the ETS guild cache implementation and got this:

```erlang
ets:match_spec_run(ets:lookup(nostrum_guilds, 377642134112567296),
                   ets:match_spec_compile([{{'$1', '$2'}, [], ['$2']}]))
```

That's entirely as expected, plenty fast.

atlas-oc:
@jchristgit Hi! User in question here.

The bug is relatively simple to work around (just use Mnesia directly instead of the `GuildCache` module), but I can see it being a pretty big "gotcha" that's hard to debug and only shows up in production. Either way, this appears to definitely be unintended behavior and can cause simple lookups to take upwards of 3x as long as always running `Api.get_guild!`, thus (in my opinion) the Mnesia cache adapters are unusable in their current state without either documentation or a patch.

atlas-oc:
I admittedly have no familiarity with QLC, but is there any particular reason why we're using it as an abstraction here instead of expanding the "base" Cache modules to delegate more functionality to each implementation, such as `fold` and `get`? I feel letting operations be entirely implementation-dependent (so that Mnesia could, for instance, directly execute an O(1) lookup) would allow for more flexible custom cache implementations, especially if they don't natively speak QLC like :ets/:dets/Mnesia.

PR #625, opened 2024-08-15
Optimize mnesia cache querying

jchristgit:
QLC queries via mnesia-based caches that would use the `{traverse, {select, MatchSpec}}` in any shape or form would cause the QLC query to be executed in two parts, the `mnesia:table` call running the entire table over the selected match specification, and then another Erlang list comprehension that would filter the results of that across a list comprehension in plain Erlang. Per the discussion in #622, this is how such a query would look like under `qlc:info/1`:

    qlc:q([
           Guild ||
               {GuildId, Guild} <-
                   mnesia:table(nostrum_guilds,
                                [{n_objects, 100},
                                 {lock, read} |
                                 {traverse,
                                  {select,
                                   [{{'_', '$1', '$2'},
                                     [],
                                     [{{'$1', '$2'}}]}]}}]),
               GuildId =:= RequestedGuildId
          ])

The issue is that neither QLC nor mnesia can cleanly optimize it here: Mnesia does not know about the condition specified in QLC, and QLC's optimization to use a `lookup_fun` is knocked out by the fact that it can't reach into the traverse call to detect the reordered columns. It might be possible to implement this in QLC itself, given it is smart enough to figure out when to use the lookup function based on the indices, together with some cooperation from Mnesia itself. Obviously, this behaviour would lead to unacceptable performance.

This commit introduces an optimization that allows guild, member and presence cache implementations to export a `query_handle/1` function that accepts a match specification guard, that is, the "middle part" of a match specification. The guard determines which rows shall be filtered. Do note, however, that this is still unable to perform a complete optimization of lookups of single records - it will still traverse the table, but in the native ETS code.

### DIFF
--- a/lib/nostrum/cache/guild_cache/mnesia.ex
+++ b/lib/nostrum/cache/guild_cache/mnesia.ex
@@ -291,7 +291,20 @@ if Code.ensure_loaded?(:mnesia) do
     @doc "Get a QLC handle for the guild cache."
     @spec query_handle :: :qlc.query_handle()
     def query_handle do
-      ms = [{{:_, :"$1", :"$2"}, [], [{{:"$1", :"$2"}}]}]
+      query_handle([])
+    end
+
+    @doc """
+    Retrieve a query handle for the table with optional match specification guards.
+
+    ## Match specification variables
+
+    - `$1`: The guild ID
+    - `$2` The guild object itself
+    """
+    @doc since: "0.10.0"
+    def query_handle(guards) do
+      ms = [{{:_, :"$1", :"$2"}, guards, [{{:"$1", :"$2"}}]}]
       :mnesia.table(@table_name, {:traverse, {:select, ms}})
     end
 
--- a/lib/nostrum/cache/member_cache/mnesia.ex
+++ b/lib/nostrum/cache/member_cache/mnesia.ex
@@ -147,7 +147,23 @@ if Code.ensure_loaded?(:mnesia) do
     @doc "Get a QLC query handle for the member cache."
     @spec query_handle :: :qlc.query_handle()
     def query_handle do
-      :mnesia.table(@table_name)
+      query_handle([])
+    end
+
+    @doc """
+    Retrieve a query handle for the table with optional match specification guards.
+
+    ## Match specification variables
+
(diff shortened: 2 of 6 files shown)


Th3-M4jor:
I don't think this improves the performance on large tables much at all

```erlang
qlc:q([ 
       Guild ||
           {GuildId, Guild} <-
               mnesia:table(nostrum_guilds,
                            [{n_objects, 100},
                             {lock, read} |
                             {traverse,
                              {select,
                               [{{'_', '$1', '$2'},
                                 [{'==', '$1', {const, 12234}}],
                                 [{{'$1', '$2'}}]}]}}]),
           GuildId =:= RequestedGuildId
      ])
```

I'll inject a few thousand fake guilds and see how it performs locally in this PR vs `0.10.0`

Th3-M4jor:
Okay this is weird, tested with 5000 guilds injected into the table and
```elixir
iex(8)> func = fn ->
...(8)> start = :erlang.monotonic_time()
...(8)> Nostrum.Cache.GuildCache.get(guild.id)
...(8)> :erlang.monotonic_time() - start
...(8)> end)
```
is usually taking 40-ish milliseconds on `0.10.0`

But an almost empty table takes under 1 millisecond on `0.10.0`

The report said it took closer to 400ms at 4000 guilds 🤔

atlas-oc:
That is strange... another metric impacted was our CPU usage after dropping `Nostrum.Cache.GuildCache.get/1` for `:mnesia.read(:nostrum_guilds, id)`:
image
The spikes correspond to America-centric server activity (with our workload, almost *every* `:MESSAGE_CREATE` event was hitting the GuildCache, so each process spawned by the Consumer was presumably doing a full table scan before anything, causing massive CPU strain).

This is also reflected by our GC statistics, where (I think?) the entire table was constantly being loaded in and out of memory, causing 200-500 MB per reclamation:
image

Th3-M4jor:
Opened #627 to add a basic benchmark for `get/1`

I may update it with one for `create/1` and possibly also add one for the `MessageCache` this weekend if I can come up with a clean way to test both `:set` and `:ordered_set` on that one.

PR #633, opened 2024-09-20
Upgrade voice gateway v8 and support per-channel encryption modes with fallback

BrandtHill:
Minor format changes v8 to support event buffering for reconnecting.
Transport encryption modes are not consistent for all servers. AES256 is preferred, but not all hardware offers it as an available mode. In these cases, ChaCha20 will be the fallback mode as it will always be supported. As a result, encryption mode is per-channel.


broman:
This looks great! Hell yeah,  Nostrum!

PR #635, opened 2024-10-15
Stop sending port stderr to stdout

BrandtHill:
Newest versions of `yt-dlp` print out a deprecation warning for Python versions less than 3.9 regardless of the `-q --no-warnings` options. This causes the first output to be non-audio and ffmpeg won't transcode anything from yt-dlp. 
Removing `:stderr_to_stdout` from erlang ports. This means any stderr will be printed to the console but will fix issues encountered when using newer yt-dlp versions with old python.


Kraigie:
Thanks!

PR #636, opened 2024-10-20
Rename `RoleSelect` and `UserSelect` Functions to Match Module Name

kyleboe:
It looks like this was just overlooked when copy/pasting the `MentionableSelect` module.


Kraigie:
Thanks!

PR #660, opened 2025-02-16
Add telemetry as optional dependency

Th3-M4jor:
Adds the `:telemetry` library as an optional dependency.

TODO:
- [ ] Documentation
- [ ] Manual Testing


lib/nostrum/telemetry_shim.ex
```diff
@@ -0,0 +1,29 @@
+if Code.ensure_loaded?(:telmetry) do
```
BrandtHill, line 1: `:telemetry`

Th3-M4jor: Thanks for noticing the typo

Th3-M4jor:
Made the requested changes and squished down to a single commit 👍

PR #665, opened 2025-02-22
Nameless voice for multi bot

BrandtHill:


BrandtHill:
Unit tests are cooked but everything else works

broman:
I’m almost finished constructing my Nostrum shrine, as soon as I can get the Ritualis Nostrumium done the unit tests should start to pass. At the moment I’m just waiting for the virginal blood to be shipped from Amazon and then I can perform the ritual.

BrandtHill:
I @broman salute your efforts but they’re still failing. Time for you to go on an Amazon return run at Whole Foods. Pick me up some sourdough and shredded parm while you’re there.

Th3-M4jor:
Hey @BrandtHill can I ask you to update all the calls to `Nostrum.TelemetryShim` to include the bot's name in the metadata? Then update the documentation on it as well?

lib/nostrum/bot.ex
```diff
@@ -59,14 +59,27 @@ defmodule Nostrum.Bot do
...
+  Inside the `handle_event/1` callback in your consumer, the calling process already
+  has the context of the bot that generated the event for any API calls that follow.
+
+  If you have a single bot running and need to call an API asynchronously, i.e. *not* as a response
+  to a gateway event in your consumer, nostrum will automatically determine the bot.
+
+  However, if you have multiple bots running *and* you need to make unprompted bot calls, you
+  will need to assist nostrum by setting the process's context:
+  - `use Nostrum.Bot, name: MyBot` at the top of the module where the API-calling functions are defined, or
+  - `Nostrum.Bot.set_bot_name(MyBot)` dynamically before each API call
```
Th3-M4jor, line 79: What do you think about also having a function similar to `Nostrum.Bot.with_bot(MyBot, fn -> ... end), which sets the bot to use and then clears it / sets it back to what it was before returning?

I'm specifically thinking something similar to how `Ecto.Repo.transaction` works.

To me that'd be easier to understand/reason about than having `use Nostrum.Bot, name: MyBot`. Its less difficult to make sense of what its doing than us patching `def` the way we do.

lib/nostrum/cache/message_cache/mnesia.ex
```diff
@@ -49,8 +49,7 @@ if Code.ensure_loaded?(:mnesia) do
 
     # allow us to override the table name for testing
     # without accidentally overwriting the production table
-    @table_name @config[:table_name] || :nostrum_messages
-    @record_name @table_name
+    @base_table_name @config[:table_name] || :nostrum_messages
```
Th3-M4jor, line 52: I know this is mostly a me problem, but because I have it persist the data for this cache across restarts, I'll have to ask for a way to force the table to have a certain name or something that ensures backwards compatibility.

broman:
> I @broman salute your efforts but they’re still failing. Time for you to go on an Amazon return run at Whole Foods. Pick me up some sourdough and shredded parm while you’re there.

Unfortunate. I'll have to pray to Craig about why and how the ritual went wrong. Please forgive me.


> Hey @BrandtHill can I ask you to update all the calls to `Nostrum.TelemetryShim` to include the bot's name in the metadata? Then update the documentation on it as well?

You can ask, he doesn't have to grant the request. He's writing code for free and you dare to ask him for more?

lib/nostrum/bot.ex
```diff
@@ -316,9 +316,47 @@ defmodule Nostrum.Bot do
...
+
+    result = function.()
+
+    case next do
+      :reset -> set_bot(prev_pid, prev_name)
+      :clear -> set_bot(nil, nil)
+      :keep -> :noop
+    end
+
+    result
```
Th3-M4jor, line 359:
```suggestion
try do
  function.()
after
  case next do
     :reset -> set_bot(prev_pid, prev_name)
     :clear -> set_bot(nil, nil)
     :keep -> :noop
  end
end
```

Just because the fn can raise and we probably should still do what's requested instead of what'd be `:keep` if an exception was caught further up.

Th3-M4jor (approved):

khionu:
Thank you very much for picking up this issue and pushing to completion!

PR #668, opened 2025-02-25
[WIP] advanced guide to managing your own application commands

Th3-M4jor:
Guide for how you might manage your own Application Commands


Th3-M4jor:
A guide that's more or less based on what I do in my bot is ready.

When reviewing be as nit-picky as you want to be with this, if this might replace Nosedrum I want it to be as good as we can make it be.

Also, please squash merge the PR when/if its ready to merge

PR #664, opened 2025-02-22
Replace most usages of Task.start with Task.Supervisor

Th3-M4jor:
Makes most places where we use `Task.start` instead call a `Task.Supervisor`.

Partitioned to keep it WebScale 😂

PR #674, opened 2025-03-02
Select only bots in registry matchspec

BrandtHill:


Th3-M4jor:
Yeah this one's on me for using the same Registry, maybe there's a better solution though.

Like using a tuple for both, such as `{:nostrum_bot, name}` and `{:nostrum_task_supervisor, name}`.

I probably should swap that on #664 anyway.

Th3-M4jor:
I pushed a change to #664 that maybe resolves this in a cleaner way

BrandtHill:
Looks good, thanks

PR #686, opened 2025-03-18
Fix issue with emojis in `Component.Option`

snapcase:
Adding emoji to `Nostrum.Struct.Component.Option` would result in an error:
> ** (UndefinedFunctionError) function Emoji.to_struct/1 is undefined (module Emoji is not available).


broman:
@snapcase good work! I know you were worried about being able to follow through on this, as you said “I can sure as hell try” when asked to write a PR. And goddamn it, you sure as hell did. I can tell you spent a long time based on the LoC changes alone. Masterful work, sir.

PR #689, opened 2025-04-26
Postpone user requests in the shard session when not connected

### LINKED ISSUE #687: Nostrum.Api.Self.update_status/2 kills all Shard.Session processes if run before all shards are ready
While waiting for Nostrum to start each shard with proper rate-limiting, each shard session is idle in the `connecting_http` state, meaning it cannot accept payloads to push to the gateway. This poses a problem for `Nostrum.Api.Self.update_status/2`, which broadcasts a `STATUS_UPDATE` payload to ***all*** shards, regardless of their status. If this function is run after e.g. the READY event on shard 0, **all queued shards will terminate** as there is no clause to accept `{:status_update, payload}` in a non-`:connected` state.

I see a couple solutions here:
- Add a `connecting_http` function clause to catch this edge case and ignore the status update on that shard (probably ok with documentation, but lazy);
- Queue the most recent execution of `update_status` to occur after all shards are ready (this would require global tracking of such a condition, which feels wrong, especially in the context of distribution);
- Add a field to `WSState` that tracks the *desired* status, and do the following:
  - If `{:status_update, payload}` is cast during a non-`:connected` state, set the field to the desired payload
  - Upon transitioning to the `:connected` state, send the desired payload if applicable and nilify the field
  - This could probably be done for `:update_voice_state` and `:request_guild_members`, as well?

The final option feels the most robust to me, and would likely play nicer with distributed sharding.

jchristgit:
Fixes #687.

### DIFF
--- a/lib/nostrum/shard/session.ex
+++ b/lib/nostrum/shard/session.ex
@@ -92,6 +92,8 @@ defmodule Nostrum.Shard.Session do
   # receive. If this reaches zero, `:gun` will stop reading events from
   # upstream. Equivalent to setting `{active, false}` on the socket at `0`.
   @standard_flow 10
+  # Requests that can be sent from a user via `:cast` messages.
+  @user_requests [:status_update, :update_voice_state, :request_guild_members]
 
   def update_status(pid, status, activity) do
     {idle_since, afk} =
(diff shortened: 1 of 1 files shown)


jchristgit:
@jb3 still needs a review if you have time

PR #703, opened 2025-10-24
Added support for "Label" components

SachsKaylee:
Discord nowadays heavily encourages the use of the "Label" component over "ActionRow" in modals. Some components like e.g. selects in modals even requiring their use since they don't have a label of their own.

On the API layer Labels accept their single child in the "component" property instead of "components". I was torn between trying to piggy back on the existing "components" field instead of adding a new one, but ultimately decided that a solution that stays closer to the API will probably be more beneficial in the long run. Very much open to input on this one though.

Thank you for the library!

### DIFF
--- a/lib/nostrum/struct/message/component.ex
+++ b/lib/nostrum/struct/message/component.ex
@@ -30,6 +30,7 @@ defmodule Nostrum.Struct.Message.Component do
     :max_length,
     :required,
     :value,
+    :values,
     :component,
     :components
   ]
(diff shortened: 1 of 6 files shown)


lib/nostrum/struct/message/component.ex
```diff
@@ -194,6 +204,14 @@ defmodule Nostrum.Struct.Message.Component do
   @typedoc since: "0.5.1"
   @type value :: String.t() | nil
 
+  @typedoc """
+  Child component for labels.
+
+  Only present for labels.
+  """
+  @typedoc since: "NEXTVERSION"
+  @type component :: [t()]
```
jchristgit: This should be a single child then, no? Instead of a list.

```suggestion
  @type component :: t()
```

SachsKaylee: Haha, absolutely. All that talk about it being a single one and I forgot it myself right away. :) Thank you, should be fixed.

jchristgit:
Could we change the default to just be an empty string or does that cause other issues then?

Issue #651, opened 2025-02-03
Add `with_counts` Option to `get_guild/2`


drozdzynski:
#### Summary  
The [Discord API Get Guild endpoint](https://discord.com/developers/docs/resources/guild#get-guild) supports a `with_counts` query parameter, which, when set to `true`, includes `approximate_member_count` and `approximate_presence_count` in the response. Nostrum’s `Nostrum.Api.get_guild/2`.

#### Proposed Change  
Create `get_guild/2` to accept a `with_counts` option:  

```elixir
Nostrum.Api.get_guild(guild_id, with_counts: true)
```

This should translate to a request with `?with_counts=true` appended to the API call.

PR #706, opened 2025-11-19
Added guild member counts 

dssecret:
The `Nostrum.Api.Guild.get/2` and `Nostrum.Api.Self.guilds/2` endpoints are missing support for the `:with_counts` option (#651). This PR adds that option to both functions and the `:approximate_member_count` and `:approximate_presence_count` fields to the guild struct.


jchristgit (approved):
Thanks!

PR #726, opened 2026-03-24
fix: use struct type in ApiError typespec

### LINKED ISSUE #725: Incorrect typespec for exceptions
I believe the defined typespecs for the exception, like [ApiError](https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/error/api_error.ex#L14-L17) are incorrect since they do not mention the module name.

I believe it should instead be:

```elixir
@type t :: %__MODULE__{ ... }
```

Sim-hu:
The `@type t` for `ApiError` was defined as a plain map (`%{...}`) instead of a struct (`%__MODULE__{...}`). Since `defexception` creates a struct, the typespec should reflect that for accurate Dialyzer analysis.

This is consistent with how all other struct types in the codebase define their `@type t` (e.g., `Nostrum.Struct.Emoji`, `Nostrum.Struct.Guild.Role`).

Closes #725

### DIFF
--- a/lib/nostrum/error/api_error.ex
+++ b/lib/nostrum/error/api_error.ex
@@ -11,7 +11,7 @@ defmodule Nostrum.Error.ApiError do
     :response
   ]
 
-  @type t :: %{
+  @type t :: %__MODULE__{
           status_code: status_code,
           response: response
         }


jchristgit:
Thanks!

